PR #10, opened 2018-06-22
kola: change paths of grub tests to flatcar

iaguis:


dongsupark:
Looks good.
But there are 2 more places in init() with `coreos.update.grub`. Do we want to change them too?

iaguis:
> But there are 2 more places in init() with coreos.update.grub. Do we want to change them too?

No, those are just test names and I'm trying to keep changes minimal.

PR #16, opened 2018-10-10
kola: add TestSymlinkFlatcar to check for a symlink to /usr/lib/flatcar

dongsupark:
To avoid potential issues with a symlink `/usr/lib/coreos` to `/usr/lib/flatcar`, let's add a simple test `TestSymlinkFlatcar` that checks for the symlink on every Flatcar build.

See also https://github.com/flatcar-linux/coreos-overlay/pull/8

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -36,6 +36,7 @@ func init() {
 			"PortSSH":          TestPortSsh,
 			"DbusPerms":        TestDbusPerms,
 			"Symlink":          TestSymlinkResolvConf,
+			"SymlinkFlatcar":   TestSymlinkFlatcar,
 			"UpdateEngineKeys": TestInstalledUpdateEngineRsaKeys,
 			"ServicesActive":   TestServicesActive,
 			"ReadOnly":         TestReadOnlyFs,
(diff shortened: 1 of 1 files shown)


kola/tests/coretest/core.go
```diff
@@ -214,6 +215,35 @@ func TestSymlinkResolvConf() error {
...
+		return fmt.Errorf("unable to lstat on %s: %v", coreosPath, err)
+	}
+	if !IsLink(f) {
+		return fmt.Errorf("coreos path %s is not a symlink.", coreosPath)
+	}
+	resolved, err := os.Readlink(coreosPath)
+	if err != nil {
+		return fmt.Errorf("unable to resolve symlink %s.", f)
+	}
+	if !strings.Contains(resolved, "flatcar") {
```
iaguis: Should we check for equality here?

dongsupark: Fixed!

kola/tests/coretest/core.go
```diff
@@ -214,6 +215,35 @@ func TestSymlinkResolvConf() error {
 	return nil
 }
 
+func TestSymlinkFlatcar() error {
+	coreosPath := "/usr/lib/coreos"
+	flatcarPath := "/usr/lib/flatcar"
+
+	f, err := os.Stat(coreosPath)
```
iaguis: You need to use `Lstat` here.

iaguis: Fixed this.

PR #18, opened 2018-10-17
Merge upstream to flatcar-master, 2018-10-17

dongsupark:
Merge upstream/master to flatcar-linux/flatcar-master.


dongsupark:
Let's not merge it for now.
It's possible to stay with old commits as much as possible when resolving merge conflicts. That's actually better, because currently most of the upstream changes in specs.go are not needed for Flatcar at all.

PR #20, opened 2019-04-30
kola: check also for the systemd cgroup driver

dongsupark:
Flatcar edge 2051.99.1 or newer has the cgroup driver "systemd" instead of "cgroupfs". So we need to also update Kola tests to avoid test failures like:

```
--- FAIL: docker.base (541.40s)
    --- FAIL: docker.base/docker-info (12.17s)
            docker.go:575: unexpected cgroup driver systemd
```

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -571,7 +571,7 @@ func testDockerInfo(expectedFs string, c cluster.TestCluster) {
 		c.Errorf("unexpected security options: %+v", info.SecurityOptions)
 	}
 
-	if info.CgroupDriver != "cgroupfs" {
+	if info.CgroupDriver != "cgroupfs" && info.CgroupDriver != "systemd" {
 		c.Errorf("unexpected cgroup driver %v", info.CgroupDriver)
 	}
 


kola/tests/misc/network.go
```diff
@@ -127,6 +127,7 @@ func NetworkListeners(c cluster.TestCluster) {
 		{"tcp", "22", "systemd"},          // ssh
 		{"udp", "68", "systemd-network"},  // dhcp6-client
 		{"udp", "546", "systemd-network"}, // bootpc
+		{"tcp", "10010", "containerd"},    // containerd
```
alban: https://github.com/moby/moby/issues/37507#issuecomment-406654334

It references `disabled_plugins = ["cri"]` in the containerd configuration file, that was changed for the Edge channel. I wonder if that was a correct change: shouldn't cri be using a unix socket instead of a tcp socket?

iaguis: will this make the test fail on non-edge builds?

dongsupark: Why would it make the non-edge tests fail? On the contrary, it will actually fix failures on both edge and non-edge channels.

The change in the cgroup driver of Docker should not change anything on non-edge builds, because non-edge builds will always have `--cgroup-driver=cgroupfs`. The cgroup driver change will only affect edge builds, where `--cgroup-driver=systemd` is being used. However, I wanted to make kola tests detect both cgroupfs and systemd, regardless of channels. That's because in the future we could have situations where the systemd cgroup driver is enabled in non-edge channels. Doing that, we can also avoid applying the edge changes back again to flatcar-master.

The change in NetworkListeners would also not affect existing test cases, because it's just adding another entry to the whitelist, AFAIU.

iaguis: I thought non-edge releases don't have `containerd` and that whitelist was strict (if the process name is in the whitelist and not on the system, it would fail). Just wanted to check but if it's just a whitelist then I'm fine with it.

dongsupark: I don't know if it works with a unix socket.
I'm fine with dropping the containerd change from this PR, until we could test it again with a new containerd config.

dongsupark:
Dropped the containerd change, kept only the cgroup driver change.

alban:
I don't have context about kola but the change looks good to me.

PR #29, opened 2019-10-25
kola: fix cgroup parameters for docker

dongsupark:
CI should run docker run with options together like `--cpu-quota=1000 --cpu-period=1000`.

That's because usually `--cpu-period` should work with `--cpu-quota`, as mentioned in https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources.

That will fix an issue of docker run hanging for a long time, when running with the systemd cgroup driver.


t-lo (approved):
LGTM, thank you, and please pardon the long delay on this PR.

PR #36, opened 2019-12-13
kola: add raid0 tests for root and data devices

dongsupark:
We should test root and data devices not only with raid1 but also with raid0, to catch issues with raid0 devices.

To be able to parametrize raid levels for RAID tests, we should create a helper function `ExecTemplate()` that reads template and parses it into an actual output string. Also make the existing raid tests use `ExecTemplate()`.

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -136,6 +136,19 @@ func init() {
 		Distros:     []string{"cl"},
 	})
 
+	// data with raid0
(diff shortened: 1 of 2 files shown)


tormath1:
opened "Dec 13, 2019" ahah - @dongsupark are you still interested by contributing to this PR ? I think it would be nice to have this new tests !

dongsupark:
Done, added changelog.

PR #43, opened 2020-01-31
kola/tests/etcd: Double timeout duration from 2.5 to 5 min

pothos:
Increase timeout for etcd cluster to come up. This also
gives more time to check the server manually before it
is deleted due to the timeout.


pothos:
Maybe not needed after all.

PR #46, opened 2020-02-03
platform/api/aws: Iterate over all subnets to run the instances

pothos:
The subnet determines the Availability Zone. If in that
zone no machines are available, running an instance
will fail. The other Availability Zones probably have
machines available and the code should try to run
instances there instead of giving up too early.

### DIFF
--- a/platform/api/aws/network.go
+++ b/platform/api/aws/network.go
@@ -304,8 +304,8 @@ func (a *API) createSubnets(vpcId, routeTableId string) error {
 	return nil
 }
 
-// getSubnetID gets a subnet for the given VPC.
-func (a *API) getSubnetID(vpc string) (string, error) {
+// getSubnetIDs gets the subnets for the given VPC.
+func (a *API) getSubnetIDs(vpc string) ([]string, error) {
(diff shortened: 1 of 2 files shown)


pothos:
I tested it by deleting and creating the subnet in `us-west-2d` where not machines are/were available.

platform/api/aws/network.go
```diff
@@ -315,14 +315,21 @@ func (a *API) getSubnetID(vpc string) (string, error) {
...
-		return "", fmt.Errorf("unable to get subnets for vpc %v: %v", vpc, err)
+		return nil, fmt.Errorf("unable to get subnets for vpc %v: %v", vpc, err)
 	}
+
+	var ids []string
+
 	for _, id := range subIds.Subnets {
 		if id.SubnetId != nil {
-			return *id.SubnetId, nil
+			 ids = append(ids, *id.SubnetId)
```
dongsupark: NIT: an unnecessary blank before `ids`.
`gofmt` is your friend. ;-)

pothos: Fixed.

PR #48, opened 2020-02-06
plume: Include -arm64 in AMI name for arm64 images

pothos:
Currently the AMI names clash for amd64 and arm64 releases.
Therefore, no AMIs for arm64 were created.
Add a "-arm64" tag to the AMI name to distinguish the
arm64 AMI names. This introduces no change to the AMI name
for the existing amd64 AMIs.


pothos:
Will try to add. Do we use `ore aws upload` anywhere?

pothos:
Have added a patch for `ore`.

PR #50, opened 2020-02-12
ore|plume: Remove PV support

pothos:
Removes uploading of the deprecated AWS PV AMIs in `ore` and `plume`.


pothos:
Have removed the functionality there, too.

PR #54, opened 2020-02-26
kola: remove tests for docker 1.12 torcx profile

dongsupark:
Since the docker 1.12 profile was removed in Edge, many kola tests started to fail to run, because of the missing docker 1.12 profile. Ideally we should be able to make a different set of profiles for each channel.
Unfortunately the current Jenkins pipeline is not able to do that. It takes into account only flatcar-master branch.

Let's remove tests for the docker 1.12 profile. It will not be so big deal, because anyway Stable and Beta have been using docker 18 since a long time.


t-lo:
Superseded by https://github.com/flatcar-linux/mantle/pull/61. @dongsupark Do you think we can close this?

dongsupark:
Yes

PR #56, opened 2020-02-27
Verify repo checkout for branch names

pothos:
Currently only SHA1 git commits can be verified
to be checked out. Add support to verify that
a branch is checked out.


pothos:
Will rework it to handle detached HEAD state, too.

PR #57, opened 2020-03-02
kola: Improve debugging abilities through SSH login

pothos:
- kola: Improve debugging abilities through SSH login  
    The tests in "kola run" used a temporary SSH key and
    directly deleted the machines after a failure.
    
    Add the same SSH key option as in "kola spawn"
    to include additional SSH keys (just "-k" is
    enough to include the agent/default SSH keys).
    Also use the option to skip machine removal as
    in "kola spawn --remove=false".

- Use the private IP because the default firewall
    rules do not allow incoming traffic on 2379 for
    the public IP.
    Also, there are no FLATCAR_*_IP* variables set up in
    /etc/environment for machines running in
    Google Compute Engine because (as with afterburn)
    the prefix is still COREOS_.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -22,6 +22,8 @@ import (
 	"sort"
 	"text/tabwriter"
 
+	"golang.org/x/crypto/ssh/agent"
+
 	"github.com/coreos/pkg/capnslog"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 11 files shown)


pothos:
Tests are now starting additional machines with cloud-config but there the bash script errors. To debug this I need a way to tell kola to use an additional SSH key besides the random key that it generates.

pothos:
The Kubernetes test passes, I'm quite surprised that it isn't outdated :)

PR #58, opened 2020-03-03
kola: adjust tests related to systemd-timesyncd for systemd >= v243

dongsupark:
Since v243, systemd-timesyncd prints out its status message like this:

```
Initial synchronization to time server 10.0.0.1:123 (10.0.0.1).
```

So we need to deal with another case for that.
Let's create a list of possible output messages from systemd-timesyncd, so it can deal with possible systemd versions.

### DIFF
--- a/kola/tests/misc/ntp.go
+++ b/kola/tests/misc/ntp.go
@@ -24,6 +24,12 @@ import (
 	"github.com/coreos/mantle/util"
 )
 
+var timesyncdMsgs = [][]byte{
+	[]byte(`Status: "Synchronized to time server 10.0.0.1:123 (10.0.0.1)."`),                    // systemd < 241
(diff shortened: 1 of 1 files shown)


kola/tests/misc/ntp.go
```diff
@@ -43,19 +49,29 @@ func NTP(c cluster.TestCluster) {
 	}
 	defer m.Destroy()
 
+	c.MustSSH(m, "sudo systemctl restart systemd-timesyncd.service")
```
pothos: I think it's better if we make sure that timesyncd is enabled instead of starting it here for the test.

dongsupark: Removed the `systemctl restart` command, and changed the list of systemd services so it keeps systemd-timesyncd.

PR #59, opened 2020-03-06
helpers/azure-*: helper scripts for manually running azure tests

t-lo:
This commit adds 2 helper scripts for manually running tests
on Azure.

1.  azure-prep-image.sh: prepare an azure image from a page blob
     for use with the run-kola script.
2.  azure-run-kola.sh: run kola tests on all images of an azure
     resource group.

The first script can be used to prepare images (for edge, alpha,
beta, and stable) of a release and copy those into separate
test resource groups, each in a different region.

The second script would then run tests on all images of all
test resource groups that match a predefined prefix, covering all
regions that have test images..

After testing concluded the resource group can be deleted e.g.
via the Azure web interface.

Signed-off-by: Thilo Fromm <thilo@kinvolk.io>

### DIFF
--- a/helper-scripts/azure-prep-image.sh
+++ b/helper-scripts/azure-prep-image.sh
@@ -1,0 +1,125 @@
+#!/bin/bash -e
+#
+# Create an image from a page blob and copy it to a custom region (so as to
+# cirumvent regional CPU limitationf when running tests)
+#
+
+# dummy "az" for testing
+#function az() {
+#	echo "EXEC: az $@"
+#}
+
+function channel_to_region() {
+	local blob_name="$1"
+	local channel=$(echo "$blob_name" \
+						| sed 's/.*\(alpha\|beta\|stable\|edge\).*/\1/')
+
+	case "$channel" in
+		alpha)	echo "uksouth";;
+		beta)	echo "germanywestcentral";;
+		stable) echo "francecentral";;
+		edge)	echo "ukwest";;
+		*)		echo "ERROR: unknown channel '$channel'" >&2
+				exit 1;;
+	esac
+}
+# --
+
+function disk_from_page_blob() {
+	local blob_name="$1"
+	local name="$2"
+	local rg="$3"
+
+	local out=$(mktemp)
+	trap "rm -f $out" EXIT
+
+	az disk create --name "$name" -g "$rg" \
+		--source https://$rg.blob.core.windows.net/publish/"$blob_name".vhd \
+		> "$out"
+	
+	# print disk id
+	cat "$out" | jq -r '.id'
+
+	rm -f "$out"
+	trap - EXIT
+}
+# --
+
+function image_from_disk() {
+	local name="$1"
+	local rg="$2"
+	local disk_id="$3"
+
+	az image create -g $rg --name "$name" \
+						   --source "$disk_id" --os-type linux
+}
+# --
+
+function copy_to_region() {
+	local name="$1"
+	local rg="$2"
+	local dest_rg="$3"
+	local dest_region="$4"
+
+	az image copy --source-resource-group $rg \
+				  --source-object-name "$name" \
+				  --target-location "$dest_region" \
+				  --target-resource-group "$dest_rg" \
+				  --cleanup
+}
+# --
+
+function cleanup() {
+	local disk_id="$1"
+	local rg="$2"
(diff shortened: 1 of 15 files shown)


pothos:
Just a quick question: What about `./ore azure create-image`?

t-lo:
> Just a quick question: What about `./ore azure create-image`?

Didn't test :)

helper-scripts/azure-run-kola.sh
```diff
@@ -0,0 +1,46 @@
...
+		name=$(basename $i)
+		echo "#############################################"
+		echo " Running tests for $name"
+		echo "---------------------------------------------"
+		run_test "$i" "$name" "$region" "$parallel"
+	done
+}
+# --
+
+if [ "$(basename $0)" = "run-kola.sh" ] ; then
```
dongsupark: s/run-kola/azure-run-kola/
Otherwise the script does not run.

helper-scripts/azure-prep-image.sh
```diff
@@ -0,0 +1,83 @@
...
+	image_from_disk "$blob_name" "$rg" "$disk_id"
+
+	echo " #### 3. Copying image to rg $dest_rg in region $dest_region"
+	copy_to_region "$blob_name" "$rg" "$dest_rg" "$dest_region"
+
+	echo " ====> All done."
+}
+# --
+
+if [ "$(basename $0)" = "prep-image.sh" ] ; then
```
dongsupark: d/prep-image/azure-prep-image/

t-lo: Legacy names from earlier during script development, meh. Thank you, good catch!

pothos:
It would be good to make sure that `ore azure create-image` works and then use that in Jenkins as we do for the other platforms, too.

t-lo:
> It would be good to make sure that `ore azure create-image` works and then use that in Jenkins as we do for the other platforms, too.

I agree. The scripts provided are for manual test runs.

helper-scripts/azure-prep-image.sh
```diff
@@ -0,0 +1,125 @@
...
+}
+# --
+
+function copy_to_region() {
+	local name="$1"
+	local rg="$2"
+	local dest_rg="$3"
+	local dest_region="$4"
+
+	az image copy --source-resource-group $rg \
```
pothos: Where is that defined? In the Docker image I don't see it `docker run --rm -it mcr.microsoft.com/azure-cli` (cli version 2.2.0):
```
bash-5.0# az image copy --help
az image: 'copy' is not in the 'az image' command group. See 'az image --help'. If the command is from an extension, please make sure the corresponding extension is installed. To learn more about extensions, please visit https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview
bash-5.0# az image copy 
az image: 'copy' is not in the 'az image' command group. See 'az image --help'. If the command is from an extension, please make sure the corresponding extension is installed. To learn more about extensions, please visit https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-overview
bash-5.0# az image --help

Group
    az image : Manage custom virtual machine images.

Subgroups:
    builder [Preview] : Manage and build image builder templates.

Commands:
    create            : Create a custom Virtual Machine Image from managed disks or snapshots.
    delete            : Deletes an Image.
    list              : List custom VM images.
    show              : Gets an image.
    update            : Update custom VM images.
```

pothos: Same when installed through the installer script.

pothos: Ah, found it: https://docs.microsoft.com/en-us/cli/azure/ext/image-copy-extension/image?view=azure-cli-latest

pothos:
To do the same now with kola on latest `flatcar-master`:
```
./kola run -platform azure --azure-image-file LOCAL_FILE --azure-location LOCATION
```
or with uploading a blob that is kept for later:
```
./ore ore azure upload-blob-arm --storage-account flatcar --resource-group flatcar --container test --blob-name NAME.vhd --file LOCAL_FILE
# prints a SAS URL which is needed for a cross-region copy
./kola run -platform azure --azure-blob-url SAS_URL --azure-location LOCATION
```

t-lo:
Closing this as it's outdated by now.

PR #61, opened 2020-03-11
kola: skip Docker torcx tests in case of Edge channel

dongsupark:
Since the docker 1.12 profile was removed in Edge, many kola tests started to fail to run, because of the missing docker 1.12 profile. Ideally we should be able to make a different set of profiles for each channel. Unfortunately the current Jenkins pipeline is not able to do that. It takes into account only flatcar-master branch.

To work around the issue, read directly a version from `/etc/os-release` to detect if the image is Edge. If the semver's minor version is 99, then it means Edge, so skip the tests.

Note, the code is only a quick-and-dirty approach. In the future, we should create a generic way to make channel names available for individual tests.

### DIFF
--- a/kola/tests/docker/torcx_docker_flag.go
+++ b/kola/tests/docker/torcx_docker_flag.go
@@ -15,8 +15,11 @@
 package docker
 
 import (
+	"fmt"
 	"regexp"
+	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/mantle/kola/cluster"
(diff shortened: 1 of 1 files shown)


t-lo:
Testing done:
- ran `docker.torcx-flag-file.cloud-config` and `docker.torcx-flag-file` on Edgel Alpha, Beta, and Stable. (only on Azure, but results should be repeatable on all other platforms)
- Tests passed for Alpha, Beta, and Stable. Tests were skipped for Edge.

LGTM!

PR #64, opened 2020-03-16
kola: Create Azure image for the whole flight

pothos:
The cluster's resource group was used to store
the created image which caused it to be copied
again for each new cluster.
Use a separate resource group for the whole
flight to store the image for all clusters.


pothos:
The standard diff is confusing, better check it with `git diff --diff-algorithm=histogram flatcar-master`

PR #67, opened 2020-03-18
Update Packet API to lastest version

pothos:
- Update Packet API to latest version
    Porting needs a new SSH client option. Other vendored files were deleted by "make vendor".
- Retry on provisioning timeouts
- Improve bonding check and lower threshold
    After https://github.com/coreos/bugs/issues/2065
    a test for "excessive bonding link status messages"
    was introduced which also is good to keep for
    https://github.com/coreos/bugs/issues/2374.
    
    However, having this message printed 10 times
    does not directly relate to an error.
    The test should check if something like
    'bond0: Gained carrier' or
    'link status definitely up for interface enp0s20f0'
    is coming at the end and then continue.
    Add a second match for these messages that skips
    the test. Also lower the threshold to see if the
    logic works well.


pothos:
I need to make the loop cover the rest of the function so that if `StartMachine` gives an error we can try with an other machine.

pothos:
Done

PR #80, opened 2020-04-09
kola: increase cgroup memory limit to 50MB

dongsupark:
With 30MB, docker run still fails in some cases.
Let's increase the cgroup memory limit to 50MB, to avoid such failures.

Other details are already described in https://github.com/flatcar-linux/mantle/pull/79.

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -261,7 +261,7 @@ func dockerResources(c cluster.TestCluster) {
 	// ref https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources
 	for _, dockerCmd := range []string{
 		// must set memory when setting memory-swap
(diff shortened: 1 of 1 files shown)


pothos:
What about `--shm-size=1m`, `--memory-reservation=10m`, and `--kernel-memory=10m`? Maybe those could be related to the crash?

dongsupark:
I am not sure I understand the question.

In the code, each line is interpreted as a set of parameters for each `docker run` command. So each line is independent of each other.
`--kernel-memory=10m` becomes `docker run --kernel-memory=10m`, and it has nothing to do with the `--memory` option.

AFAIK there has been no case where the `--kernel-memory=10m` test failed.
Failed one was either `--memory=30m --memory-swap=30m` or `--memory=30m --oom-kill-disable=true`.

pothos:
Ok, thanks for explaining!

PR #90, opened 2020-04-20
ESX: Add kola option to work without a base VM

pothos:
Cloning from a base VM to spare the image upload is not supported by
the free version of ESXi. It is also useful for "kola spawn" to work
without a base VM.
Add an option to create a VM directly from an uploaded image.
The userdata is set over guestinfo variables instead of modifying the
default values of the OVF metadata.

# How to use/testing done

Create a VM on a [Packet.net ESXi server](https://www.packet.com/resources/guides/esxi/):
```
./kola spawn --platform esx --esx-config-file esx.json --esx-ova-path flatcar_production_vmware_ova.ova  -d
```
By default this give an error because the VM expects DHCP to be set up which isn't the case in the default Packet.net ESXi configuration. To work around this I used a custom userdata in kola to set a password and log in to the VM over the VGA console through the web UI.

### DIFF
--- a/platform/api/esx/api.go
+++ b/platform/api/esx/api.go
@@ -55,6 +55,7 @@ type Options struct {
 	User       string
 	Password   string
 	BaseVMName string
+	OvaPath    string
 }
 
 var plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "platform/api/esx")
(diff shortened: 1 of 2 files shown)


platform/api/esx/api.go
```diff
@@ -585,7 +637,7 @@ func (a *API) updateOVFEnv(vm *object.VirtualMachine, userdata string) error {
 	}
 
 	for _, item := range mvm.Config.VAppConfig.(*types.VmConfigInfo).Property {
-		if item.Id == "guestinfo.coreos.config.data" {
+		if item.Id == "guestinfo.ignition.config.data" {
```
marga-kinvolk, line 640: This change (and the next one) seems unrelated to the overall change and I see no mention of it in the commit.

Was this a bug that wasn't working before? Could this affect something that was already configured with the wrong keys?

pothos: I've added it to the commit message. The guest variables are set directly in one case and I used the new generic names there, and also updated them here.

PR #93, opened 2020-04-22
esx: Allow to configure static IPs

pothos:
- esx: Allow to configure static IPs    
    The default ESXi setup does not have a DHCP server but expects
    static IPs to be used (e.g., see https://www.packet.com/resources/guides/esxi/).
    Allocate static IPs for each machine from a number of addresses
    starting with a specific address given inside the subnet (in Packet this
    is the next IP address after the host's IP, see link above).
    A networkd configuration file is used with Ignition but cloud-config writes its
    own networkd configuration file based on the guestinfo.interface* variables.
    The coreos-metadata service uses the COREOS_CUSTOM_* variables that are known
    to the Container Linux Config Transpiler, too, and got some robustness fixes.

# How to use

```
./kola run -d --platform esx --esx-config-file esx.json --esx-ova-path flatcar_production_vmware_ova.ova  --esx-first-static-ip A.B.C.D --esx-first-static-ip-private E.F.G.H --esx-static-ips 4 --esx-gateway I.J.K.L --esx-gateway-private M.N.O.P --esx-subnet-size 29
```

Because we only have 4 static IPs, do not set `--parallel X` because kola will wait anyway until a machine terminates again and frees an IP.

# Testing done

Use an ESX server on Packet and use the gateways as visible in the network setting of the Packet web UI, and use the next IP after the server IP as available first static IPs (see link above). When the ESX server is provisioned, a /28 subnet can be specified instead of the defautl /29 but I didn' test this.
I fixed most test failures and skipped tests that rely on network connectivity in the initramfs.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -118,6 +118,12 @@ func init() {
 	sv(&kola.ESXOptions.Profile, "esx-profile", "", "ESX profile (default \"default\")")
 	sv(&kola.ESXOptions.BaseVMName, "esx-base-vm", "", "ESX base VM name")
 	sv(&kola.ESXOptions.OvaPath, "esx-ova-path", "", "ESX VM image to upload instead of using the base VM (build with: ./image_to_vm.sh --format=vmware_ova ...)")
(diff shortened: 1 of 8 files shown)


pothos:
This is ready for review now. I'll switch back to `ignition.config` variables ones https://github.com/flatcar-linux/coreos-cloudinit/pull/4 is referenced in `coreos-overlay`.

PR #96, opened 2020-04-23
cmd/kola: add developer as a supported channel

marga-kinvolk:
This change adds "developer" as one of the valid channel values.

Needed for: kinvolk/PROJECT-flatcar-linux#294

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -40,7 +40,7 @@ var (
 	kolaArchitectures  = []string{"amd64"}
 	kolaPlatforms      = []string{"aws", "azure", "do", "esx", "gce", "openstack", "packet", "qemu", "qemu-unpriv"}
 	kolaDistros        = []string{"cl", "fcos", "rhcos"}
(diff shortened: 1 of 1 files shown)


pothos:
The channel `developer` doesn't give meaningful information which tests should be run. I would rather like that the Jenkins job uses `stable` as a default to run the tests, or even gets a new parameter to hand the correct information over.

marga-kinvolk:
Uhm, I see your point about channel exclusion not working with the developer channel, yes.

The problem is that what is the right channel will depend on the specific build being triggered. Any default we pick will be wrong for builds of a different channel, so I guess it should be passed through the pipeline.

What do you think about adding a CHANNEL_BASE parameter to the os/manifest pipeline? (The name comes from the RELEASE_BASE that we already have... It could also be TEST_CHANNEL, but we might want to use it for more than just tests...)

Which ever parameter we pick, it would need to be wired all the way through until the kola jobs.

pothos:
Yes, passing it through from `os/manifest` is the laborious part ;)

PR #98, opened 2020-04-24
esx: Optionally read static IP config from ESX profile

pothos:
- esx: Optionally read static IP config from ESX profile
    Instead of a long list of command line options, the static IP configuration
    should be part of the config file.
    The values can be left out and the command line options get preference.
- esx: Always remove created VM on error
- esx: Free static IP up when device creation failed

### DIFF
--- a/platform/api/esx/api.go
+++ b/platform/api/esx/api.go
@@ -156,6 +156,24 @@ func New(opts *Options) (*API, error) {
 		if opts.Password == "" {
 			opts.Password = profile.Password
 		}
+		if opts.StaticIPs == 0 {
+			opts.StaticIPs = profile.StaticIPs
+		}
+		if opts.FirstStaticIp == "" {
+			opts.FirstStaticIp = profile.FirstStaticIp
+		}
+		if opts.FirstStaticIpPrivate == "" {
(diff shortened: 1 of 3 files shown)


platform/api/esx/api.go
```diff
@@ -383,11 +383,13 @@ func (a *API) CreateDevice(name string, conf *conf.Conf, ips *IpPair) (*ESXMachi
 		plog.Debugf("Configuring userdata %q", userdata)
 		err = a.updateGuestVariable(vm, "guestinfo.ignition.config.data", userdata)
 		if err != nil {
+			_ = a.deleteDevice(vm)
```
marga-kinvolk, line 386: Instead of repeating this call every time, I think it would be better to use a defer right after creating the VM.
I think this is would you'd need:
```
	defer func() {
		if err != nil {
			_ = a.deleteDevice(vm)
		}
	}()
```

pothos: Definitely but that only works if `err` is not shadowed by accident. Added a comment on this and converted a case which is not strictly shadowing `err` but would be if it is moved into a bracket later.

marga-kinvolk: Alright, looks good!

PR #107, opened 2020-05-25
Add a gangue subcommand to clean up files in GCS

marga-kinvolk:
# Add a gangue subcommand to clean up files in GCS

By adding developer builds, we've started generating a lot of files in GCS that are actually ephemeral. This change adds a new `prune` subcommand to the gangue command (until now only `get` existed). This subcommand can be used to delete files under a specific path if they are older that some amount of days.

# How to use / Testing done

```
go build -o bin/gangue github.com/coreos/mantle/cmd/gangue
bin/gangue prune flatcar-jenkins developer/sdk \
    --json-key /path/to/creds/gce-service-account.json --days=15
```

This will delete all files older than 15 days under the `developer/sdk` path.  I've run this for sdk and for boards and verified that the files deleted were correct.

This change is part of the clean up needed for building PR branches (kinvolk/PROJECT-flatcar-linux#294)

### DIFF
--- a/cmd/gangue/prune.go
+++ b/cmd/gangue/prune.go
@@ -1,0 +1,105 @@
+// Copyright 2020 Kinvolk GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


cmd/gangue/prune.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2017 CoreOS, Inc.
```
pothos, line 1:
```suggestion
// Copyright 2020 Kinvolk GmbH
```

marga-kinvolk: Applied, thanks

PR #110, opened 2020-06-12
platform.go: add status and journal output to failed systemd units

t-lo:
# Add status and journal output to failed systemd units

`CheckMachine` in `platform/platform.go` returns an error if one or multiple systemd unit(s) failed to start. While it prints the failed unit name it does not provide details on the failure.

This PR adds output from `systemctl status <unit>` and `journalctl -b -u <unit>` for a failed unit.

# How to use
Run a test with `kola` using an OS image that fails to start a unit. The error output should provide details (from the journal) on the unit's failure.

# Testing done
Tested with a transient issue of `kubelet.service` to connect to `quay.io`, making the service fail in some test runs and succeed in others.

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


t-lo:
Thanks for testing, will fix!

PR #114, opened 2020-07-02
azure: use one Resource Group for all tests

pothos:
Each kola cluster created and destroyed its own Resource Group which
took 2-3 minutes.
Reuse the image Resource Group if it was created for the flight and
run all tests in it. To keep the amount of left over VM network objects
small, the IP addresses and NICs are deleted when an instance is
deleted - this could also be removed again if we see that deleting
these objects takes a lot of time.

# How to use/testing done

```
./kola run -d --parallel 1 -k --platform azure --azure-blob-url 'https://flatcar.blob.core.windows.net/publish/flatcar-linux-2513.1.0-alpha.vhd?se=2099-12-31T23%3A59%3A59Z&sig=q93Hf6UggelIS3VQ%2F9sIjoz0CeREpX93o9tgTnRBrqk%3D&sp=rl&sr=b&sv=2015-02-21' --azure-location YOURLOCATION --azure-profile azureProfile.json --azure-auth service-credentials.json
```

### DIFF
--- a/platform/api/azure/instance.go
+++ b/platform/api/azure/instance.go
@@ -147,7 +147,10 @@ func (a *API) CreateInstance(name, userdata, sshkey, resourceGroup, storageAccou
 		return true, nil
 	})
 	if err != nil {
-		a.TerminateInstance(name, resourceGroup)
+		_, _ = a.compClient.Delete(resourceGroup, name, nil)
+		_, _ = a.intClient.Delete(resourceGroup, *nic.Name, nil)
+		_, _ = a.ipClient.Delete(resourceGroup, *ip.Name, nil)
+		// TODO: remove disk which doesn't get removed automatically
 		return nil, fmt.Errorf("waiting for machine to become active: %v", err)
 	}
 
@@ -174,9 +177,20 @@ func (a *API) CreateInstance(name, userdata, sshkey, resourceGroup, storageAccou
 	}, nil
 }
 
-func (a *API) TerminateInstance(name, resourceGroup string) error {
-	_, err := a.compClient.Delete(resourceGroup, name, nil)
-	return err
+// TerminateInstance deletes a VM created by CreateInstance with the public IP address and
+// NIC created for it. Currently it does not delete the OS disk that is created (see TODO).
+func (a *API) TerminateInstance(machine *Machine, resourceGroup string) error {
+	if _, err := a.compClient.Delete(resourceGroup, machine.ID, nil); err != nil {
+		return err
+	}
+	if _, err := a.intClient.Delete(resourceGroup, machine.InterfaceName, nil); err != nil {
+		return err
+	}
+	if _, err := a.ipClient.Delete(resourceGroup, machine.PublicIPName, nil); err != nil {
+		return err
+	}
+	// TODO: remove disk which doesn't get removed automatically
+	return nil
 }
 
 func (a *API) GetConsoleOutput(name, resourceGroup, storageAccount string) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


platform/api/azure/instance.go
```diff
@@ -147,7 +147,10 @@ func (a *API) CreateInstance(name, userdata, sshkey, resourceGroup, storageAccou
 		return true, nil
 	})
 	if err != nil {
-		a.TerminateInstance(name, resourceGroup)
```
marga-kinvolk, line 150: You modified TerminateInstance to do what the code here does... So I think you should still call TerminateInstance here?

platform/api/azure/instance.go
```diff
@@ -174,9 +177,18 @@ func (a *API) CreateInstance(name, userdata, sshkey, resourceGroup, storageAccou
 	}, nil
 }
 
-func (a *API) TerminateInstance(name, resourceGroup string) error {
-	_, err := a.compClient.Delete(resourceGroup, name, nil)
-	return err
+func (a *API) TerminateInstance(machine *Machine, resourceGroup string) error {
```
marga-kinvolk, line 180: Could you please add a function documentation comment as to what resources do and don't get terminated here?

platform/machine/azure/cluster.go
```diff
@@ -85,8 +85,10 @@ func (ac *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
 
 func (ac *cluster) Destroy() {
 	ac.BaseCluster.Destroy()
-	if e := ac.flight.api.TerminateResourceGroup(ac.ResourceGroup); e != nil {
-		plog.Errorf("Deleting resource group %v: %v", ac.ResourceGroup, e)
+	if ac.ResourceGroup != ac.flight.ImageResourceGroup {
```
marga-kinvolk, line 88: Can you add a comment here explaining what the flight.ImageResourceGroup is and why the conditional makes sense? This probably should be a function documentation comment as well, indicating what will Destroy() destroy.

platform/machine/azure/flight.go
```diff
@@ -146,19 +152,25 @@ func (af *flight) NewCluster(rconf *platform.RuntimeConfig) (platform.Cluster, e
 		ac.sshKey = af.FakeSSHKey
 	}
 
-	ac.ResourceGroup, err = af.api.CreateResourceGroup("kola-cluster")
-	if err != nil {
-		return nil, err
-	}
+	if af.ImageResourceGroup != "" && af.ImageStorageAccount != "" {
```
marga-kinvolk, line 155: If I understand this change correctly, the idea is that this is initialized the first time that this function is called, and after that it keeps re-using these values, is that correct? If so, could you add this to the function doc? Maybe "If no resource groups and storage accounts have been created for this flight yet, this will also create them" ?

platform/api/azure/instance.go
pothos: I thought so but the machine object is not created yet and the other option is to have three arguments in the TerminateInstance function.

platform/api/azure/instance.go
pothos: Thanks, done

platform/machine/azure/cluster.go
pothos: Done

platform/machine/azure/flight.go
pothos: The function is called at least for every test, and it is not initializing something to reuse it. The function either uses the flight's Resource Group or creates a new one.
Added the documentation.

marga-kinvolk: Thanks for the extra docs, it makes more sense now! :)

PR #115, opened 2020-07-03
cmd/plume: Add a prune command to clean up developer images

marga-kinvolk:
# cmd/plume: Add a prune command to clean up developer images

When running developer builds, plume uploads prerelease images to both Azure and AWS. This change adds a new "prune" subcommand for plume, to get rid of artifacts uploaded to old builds.

It also adds or improves logging messages on functions. This extra logging was needed for debugging while developing the code.

# How to use / Testing done

```
go1.14.4 build -o bin/plume github.com/coreos/mantle/cmd/plume
./bin/plume prune --channel developer --version none --azure-auth ../../.creds/azureCredentials.json --azure-profile ../../.creds/azureProfile.json --aws-credentials ../../.creds/aws-release-credentials --days 20 --debug
```

I've already run this for Azure and AWS with days=20, so to test it, it needs to be run with a smaller number or it won't do anything.

Needed for: kinvolk/PROJECT-flatcar-linux#294

### DIFF
--- a/cmd/plume/prune.go
+++ b/cmd/plume/prune.go
@@ -1,0 +1,229 @@
+// Copyright 2020 Kinvolk GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+
+	"github.com/coreos/mantle/platform/api/aws"
+	"github.com/coreos/mantle/platform/api/azure"
+)
+
+var (
+	days        int
+	pruneDryRun bool
+	cmdPrune    = &cobra.Command{
+		Use:   "prune --channel CHANNEL [options]",
+		Short: "Prune old release images for the given channel.",
+		Run:   runPrune,
+		Long:  `Prune old release images for the given channel.`,
+	}
+)
+
+func init() {
+	cmdPrune.Flags().IntVar(&days, "days", 30, "Minimum age in days for files to get deleted")
+	cmdPrune.Flags().StringVar(&awsCredentialsFile, "aws-credentials", "", "AWS credentials file")
+	cmdPrune.Flags().StringVar(&azureProfile, "azure-profile", "", "Azure Profile json file")
+	cmdPrune.Flags().StringVar(&azureAuth, "azure-auth", "", "Azure Credentials json file")
+	cmdPrune.Flags().StringVar(&azureTestContainer, "azure-test-container", "", "Use another container instead of the default")
+	cmdPrune.Flags().BoolVarP(&pruneDryRun, "dry-run", "n", false,
+		"perform a trial run, do not make changes")
+	AddSpecFlags(cmdPrune.Flags())
+	root.AddCommand(cmdPrune)
+}
+
+func runPrune(cmd *cobra.Command, args []string) {
+	if len(args) > 0 {
+		plog.Fatal("No args accepted")
+	}
+
+	// Override specVersion as it's not relevant for this command
+	specVersion = "none"
+
+	spec := ChannelSpec()
+	ctx := context.Background()
+	pruneAWS(ctx, &spec)
+	pruneAzure(ctx, &spec)
+}
+
+func pruneAzure(ctx context.Context, spec *channelSpec) {
+	if spec.Azure.StorageAccount == "" || azureProfile == "" {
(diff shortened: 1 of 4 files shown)


cmd/plume/prune.go
```diff
@@ -0,0 +1,192 @@
...
+		Long:  `Prune old release images.`,
+	}
+)
+
+func init() {
+	cmdPrune.Flags().IntVar(&days, "days", 30, "Minimum age in days for files to get deleted")
+	cmdPrune.Flags().StringVar(&awsCredentialsFile, "aws-credentials", "", "AWS credentials file")
+	cmdPrune.Flags().StringVar(&azureProfile, "azure-profile", "", "Azure Profile json file")
+	cmdPrune.Flags().StringVar(&azureAuth, "azure-auth", "", "Azure Credentials json file")
+	cmdPrune.Flags().StringVar(&azureTestContainer, "azure-test-container", "", "Use test container instead of default")
```
pothos, line 47: Maybe _use other container than the default_ would make it more clear?

cmd/plume/prune.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+func pruneAWS(ctx context.Context, spec *channelSpec) {
+	if spec.AWS.Image == "" || awsCredentialsFile == "" {
+		plog.Notice("AWS image pruning disabled.")
+		return
+	}
+
+	for _, part := range spec.AWS.Partitions {
```
pothos, line 135: I think it would be better to be able to specify one partition instead of iterating over all here.

pothos: Or maybe I'm missing that the channel and partition is passed as argument

pothos: Ah, ok, found it ;)

cmd/plume/prune.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2016 CoreOS, Inc.
```
pothos, line 1: I think we can remove this line or change it

marga-kinvolk: Indeed. Sorry about that, copy & paste mistake.

cmd/plume/prune.go
marga-kinvolk: Done

cmd/plume/prune.go
marga-kinvolk: I've added a comment to clarify this. Thanks.

cmd/plume/prune.go
```diff
@@ -0,0 +1,194 @@
...
+					if !pruneDryRun {
+						// Construct the s3ObjectPath in the same manner it's constructed for upload
+						var version string
+						for _, t := range image.Tags {
+							if *t.Key == "Version" {
+								version = *t.Value
+							}
+						}
+						imageFileName := spec.AWS.Image
+						s3ObjectPath := fmt.Sprintf("%s-usr/%s/%s", *image.Architecture, version, strings.TrimSuffix(imageFileName, filepath.Ext(imageFileName)))
```
pothos, line 180: `*image.Architecture` returns `x86_64` instead of `amd64` and needs some postprocessing.

cmd/plume/prune.go
```diff
@@ -0,0 +1,194 @@
...
+			now := time.Now()
+			for _, blob := range blobs {
+				lastModifiedDate, err := time.Parse(time.RFC1123, blob.Properties.LastModified)
+				if err != nil {
+					plog.Warningf("Error converting last modified date (%v): %v", blob.Properties.LastModified, err)
+				}
+				duration := now.Sub(lastModifiedDate)
+				daysOld := int(duration.Hours() / 24)
+				if daysOld >= days {
+					plog.Infof("Old blob %q: %d days old\n", blob.Name, daysOld)
```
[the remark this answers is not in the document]
pothos: Ah, and also check the channel, forgot that in the review (`--channel edge` currently still enumerates other channels like `flatcar-linux-1772.0.0-alpha.vhd`).

cmd/plume/prune.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/coreos/mantle/platform/api/aws"
+	"github.com/coreos/mantle/platform/api/azure"
+)
+
+var (
+	days        int
+	pruneDryRun bool
+	cmdPrune    = &cobra.Command{
+		Use:   "prune [options]",
+		Short: "Prune old release images.",
```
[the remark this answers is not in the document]
pothos: and maybe document that it takes `--channel` into account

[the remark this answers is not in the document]
marga-kinvolk: Yeah, I mentioned that in the how to test command (I was using `--version none`), but I figured that instead of documenting this brokenness, it's better to just override it. I've added a line to always set the version to none.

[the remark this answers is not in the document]
marga-kinvolk: I've documented the fact that it requires the channel now.

cmd/plume/prune.go
marga-kinvolk: Ok, I think I've fixed this although I don't think it matters because the S3 bucket is empty (I think whatever gets uploaded there gets deleted once it turns into a snapshot).

cmd/plume/prune.go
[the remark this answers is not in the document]
marga-kinvolk: Uhm... Ok. I had only tested it with the developer channel, which goes to a different bucket, so it's not relevant. I'll figure out a way of filtering the blobs per channel.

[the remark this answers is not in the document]
marga-kinvolk: Alright this was a bit more involved than expected, but it's working now. It checks that the stored metadata filename is the one from the spec, and also that the blob name includes the channel name being processed.

PR #117, opened 2020-07-20
sdk/repo/verify: Support verifying that a tag was checked out

pothos:
Besides commit IDs and branches, the manifest can specify a revision
as tag via refs/tags/TAG.
Support manifests with tags when verifying with cork that the
repositories are in the desired state.

# How to use

Create and build manifests with `refs/tags/TAG`

# Testing done

```
$EDITOR .repo/manifest.xml # modify the reference for afterburn to refs/tags/v0.1.0
cd src/third_party/afterburn
git checkout v0.1.0
cd -
 ~/kinvolk/mantle/bin/cork verify
# either works complains about something unrelated
cd -
git checkout flatcar-master
cd -
 ~/kinvolk/mantle/bin/cork verify
# complains with 2020-07-20T11:50:43Z sdk/repo: Project dir src/third_party/afterburn at tags [], expected refs/tags/v0.1.0 (ignore the refs/tags/ prefix)
```

### DIFF
--- a/sdk/repo/verify.go
+++ b/sdk/repo/verify.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/coreos/pkg/capnslog"
@@ -135,7 +136,7 @@ func isSHA1(s string) bool {
 	return err == nil && len(b) == sha1.Size
 }
 
-func (r *repo) projectBranches(p Project) ([]string, error) {
+func (r *repo) projectBranches(p Project) (map[string]struct{}, error) {
 	git := exec.Command("git", "show", "-s", "--pretty=%D", "HEAD")
 	git.Dir = filepath.Join(r.root, p.Path)
 	git.Stderr = os.Stderr
@@ -144,7 +145,8 @@ func (r *repo) projectBranches(p Project) ([]string, error) {
 		return nil, err
 	}
 
-	var branches []string
+	var exists struct{}
+	branches := make(map[string]struct{})
 
 	for _, val := range strings.Split(string(out), ",") {
 		ref := strings.TrimSpace(val)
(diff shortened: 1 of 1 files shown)


sdk/repo/verify.go
```diff
@@ -162,6 +162,38 @@ func (r *repo) projectBranches(p Project) ([]string, error) {
...
+	git := exec.Command("git", "show", "-s", "--pretty=%D", "HEAD")
+	git.Dir = filepath.Join(r.root, p.Path)
+	git.Stderr = os.Stderr
+	out, err := git.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	var tags []string
+
```
marga-kinvolk, line 175: Can you add sample output in a comment, to help understand what the code does?

sdk/repo/verify.go
```diff
@@ -162,6 +162,38 @@ func (r *repo) projectBranches(p Project) ([]string, error) {
...
+		if strings.HasPrefix(ref, "tag:") {
+			tagName := strings.Split(ref, " ")[1]
+			tags = append(tags, tagName)
+			continue
+		}
+
+		// In case "m/refs/tags/tagname" exists besides "refs/tags/tagname", handle any prefix
+		if strings.Contains(ref, "refs/tags/") {
+			tagName := strings.Split(ref, "refs/tags/")[1]
+			// The above breaks tags that itself include refs/tags/ but such tags are not expected
```
marga-kinvolk, line 188: You can use SplitN to avoid this (https://golang.org/pkg/strings/#SplitN)

sdk/repo/verify.go
```diff
@@ -162,6 +162,38 @@ func (r *repo) projectBranches(p Project) ([]string, error) {
...
+			tags = append(tags, tagName)
+			continue
```
marga-kinvolk, line 190: This continue is weird... Are you expecting to add more conditional clauses to the loop in the future?

sdk/repo/verify.go
```diff
@@ -232,10 +264,32 @@ func VerifySync(name string) error {
...
 			}
+		} else if strings.HasPrefix(project.Revision, "refs/tags/") {
+			tags, err := manifest.projectTags(project)
+			if err != nil {
+				return err
+			}
+
+			found := false
+
+			for _, tag := range tags {
```
marga-kinvolk, line 279: This block wouldn't be needed if the result from the function was a map of bool rather than a list. Given that you're creating it specially for this, wouldn't it make more sense to return a map?

sdk/repo/verify.go
```diff
@@ -232,10 +264,32 @@ func VerifySync(name string) error {
...
+				tagRef := "refs/tags/" + tag
+
+				if tagRef == project.Revision {
```
marga-kinvolk, line 282: I don't understand why you go through the trouble of removing the tags/refs/ part in the other function if you're going to add it here... Do you need the isolated tagname somewhere?

pothos: In the normal case this is just `"tag: v0.1.0"` and I didn't want to add it there. But maybe remove it here from `project.Revision` once in beginning.

sdk/repo/verify.go
pothos: Yes, or Go could have a function to check if a slice contains a value, instead of everybody rewriting it, let's wait for generics?

sdk/repo/verify.go
pothos: Thanks, good hint, I also added it for cutting the `tag: `.

sdk/repo/verify.go
pothos: Sure

sdk/repo/verify.go
pothos: Empty structs or leave this for loop?

marga-kinvolk: I don't care about empty structs vs boolean. I personally find the boolean clearer, but it might just be because I've been using that "set" implementation for a while.  Whatever you think is clearer.

I really prefer to avoid a useless for-loop, not because of optimization, but rather because the code is just harder to read with that for-loop in the middle.

sdk/repo/verify.go
```diff
@@ -162,6 +162,34 @@ func (r *repo) projectBranches(p Project) ([]string, error) {
...
+	}
+
+	// Extract tags from output like "HEAD, tag: v2051.99.1" or "HEAD -> flatcar-master, github/flatcar-master"
+	var tags []string
+
+	for _, val := range strings.Split(string(out), ",") {
+		ref := strings.TrimSpace(val)
+
+		if strings.HasPrefix(ref, "tag:") {
+			tagName := strings.SplitN(ref, " ", 2)[1]
```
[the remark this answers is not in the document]
pothos: Good idea

sdk/repo/verify.go
pothos: Rewrote it with maps but that needed the `reflect.ValueOf(tags).MapKeys()` trick to have a list of found tags to print out.

PR #122, opened 2020-08-20
kola/tests/locksmith: Next try to make coreos.locksmith.tls robust

pothos:
The certificate may be generated too early and thus have a wrong time.
Generate the certificate after systemd-timesync set the correct system
time.

# How to use

Run test with kola

# Testing done

None


t-lo (approved):
LGMT, thank you!

PR #123, opened 2020-08-20
kola/tests/locksmith: Wait for system to be synchronized

pothos:
The time-sync.target and time-set.target is not enough because the
actual syncing still takes place after the service is started:
  systemd[1]: Reached target System Time Set.
  systemd[1]: Reached target System Time Synchronized.
  […]
  systemd-timesyncd[810]: Initial synchronization to time server 199.102.46.78:123 (2.flatcar.pool.ntp.org).
Introduce a waiter to check for the synchronization.

# How to use

`./kola run -d --basename=jenkins-123 --esx-config-file esx.json --esx-ova-path flatcar_production_vmware_ova.ova  --platform=esx coreos.locksmith.tls`

# Testing done
 
Ran the above with the current Stable image.


kola/tests/locksmith/locksmith.go
```diff
@@ -64,7 +64,7 @@ etcd:
...
...s 3 -in /etc/ssl/certs/etcd-cert-self.pem -out /etc/ssl/certs/etcd-cert.pem\nExecStart=/usr/bin/openssl req -config /etc/ssl/etcd.cnf -x509 -nodes -newkey rsa:4096 -sha512 -days 3 -extensions etcd_ca -subj '/CN=locksmith CA' -out /etc/ssl/certs/ca-locksmith-cert.pem -keyout /etc/ssl/certs/ca-locksmith-key.pem\nExecStart=/usr/bin/openssl req -config /etc/ssl/etcd.cnf -x509 -nodes -newkey rsa:4096 -sha512 -days 3 -extensions etcd_client -subj '/CN=locksmith client' -out /etc/ssl/certs/locksmith-cert-self.pem -keyout /etc/ssl/certs/locksmith-key.pem\nExecStart=/usr/bin/openssl x509 -extfile /etc/ssl/etcd.cnf -CA /etc/ssl/certs/ca-locksmith-cert.pem -CAkey /etc/ssl/certs/ca-locksmith-key.pem -CAcreateserial -sha512 -days 3 -in /etc/ssl/certs/locksmith-cert-self.pem -out /etc/ssl/certs/locksmith-cert.pem\nExecStart=/usr/bin/chmod 0644 /etc/ssl/certs/ca-etcd-cert.pem /etc/ssl/certs/ca-etcd-key.pem /etc/ssl/certs/ca-locksmith-cert.pem /etc/ssl/certs/ca-locksmith-key.pem /etc/ssl/certs/etcd-cert.pem /etc/ssl/certs/etcd-key.pem /etc/ssl/certs/locksmith-cert.pem /etc/ssl/certs/locksmith-key.pem\nExecStart=/usr/bin/ln -fns ca-etcd-cert.pem /etc/ssl/certs/etcd.pem\nExecStart=/usr/bin/c_rehash"
```
t-lo, line 67: How about:
```
e=600; for i in $(seq $e); do echo Waiting for time sync $i/$e; timedatectl | grep -q "System clock synchronized: yes" && break; sleep 1; done
```

t-lo (changes requested):

kola/tests/locksmith/locksmith.go
pothos: Applied

PR #125, opened 2020-09-04
Update dependency to golang.org/x/text v0.3.3

dongsupark:
Update dependency `golang.org/x/text` to v0.3.3, mainly to address issue in its unicode package which could lead to the UTF-16 decoder entering an infinite loop, causing the program to crash or run out of memory.

See also https://groups.google.com/g/golang-announce/c/bXVeAmGOqz0 .


marga-kinvolk (approved):
LGTM

PR #126, opened 2020-09-07
Clarify SSH check error message for new machines

pothos:
The CheckMachine functions waits until an SSH connection runs
"systemctl is-system-running" with an output like "running" or
"degraded". However, the error message always suggested that SSH
would be unable to connect because it only stated "ssh unreachable".
Extend the error message to include the case of the system not being
ready and give more context for the "systemctl is-system-running"
output if it failed.

# How to use

Sometimes the error is encountered on Azure with: cl.etcd-member.discovery cl.locksmith.cluster cl.flannel.vxlan cl.flannel.udp

# Testing done

`make test` and reproduced the error for `cl.flannel.udp`:
```
--- FAIL: cl.flannel.udp (1527.51s)
        harness.go:519: Cluster failed starting machines: machine "jenkins-azure-89872ce22c" failed basic checks: ssh unreachable or system not ready: failure checking if machine is running: systemctl is-system-running returned stdout: "starting", stderr: "", err: Process exited with status 1, systemctl list-jobs returned stdout: "JOB UNIT                        TYPE  STATE  \n453 flanneld.service            start waiting\n454 flannel-docker-opts.service start waiting\n353 multi-user.target           start waiting\n464 docker.service              start waiting\n437 etcd-member.service         start running\n\n5 jobs listed.", stderr: "", err: <nil>

```

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -328,16 +328,22 @@ func CheckMachine(ctx context.Context, m Machine) error {
 		}
 		out, stderr, err := m.SSH("systemctl is-system-running")
 		if !bytes.Contains([]byte("initializing starting running stopping"), out) {
-			return nil // stop retrying if the system went haywire
+			return nil // stop retrying if the system went haywire, e.g., "degraded"
 		}
+		jobs := ""
+		if bytes.Contains([]byte("starting"), out) {
+			startingOut, startingStderr, startingErr := m.SSH("systemctl list-jobs")
+			jobs = fmt.Sprintf(", systemctl list-jobs returned stdout: %q, stderr: %q, err: %v", startingOut, startingStderr, startingErr)
+		}
+		// For "running" the exit code is 0 thus err is nil but not for, e.g., "starting" where the exit code is 1
 		if err != nil {
-			return fmt.Errorf("could not check if machine is running: %s: %v: %s", out, err, stderr)
+			return fmt.Errorf("failure checking if machine is running: systemctl is-system-running returned stdout: %q, stderr: %q, err: %v%s", out, stderr, err, jobs)
 		}
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


platform/platform.go
```diff
@@ -328,16 +328,17 @@ func CheckMachine(ctx context.Context, m Machine) error {
 		}
 		out, stderr, err := m.SSH("systemctl is-system-running")
 		if !bytes.Contains([]byte("initializing starting running stopping"), out) {
-			return nil // stop retrying if the system went haywire
+			return nil // stop retrying if the system went haywire, e.g., "degraded"
 		}
+		// For "running" the exit code is 0 thus err is nil but not for, e.g., "starting" where the exit code is 1
 		if err != nil {
-			return fmt.Errorf("could not check if machine is running: %s: %v: %s", out, err, stderr)
+			return fmt.Errorf("checking if machine is running: systemctl is-system-running: \"%s\". %v: \"%s\"", out, err, stderr)
```
marga-kinvolk, line 335: Tip: %q already quotes the string.
I'm not sure this message is clear enough, given how many strings are being concatenated... I think we need to be more verbose about what each string is:

Failure while checking if machine is running: systemctl is-system-running returned stdout: %q, stderr: %q, err: %v

pothos: I thought about `%q` but it also turns newlines to `\n`.

The message can also be about SSH connection errors.

marga-kinvolk: Sorry, I guess the string should be:
failure checking if machine is running: systemctl is-system-running returned stdout: %q, stderr: %q, err: %v

pothos: Thanks, I also tried to include the output of `systemctl list-jobs` now if `starting` is the last state.

marga-kinvolk (approved):
LGTM

PR #129, opened 2020-09-09
Jenkins: Build with Go 1.15

pothos:
# How to use

Build mantle in Jenkins.

# Testing done

```
docker run --rm -it -v `pwd`:/usr/src/myapp -w /usr/src/myapp -e CGO_ENABLED=1 golang:1.15 ./build
docker run --rm -it -v `pwd`:/usr/src/myapp -w /usr/src/myapp -e CGO_ENABLED=1 golang:1.15 ./test
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,11 +21,11 @@ node('amd64 && docker') {
     }
 
     stage('Build') {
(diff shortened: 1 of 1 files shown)


marga-kinvolk (changes requested):
I don't think this is a good idea. Go1.15 is breaking our builds with a signature verification that we can't fulfill. We shouldn't move to it until we can fulfill it.

marga-kinvolk (approved):
The issue with go1.15 is fixed, this can be merged in now.

PR #133, opened 2020-09-18
plume: Add a private flag to spec to maintain a private bucket

sayanchowdhury:
# plume: Add a private flag to spec to maintain a private bucket

# How to use

```
./bin/plume --private
```


marga-kinvolk (approved):
LGTM

PR #134, opened 2020-09-21
Don't check for the version file if the bucket is a private one

sayanchowdhury:
# Don't check for the version file if the bucket is a private

While doing the sanity check, don't check if the bucket is private.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -49,7 +49,8 @@ var (
 		RunE:  runPreRelease,
 	}
 
-	platforms = map[string]platform{
+	privateBucketSuffix = "private"
+	platforms           = map[string]platform{
 		"aws": platform{
 			displayName: "AWS",
 			handler:     awsPreRelease,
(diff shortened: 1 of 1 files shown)


cmd/plume/prerelease.go
```diff
@@ -182,7 +183,8 @@ func runCLPreRelease(cmd *cobra.Command) error {
 	}
 
 	// Sanity check!
-	if vertxt := src.Object(src.Prefix() + "version.txt"); vertxt == nil {
+	vertxt := src.Object(src.Prefix() + "version.txt")
+	if strings.Contains(spec.SourceURL(), privateBucketSuffix) == false && vertxt == nil {
```
pothos, line 187: Nit: Since it's not a suffix when used with `Contains`, I would do:
```suggestion
	if !strings.Contains(spec.SourceURL(), "private") && vertxt == nil {
```

sayanchowdhury: The variable is to be more explicit, and to explain the item we are searching for is the suffix for private buckets.

marga-kinvolk (approved):
LGTM

pothos (approved):

PR #135, opened 2020-09-21
plume: Rename the images to use pro instead of premium

sayanchowdhury:
# plume: Rename the images to use pro instead of premium

Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>


marga-kinvolk (approved):
LGTM

PR #136, opened 2020-09-22
Add LTS channel to plume and kola

pothos:
This makes kola accept the `lts` channel parameter for test filtering and also registers the LTS channel to plume for Azure/AWS but doesn't make AMIs public.

# How to use

Use with Jenkins

# Testing done

None

### DIFF
--- a/cmd/plume/containerlinux.go
+++ b/cmd/plume/containerlinux.go
@@ -179,6 +179,28 @@ var (
 			),
 			AWS: newAWSSpec(),
 		},
+		"lts": channelSpec{
+			BaseURL:        "gs://flatcar-jenkins-private/lts/boards",
+			BasePrivateURL: "gs://flatcar-jenkins-private/lts/boards",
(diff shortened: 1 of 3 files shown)


cmd/plume/containerlinux.go
```diff
@@ -141,6 +141,19 @@ var (
...
 		},
+		"lts": channelSpec{
+			BaseURL:      "gs://flatcar-jenkins-private/lts/boards",
+			Boards:       []string{"amd64-usr"},
+			Destinations: []storageSpec{},
+			GCE:          gceSpec{},
+			Azure: newAzureSpec(
+				azureEnvironments,
+				"publish",
+				"Flatcar LTS",
```
sayanchowdhury, line 152: There is another category field that is added for Pro/Regular

pothos: Thanks for catching this, rebased and updated

sayanchowdhury (approved):

PR #144, opened 2020-10-27
kola: fix S3 URLs for ignition tests

dongsupark:
Since the original S3 bucket `kola-fixtures` was renamed to `rh-kola-fixtures`, we are not able to use the bucket any more.
So we have set up a Flatcar-specific bucket for kola tests. `kola-fixtures` in the `eu-central-1` region.

Those can be accessed via
https://kola-fixtures.s3.eu-central-1.amazonaws.com/resources/...

## Testing done

TBD

### DIFF
--- a/kola/tests/ignition/resource.go
+++ b/kola/tests/ignition/resource.go
@@ -127,7 +127,7 @@ func init() {
 			      "filesystem": "root",
 			      "path": "/var/resource/http",
 			      "contents": {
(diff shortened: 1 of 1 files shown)


t-lo:
Looks good in general, let's go ahead and do a test run.

dongsupark:
A test run passed.
Merging.

PR #146, opened 2020-11-18
packet: Allow to set explicit URLs for the kernel and cpio files

pothos:
If appending the default file names to the base URL is not resulting in
the correct URL for the kernel and cpio files, there was no way to
specify the correct URLs for both explicitly.
Add two new options that default to the previous behavior if not used.

# How to use/testing done

```
bin/ore packet --api-key X --project X --storage-url X --gs-json-key X create-device --installer-image-kernel-url X --installer-image-cpio-url X --userdata-file X --hostname X
```


vbatts:
rebased, to pull in workflow actions

PR #148, opened 2020-11-24
Remove Fedora and Fcos code from plume

marga-kinvolk:
# Remove Fedora and Fcos code from plume

Plume's code currently includes a bunch of functions to upload Fedora and Fcos images that we're not using and that gets in the way when making changes to plume.  This PR removes all Fedora and Fcos functions from our code, keeping only the code that we're using for Flatcar.

It also replaces CoreOS strings with Flatcar (in command descriptions and comments, not copyright notices or github paths).

# How to use / Testing done

I've only verified that the code builds, with: `go1.15.1 build -o bin/plume github.com/coreos/mantle/cmd/plume`, I'm not sure how to test this properly without causing a release.


pothos (approved):
Looks good

PR #154, opened 2020-12-11
kola/tests/misc/network.go: Disallow the CRI plugin to listen on TCP

pothos:
The containerd cri plugin used a TCP port in the past and for Edge it
was allowed to be opened.
Remove this rule because containerd should not listen on TCP anymore
(and for the Alpha, Beta, Stable channels never did).

# How to use

# Testing done


t-lo (approved):
You [beat me by a whole hour](https://github.com/kinvolk/mantle/commit/99b87d6279e640ccb34e400fe8336efa4d4a1707), great work! LGTM.

PR #157, opened 2020-12-16
cmd/plume: Don't try to publish GCE LTS images

pothos:
This does not work at the moment and will only be done for GCE Pro.


marga-kinvolk (approved):
LGTM

PR #158, opened 2021-01-05
kola: Add a filter to run tests based on offering

sayanchowdhury:
# kola: Add a filter to run tests based on offering

With the introduction of the Pro offering, it is not possible to
test Pro specific features or test just for Pro without them runing
on the basic offering as well. This changes introduces a new flag
that helps building a clear distinction between Pro and the nornal
offering.

Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

# How to use

```sh
./build kola
./bin/kola run --offering=pro
```

# Testing done

Not done yet.


pothos (approved):
Tested with `bin/kola list --filter --offering pro --platform qemu` after adding some annotations to tests

sayanchowdhury:
I've updated tests with the Offering param so that it is easier and cleaner in jenkins to just pass the parameter `basic` or `pro`.

kola/register/register.go
```diff
@@ -58,6 +58,8 @@ type Test struct {
 	ExcludeDistros   []string // blacklist of distributions to ignore -- defaults to none
 	Channels         []string // whitelist of channels to run test against -- defaults to all
 	ExcludeChannels  []string // blacklist of channels to ignore -- defaults to none
+	Offerings        []string // blacklist of offerings to run the test against -- defaults to all
```
marga-kinvolk, line 61: This one should be whitelest

kola/tests/coretest/core.go
```diff
@@ -72,7 +72,8 @@ func init() {
 			"Useradd":          TestUseradd,
 			"MachineID":        TestMachineID,
 		},
-		Distros: []string{"cl"},
+		Distros:   []string{"cl"},
+		Offerings: []string{"basic", "pro"},
```
marga-kinvolk, line 76: I'm a bit confused here, do we really need to specify this field for each and every test? We don't do this for Architectures or Channels... Why do we need to do it for Offerings?

kola/register/register.go
sayanchowdhury: Done.

kola/tests/coretest/core.go
sayanchowdhury: I've updated `Offerings` similar to how `Architectures` and `Channels` are. I was thinking for a future case, where we have more offers, and miscalculated. Sorry for the confusion.

marga-kinvolk (approved):
LGTM

PR #161, opened 2021-03-18
kola/tests/flannel: use docker0's interface address as destination

pothos:
In case flanneld restarted, it does not work anymore to ping its
bridge address .0/32 attached to the flannel0 interface but only the
other addresses like .1/24 on docker0. This is the case with rkt
already but now with docker we have an additional restart of flanneld
when docker gets restarted to apply the flannel settings. Now this
problem directly pops up but before it would normally not happen in
the test.

# How to use/testing done

Run kola test `cl.flannel.udp` on Equinix Metal or another cloud platform (not QEMU)


marga-kinvolk (approved):
LGTM

PR #162, opened 2021-03-19
Update Kubernetes test

pothos:
- kola/tests/kubernetes: do not require Flannel to start before Docker
    Now that flannel itself is started with Docker, we can't start Flannel
    before Docker which results in a deadlock but instead have to rely on
    Flannel to restart Docker to apply its settings.
- kola/tests/kubernetes: ensure /run/flannel exists and simplify setup
    Placing the symlink in /run/flannel/ could fail if the directory did
    not exist before. Instead of setting up a symlink, directly write the
    template to /run/flannel/ which takes care of creating the directory.
- kola/tests/kubernetes: use coreos-metadata to find the IP address
    The nodes were assumed to use coreos-cloudinit when an empty
    configuation is given and in result, set up /etc/environment.
    Use coreos-metadata (afterburn) explicitly to find the IP address
    on any cloud environment. Also, add a few more comments about the
    set up and force that no local etcd can start on the K8s controller.
- kola/tests/kubernetes: switch to 192.168.0.0/16 network
    The 10.x.0.0/16 networks overlapped with the private network 10.0.0.0/8
    used in Equinix Metal and can also overlap with Digital Ocean private
    IP addresses. The 192.168.0.0/16 network seems to be not used by AWS
    nor the internal Docker network at the moment.
- kola/tests/kubernetes: test on more platforms
    The tests were only running on GCE and by enabling it on more platforms
    we have a better chance to catch errors early.
    For VMware in our setup there are 4 tests in parallel for 12 slots and
    thus we have to limit the cluster size to 3 nodes.
- kola/tests/kubernetes: drop kubelet-wrapper test
    The kubelet-wrapper script will be removed.

Can be done in a follow-up PR:
TODO: Replace running `kubelet-wrapper` with Docker, remove supporting `rkt` as K8s runtime
TODO: Try to fix running on `esx` and `packet` platforms

# How to use/testing done

Tested on AWS, DO, GCE


t-lo (approved):
LGTM, thank you.

pothos:
The test should still work with rkt instead of docker, otherwise we can add `ExcludeChannels: []string{"beta", "stable", "lts"},` and remove each channel when a > 2801 release comes in it.

PR #163, opened 2021-03-23
cmd/cork: Allow to apply a patch on top of the manifest references

pothos:
Sometimes the git branches the manifest refers to should not be
changed but still a modification introduced. This can now be done
in the form of a patch file for each of the three repositories that
define what gets build. A single patch file is taken but "git am"
is able to process the concatenation of multiple files well. The
three-way merge option is set to be able to use slightly outdated
patches as long as they get merged automatically by git.

# How to use/testing done

I created dummy patch files with `git format-patch -1` on top of `main` in each of my own external repositories and then updated an existing SDK to apply these patches after processing the manifest for `main`:
```
./cork update --scripts-patch ~/kinvolk/flatcar-scripts/0001-test.patch --overlay-patch ~/kinvolk/coreos-overlay/0001-sys-apps-baselayout-enable-pam_faillock-with-relaxed.patch --portage-patch ~/kinvolk/portage-stable/0001-test-portage.patch 
```
The patches got applied and persisted even after the command terminated.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -74,6 +74,10 @@ var (
 
 	verifyKeyFile string
 
+	scriptsPatch string
+	portagePatch string
+	overlayPatch string
+
 	setupCmd = &cobra.Command{
 		Use:   "setup",
 		Short: "Setup the system for SDK use",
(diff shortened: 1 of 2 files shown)


sayanchowdhury (approved):
LGTM!

PR #167, opened 2021-04-13
kola: provide internet access to qemu VM

tormath1:
In this PR, we allow QEMU VMs to access Internet - it was preventing some tests to run because of this dependency.

# How to use

Run kola against tests requiring Internet access, example:
```
$ sudo ./bin/kola run -b cl -p qemu --qemu-image ...  cl.rkt.etcd3
=== RUN   cl.rkt.etcd3
qemu-img: warning: Deprecated use of backing file without explicit backing format (detected format of qcow2)
--- PASS: cl.rkt.etcd3 (27.66s)
PASS, output in _kola_temp/qemu-2021-04-13-1711-9135
```

# What's missing

- ~update readme with more information~ the user does not have to perform any manual action. They are done in https://github.com/kinvolk/mantle/pull/167/commits/3d69f85f3b033dd389f05e83fb23c6f9ce7a5e59
- [x] add a teardown method (to remove what kola created)
- [x] run all tests with `RequiresInternetAccess` flag and remove this flag


platform/local/dnsmasq.go
```diff
@@ -108,6 +111,63 @@ func newInterface(s byte, i uint16) *Interface {
...
+	return nil
+}
+
+// generateVethPair creates and returns a map holding
+// a pair of veth based on the pid
+// where:
+// [0] is the network namespace veth
+// [1] is the host namespace veth
+func generateVethPair(pid int) []string {
+	x := pid%(31-16) + 16
```
pothos, line 135: This uses the last bits twice for both x and y. You can either divide in one case to shift the bits right or use a bitmask with `&` instead of `%`.

platform/local/dnsmasq.go
```diff
@@ -108,6 +111,63 @@ func newInterface(s byte, i uint16) *Interface {
...
+	y := pid % 255
```
pothos, line 136:
```suggestion
	y := pid % 256
```
255 is still fine, or?

platform/local/dnsmasq.go
pothos, line 135:
```suggestion
	x := (pid/256)%(31-16) + 16
```
here the division approach

platform/local/dnsmasq.go
```diff
@@ -147,6 +207,85 @@ func newSegment(s byte) (*Segment, error) {
 		return nil, fmt.Errorf("LinkSetUp() failed: %v", err)
 	}
 
+	// we first create a virtual ethernet pair in current network namespace
+	// based on the current process ID
+	pid := os.Getpid()
+	pair := generateVethPair(pid)
```
pothos, line 213: Since we don't use all bits of the PID there still can be clashes. Instead of erroring out, can we wrap this in a loop to detect a clash and retry? Another option coming to my mind is to create another global OS object like a TCP socket where the listening port is unique and has the right bit size for the values we want to have unique. With `0` as port the kernel will find a free port. E.g., using the two bytes of the TCP port means the address can be calculated as `172.(0001 and first half byte).(second half byte and third half byte).(fourth half byte and either 0000 or 0001)`.

platform/local/dnsmasq.go
```diff
@@ -147,6 +207,85 @@ func newSegment(s byte) (*Segment, error) {
...
+	peer, err = netlink.LinkByName(peer1[0])
+	if err != nil {
+		return nil, fmt.Errorf("unable to get link by name: %w", err)
+	}
+
+	if err := setupLink(peer, peer1[1], false); err != nil {
+		return nil, fmt.Errorf("unable to set up link: %w", err)
+	}
+
+	current, err := netns.GetFromName("kola-test")
```
pothos, line 280: Does this name conflict with other kola instances running on the host?

system/ns/enter.go
```diff
@@ -15,32 +16,25 @@
...
 	"github.com/vishvananda/netns"
 )
 
-// NsEnter locks the current goroutine the OS thread and switches to a
-// new network namespace. The returned function must be called in order
-// to restore the previous state and unlock the thread.
+// NsEnter switches to a new network namespace. The returned function must be called in order
+// to restore the previous state
 func Enter(ns netns.NsHandle) (func() error, error) {
-	runtime.LockOSThread()
```
pothos, line 27: I think this has to stay: https://golang.org/pkg/runtime/#LockOSThread
Otherwise it seems that there is no guarantee in which network namespace the Go routine will be.

system/ns/enter.go
```diff
@@ -52,15 +46,12 @@ func Enter(ns netns.NsHandle) (func() error, error) {
...
-
 	origns, err := netns.Get()
 	if err != nil {
 		return netns.None(), err
 	}
 	defer origns.Close()
 	defer netns.Set(origns)
 
-	return netns.New()
+	return netns.NewNamed("kola-test")
```
pothos, line 56: This creates a single global namespace, or? Then we couldn't run many kola instances in parallel.

Do you think there is a way to keep the functions as they are and try to rather change where the new network namespace gets created and pass its handle along so that `.Enter()` can be used?

platform/local/dnsmasq.go
```diff
@@ -147,6 +207,85 @@ func newSegment(s byte) (*Segment, error) {
...
+	if err := configureNAT(peer0[0]); err != nil {
+		return nil, fmt.Errorf("unable to configure NAT: %w", err)
+	}
+
+	root, err := netns.GetFromPid(1)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get NS from PID 1: %w", err)
+	}
+
+	if err := netns.Set(root); err != nil {
```
pothos, line 267: Seems this is not guaranteed due to how Go routines are scheduled and suspendable. Instead, the original version of `.Enter()` should be used to switch and go back.

system/ns/enter.go
tormath1: It was for testing purposes - using named namespace, it's easier to `nsenter`.

platform/local/dnsmasq.go
pothos: This may help: http://jodies.de/ipcalc?host=172.16.0.0&mask1=31&mask2=

system/ns/enter.go
pothos: Ah, ok, then without it one has to use the PID: `nsenter --net -t PID`

tormath1:
@pothos thanks for the review - comments have been addressed. :)

platform/local/dnsmasq.go
```diff
@@ -108,6 +112,70 @@ func newInterface(s byte, i uint16) *Interface {
...
+// [0] is the network namespace veth
+// [1] is the host namespace veth
+// [2] is the next hop for adding route
+func generateVethPair(seed int) []string {
+	// we basically use each single bit of the seed
+	// to generate an unique IP range
+	x := 0x10 | (seed & 0x0F)
+	y := ((seed >> 4 & 0x0F) << 4) | (seed >> 8 & 0x0F)
+	hostZ := ((seed >> 12 & 0x0F) << 4) | 0x00
+	nsZ := ((seed >> 12 & 0x0F) << 4) | 0x01
```
pothos, line 142: I get confused^^, maybe just a single shift and then masks:
```suggestion
	start := 0b10101100000100000000000000000000 // 172.16.0.0/12
	host := start | (seed <<1) | 0
	ns := host | 1
	x := host & 0xFF0000
	y := host & 0xFF00
	hostZ := host & 0xFF
	nsZ := ns & 0xFF
```

platform/local/dnsmasq.go
```diff
@@ -108,6 +112,70 @@ func newInterface(s byte, i uint16) *Interface {
...
+
+	baseIP := fmt.Sprintf("172.%d.%d", x, y)
+
+	topology := make([]string, 3)
+	topology[0] = fmt.Sprintf("%s|%s", fmt.Sprintf("kola-ns-%d%d", x, y), fmt.Sprintf("%s.%d/31", baseIP, nsZ))
```
pothos, line 147: maybe `-%d%d%d` with the additional `, hostZ` to not have the same name twice

platform/local/dnsmasq.go
```diff
@@ -147,6 +215,98 @@ func newSegment(s byte) (*Segment, error) {
...
 
+	// we first create an unique virtual ethernet pair in current network namespace accross the multiple kola instances
+	// we use linux network random port attribution to assert the uniqueness of the IP range in order to avoid IP
+	// range clashes
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return nil, fmt.Errorf("unable to listen on random port: %w", err)
+	}
+
+	defer listener.Close()
```
pothos, line 226: the `listener` needs to live on and only be closed when the network namespace is teared down

tormath1:
@pothos thanks for reviewing again - comments have been addressed. PR is now ready to be fully reviewed :) If it looks good for you, I will rebase my fixups and reword some commits since logic has changed on the road.

platform/local/dnsmasq.go
```diff
@@ -108,6 +118,74 @@ func newInterface(s byte, i uint16) *Interface {
...
+	return nil
+}
+
+// generateVethPair creates and returns a map holding
+// a pair of veth based on a seed
+// where:
+// [0] is the network namespace veth
+// [1] is the host namespace veth
+// [2] is the next hop for adding route
+func generateVethPair(seed int) []string {
```
pothos, line 142: maybe check that the seed value is <= 0xffff

platform/local/dnsmasq.go
```diff
@@ -147,6 +225,120 @@ func newSegment(s byte) (*Segment, error) {
 		return nil, fmt.Errorf("LinkSetUp() failed: %v", err)
 	}
 
+	// we first create an unique virtual ethernet pair in current network namespace accross the multiple kola instances
+	// we use linux network random port attribution to assert the uniqueness of the IP range in order to avoid IP
+	// range clashes
+	listener, err := net.Listen("tcp", ":0")
```
pothos, line 231: This needs to be done in the host network namespace (`rootExit, err := ns.Enter(root)` and leave again)

pothos (commented):
I think we should also remove the `RequiresInternetAccess` flag and `PlatformsNoInternet` definitions from `kola/register/register.go`. Or keep the `RequiresInternetAccess` flag untouched and just remove `qemu` from `PlatformsNoInternet`.

tormath1:
> I think we should also remove the `RequiresInternetAccess` flag and `PlatformsNoInternet` definitions from `kola/register/register.go`. Or keep the `RequiresInternetAccess` flag untouched and just remove `qemu` from `PlatformsNoInternet`.

@pothos I removed `RequiresInternetAccess` flag in (https://github.com/kinvolk/mantle/pull/167/commits/55ea9ee28ba247288b8adbb0e37ae886bea06956 and its parent commit) but not _all_ flag since we are hitting the issue :
```
=== RUN   cl.locksmith.cluster
--- FAIL: cl.locksmith.cluster (0.10s)
        harness.go:523: Cluster failed starting machines: rendering Container Linux config for platform "": error: platform must be specified to use templating
```

which should be tracked and addressed in another issue. It's better for now to keep this definitions then.

pothos:
The failing tests can be excluded with `ExcludePlatforms: []string{"qemu"},` until then

pothos:
`cl.locksmith.cluster` needs to be excluded

pothos:
Looks good, thanks!
The only thing I can think of is maybe checking if there is a `172.something` network already present like from Docker and not pick addresses in that range (maybe a retry loop). This is not a big problem but it could happen that a container IP address clashes and then the container can't access the network while kola runs…

tormath1:
@pothos I added the retry logic: we provide the list of the host networks to the `generate` function then we check if the veth generated host IP is in one of this networks.

platform/local/dnsmasq.go
```diff
@@ -108,6 +129,84 @@ func newInterface(s byte, i uint16) *Interface {
...
+	return nil
+}
+
+// generateVethPair creates and returns a map holding
+// a pair of veth based on a seed
+// where:
+// [0] is the network namespace veth
+// [1] is the host namespace veth
+// [2] is the next hop for adding route
+func generateVethPair(seed int, hostsIP []*net.IPNet) ([]string, error) {
```
pothos, line 153:
```suggestion
func generateVethPair(seed int, hostsNetworks []*net.IPNet) ([]string, error) {
```

platform/local/dnsmasq.go
```diff
@@ -108,6 +129,84 @@ func newInterface(s byte, i uint16) *Interface {
...
+	host |= seed
+	ns := host | 1
+
+	hostIP := make(net.IP, 4)
+	nsIP := make(net.IP, 4)
+
+	binary.BigEndian.PutUint32(hostIP, uint32(host))
+	binary.BigEndian.PutUint32(nsIP, uint32(ns))
+
+	for _, host := range hostsIP {
```
pothos, line 171:
```suggestion
	for _, network := range hostsNetworks {
```

platform/local/dnsmasq.go
```diff
@@ -108,6 +129,84 @@ func newInterface(s byte, i uint16) *Interface {
...
+		if host.Contains(hostIP) {
```
pothos, line 172:
```suggestion
		if network.Contains(hostIP) {
```

platform/local/dnsmasq.go
```diff
@@ -147,6 +249,175 @@ func newSegment(s byte) (*Segment, error) {
...
+
+			for _, address := range addresses {
+				ips = append(ips, address.IPNet)
+			}
+		}
+
+		pair, err = generateVethPair(p, ips)
+		if err != nil {
+			if errors.Is(err, ErrIPClash) {
+				isValid = false
```
pothos, line 310:
```suggestion
				isValid = false
				plog.Debugf("failed to use port %v as unique seed for the veth pair due to address clash, retrying", p)
```

pothos (approved):

PR #168, opened 2021-04-19
platform: allow CLC templating for dynamic IP address insertion

pothos:
-  platform: allow CLC templating for dynamic IP address insertion
    The tests that require coreos-metadata and CLC templating could not
run because for QEMU only a static replacement was done before the
CLC configuration was rendered.
Add support for the template variables by treating QEMU as a custom
platform for the transpiler, making use of a coreos-metadata service
that has the IP addresses hardcoded.

- platform/machine/*/cluster.go: update custom coreos-metadata service
    
    The coreos-metadata service in the Flatcar OEMs normally writes to
    a "flatcar" file and symlinks it.
    Do the same for the custom ESX and QEMU coreos-metadata services.

- platform/local/dnsmasq: hardcode DNS server for QEMU tests
    
    The system may have 127.0.0.53 as DNS server which is the
    systemd-resolved stub. This is no reachable from the dnsmasq's
    network namespace which has its own loopback interface.
    Hardcode the DNS servers to Cloudflare's and Google's to ensure
    DNS is working regardless of the host /etc/resolv.conf file.

- kola/tests/*: enable more tests for QEMU

# How to use/testing done

Run QEMU tests for
```
cl.locksmith.cluster cl.etcd-member.discovery cl.etcd-member.v2-backup-restore cl.etcd-member.etcdctlv3 cl.flannel.vxlan cl.flannel.udp coreos.ignition.security.tls
```

### DIFF
--- a/kola/tests/etcd/discovery.go
+++ b/kola/tests/etcd/discovery.go
@@ -38,7 +38,7 @@ func init() {
   initial_advertise_peer_urls: http://{PRIVATE_IPV4}:2380
   discovery:                   $discovery`),
 		Distros:          []string{"cl"},
-		ExcludePlatforms: []string{"qemu", "qemu-unpriv"},
+		ExcludePlatforms: []string{"qemu-unpriv"},
 	})
 
 	register.Register(&register.Test{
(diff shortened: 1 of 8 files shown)


tormath1 (approved):
LGTM ! Thanks :)

PR #171, opened 2021-04-26
kola: add kubeadm tests

tormath1:
In this PR, we add kubernetes tests using `kubeadm` and `ignition` to setup the cluster.

A `user-data` script will be uploaded through ignition on both worker and controller in order to run the required commands. `ignition` will take care of pulling required binaries / uploading various files (user-data, manifest).

Test will assert that there is two running nodes and it will deploy a nginx pod.

# How to use

```
./build kola
```
# Testing done

It has been tested on `qemu` platform with `alpha-2823.0.0`:

```bash
$ sudo ./bin/kola run -b cl -p qemu --qemu-image flatcar_production_qemu_image.img kubeadm.base
=== RUN   kubeadm.base
=== RUN   kubeadm.base/node_readiness
=== RUN   kubeadm.base/nginx_deployment
--- PASS: kubeadm.base (317.05s)
        cluster.go:117: Created symlink /etc/systemd/system/multi-user.target.wants/kubelet.service → /etc/systemd/system/kubelet.service.
        cluster.go:117: 	[WARNING Service-Docker]: docker service is not enabled, please run 'systemctl enable docker.service'
        cluster.go:117: 	[WARNING IsDockerSystemdCheck]: detected "cgroupfs" as the Docker cgroup driver. The recommended driver is "systemd". Please follow the guide at https://kubernetes.io/docs/setup/cri/
    --- PASS: kubeadm.base/node_readiness (52.85s)
    --- PASS: kubeadm.base/nginx_deployment (22.12s)
PASS, output in _kola_temp/qemu-2021-04-26-1847-14719
```

and on GCP with latest alpha (`alpha-2857.0.0`) (note `--gce-machinetype=n1-standard-2`):
```
./bin/kola run --basename=${BASENAME} \
    --gce-image=${IMAGE_NAME} \
    --gce-json-key=${GCE_JSON_KEY} \
    --parallel=4 --gce-machinetype=n1-standard-2 --platform=gce \
    ${TEST_NAME}
=== RUN   kubeadm.base
=== RUN   kubeadm.base/node_readiness
=== RUN   kubeadm.base/nginx_deployment
--- PASS: kubeadm.base (294.59s)
    --- PASS: kubeadm.base/node_readiness (25.84s)
    --- PASS: kubeadm.base/nginx_deployment (15.90s)
PASS, output in _kola_temp/gce-2021-04-30-1131-290038
```

# What's missing

- [x] try to use ignition instead of script to provision the nodes (worker and master)
- [x] test on other platforms
- [x] add some unit-tests maybe to test the logic
- [x] ~decide what we should do: should we merge kubeadm and kubernetes tests ? should we drop the current kubernetes tests in favor of this one ?~ let's keep them in a separate test, then we will remove slowly the kubernetes ones

### DIFF
--- a/kola/tests/kubeadm/templates.go
+++ b/kola/tests/kubeadm/templates.go
@@ -1,0 +1,384 @@
+// Copyright 2021 Kinvolk GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package kubeadm
+
+var (
+	workerConfig = `systemd:
+  units:
+    - name: prepare-cni-plugins.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Unpack CNI plugins to /opt/cni/bin
+        ConditionPathExists=!/opt/cni/bin
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        Environment=CNI_VERSION={{ .CNIVersion }}
+        ExecStartPre=/usr/bin/mkdir --parents /opt/cni/bin
+        ExecStartPre=/usr/bin/tar -v --extract --file "/opt/cni-plugins-linux-{{ .Arch }}-${CNI_VERSION}.tgz" --directory /opt/cni/bin --no-same-owner
+        ExecStart=/usr/bin/rm "/opt/cni-plugins-linux-{{ .Arch }}-${CNI_VERSION}.tgz"
+        [Install]
+        WantedBy=multi-user.target
+    - name: prepare-critools.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Unpack CRI tools to /opt/bin
+        ConditionPathExists=!/opt/bin/crictl
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        Environment=CRICTL_VERSION={{ .CRIctlVersion }}
+        Environment=DOWNLOAD_DIR={{ .DownloadDir}}
+        ExecStartPre=/usr/bin/mkdir --parents "${DOWNLOAD_DIR}"
+        ExecStartPre=/usr/bin/tar -v --extract --file "/opt/crictl-${CRICTL_VERSION}-linux-{{ .Arch }}.tar.gz" --directory "${DOWNLOAD_DIR}" --no-same-owner
+        ExecStart=/usr/bin/rm "/opt/crictl-${CRICTL_VERSION}-linux-{{ .Arch }}.tar.gz"
+        [Install]
+        WantedBy=multi-user.target
+storage:
+  files:
+    - path: /opt/cni-plugins-linux-{{ .Arch }}-{{ .CNIVersion }}.tgz
+      filesystem: root
+      mode: 0644
+      contents:
+        remote:
+          url: https://github.com/containernetworking/plugins/releases/download/{{ .CNIVersion }}/cni-plugins-linux-{{ .Arch }}-{{ .CNIVersion }}.tgz
+          verification:
+            hash:
+              function: sha512
+              sum: {{ .CNISum }}
+    - path: /opt/crictl-{{ .CRIctlVersion }}-linux-{{ .Arch }}.tar.gz
+      filesystem: root
+      mode: 0644
+      contents:
+        remote:
+          url: https://github.com/kubernetes-sigs/cri-tools/releases/download/{{ .CRIctlVersion }}/crictl-{{ .CRIctlVersion }}-linux-{{ .Arch }}.tar.gz
+          verification:
+            hash:
+              function: sha512
+              sum: {{ .CRIctlSum }}
+    - path: {{ .DownloadDir }}/kubeadm
+      filesystem: root
+      mode: 0755
+      contents:
+        remote:
+          url: https://storage.googleapis.com/kubernetes-release/release/{{ .Release }}/bin/linux/{{ .Arch }}/kubeadm
+          verification:
+            hash:
+              function: sha512
+              sum: {{ .KubeadmSum }}
+    - path: {{ .DownloadDir }}/kubelet
+      filesystem: root
+      mode: 0755
+      contents:
+        remote:
+          url: https://storage.googleapis.com/kubernetes-release/release/{{ .Release }}/bin/linux/{{ .Arch }}/kubelet
+          verification:
+            hash:
+              function: sha512
+              sum: {{ .KubeletSum }}
+    - path: /home/core/install.sh
+      filesystem: root
+      mode: 0755
+      contents:
+        remote:
+          url: "data:text/plain;base64,{{ .WorkerScript }}"`
+
+	masterConfig = `systemd:
+  units:
+    - name: prepare-cni-plugins.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Unpack CNI plugins to /opt/cni/bin
+        ConditionPathExists=!/opt/cni/bin
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        Environment=CNI_VERSION={{ .CNIVersion }}
+        ExecStartPre=/usr/bin/mkdir --parents /opt/cni/bin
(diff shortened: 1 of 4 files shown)


kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,205 @@
...
+var (
+	masterScript = `#!/bin/bash
+export CNI_VERSION={{ .CNIVersion }}
+export CRICTL_VERSION={{ .CRIctlVersion }}
+export RELEASE_VERSION={{ .ReleaseVersion }}
+export DOWNLOAD_DIR={{ .DownloadDir }}
+export RELEASE={{ .Release }}
+export PATH="${PATH}:${DOWNLOAD_DIR}"
+
+# this block redirects everything to /dev/null in order to print out only the final
```
pothos, line 25: This makes it hard to catch any `curl` errors. In general I would say `set -euo pipefail` should be the first thing that script does to run in a stricter mode.

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,205 @@
...
+prefix="https://"
+short_url=${URL#"${prefix}"}
+
+cat << EOF
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: JoinConfiguration
+discovery:
+  bootstrapToken:
+    apiServerEndpoint: ${short_url}
+    token: $(kubeadm token create)
```
pothos, line 129: errors of this subshell won't be caught, it's better to first run this line in a single assignment to a variable and then use this variable here

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,205 @@
...
+    caCertHashes:
+    - sha256:$(openssl x509 -pubkey -in /etc/kubernetes/pki/ca.crt | openssl rsa -pubin -outform der 2>/dev/null | openssl dgst -sha256 -hex | sed 's/^.* //')
```
pothos, line 131: (same)

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,205 @@
...
+        /etc/systemd/system/kubelet.service.d \
+	${HOME}/.kube \
+	/home/core/.kube
+   
+    # we download and install the various requirements
+    # * cni plugins
+    # * crictl
+    # * kubelet service and kubeadm dropin
+    # * kubelet, kubeadm and kubectl binaries
+    curl -sSL "https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-linux-amd64-${CNI_VERSION}.tgz" |
```
pothos, line 42: To make curl more robust one can add something like `--retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20`

pothos: and `--fail` also makes sense to report an error on 404

tormath1:
@pothos comments have been addressed. Ignition is also now used to download binaries file and also to upload "`user-data`" and nginx manifest.

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,384 @@
...
+    tee /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+
+systemctl enable --now kubelet
+
+cat << EOF > worker-config.yaml
+{{ .WorkerConfig }}
+EOF
+
+mainLink=$(ip route get 1 | grep -o -P ' dev .*? ' | cut -d ' ' -f 3)
+ipv4=$(ip -4 addr show "${mainLink}" | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
```
pothos, line 381:
```suggestion
systemctl start coreos-metadata
ipv4=$(cat /run/metadata/flatcar | grep -v IPV6 | grep IP | grep -E '(PRIVATE|LOCAL)' | cut -d = -f 2)
```
as in the existing K8s test

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,384 @@
...
+    systemctl enable --quiet --now kubelet
+    kubeadm config images pull
+    kubeadm init --config kubeadm-config.yaml
+    cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+    cp -i /etc/kubernetes/admin.conf /home/core/.kube/config
+    chown -R core:core /home/core/.kube; chmod a+r /home/core/.kube/config;
+
+    kubectl create -f https://docs.projectcalico.org/manifests/tigera-operator.yaml
+    kubectl apply -f calico.yaml
+} > /dev/null 2>&1
```
pothos, line 312: What is the purpose of this silencing? It will make debugging harder

tormath1: In this [lines](https://github.com/kinvolk/mantle/blob/32d359014c9422906b4e22871b86b399b11dd646/kola/tests/kubeadm/kubeadm.go#L169-L176), we execute the install script - the output of this script is the generated worker config. That said, an alternative would be to upload this [section](https://github.com/kinvolk/mantle/blob/32d359014c9422906b4e22871b86b399b11dd646/kola/tests/kubeadm/templates.go#L315-L334) as inline content in the ignition process then to execute it _after_ the `install` script.

pothos: Ok, didn't have a close look, but in this case it should be possible to only redirect stdout to /dev/stderr, or?

tormath1: it works like a charm - I've certainly been a bit too extreme with this redirect.

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,384 @@
...
+    - path: /opt/cni-plugins-linux-amd64-{{ .CNIVersion }}.tgz
+      filesystem: root
+      mode: 0644
+      contents:
+        remote:
+          url: https://github.com/containernetworking/plugins/releases/download/{{ .CNIVersion }}/cni-plugins-linux-amd64-{{ .CNIVersion }}.tgz
+          verification:
+            hash:
+              function: sha512
+              sum: a8fd7ffce7f4f1fb00ac2e4840b7961efc3ec844c5e99035c72baa95f038b316552ce71f8a17f82c04d614da490689953a7a80008d7b9be563187673813934b5
```
pothos, line 63: Just an idea, maybe the checksum should also be a variable when the version is one

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,384 @@
...
+        nodeSelector: all()
+  flexVolumePath: /opt/libexec/kubernetes/kubelet-plugins/volume/exec/
+EOF
+
+{
+    systemctl enable --quiet --now kubelet
+    kubeadm config images pull
+    kubeadm init --config kubeadm-config.yaml
+    cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+    cp -i /etc/kubernetes/admin.conf /home/core/.kube/config
```
pothos, line 307: These two lines are the same, or?

pothos: `cp -i` means `--interactive` and is probably not the right behavior here?

kola/tests/kubeadm/templates.go
tormath1: I thought the same last week.
Having something like a `binary` struct 
```go
type Binary struct {
    Name string
    Version string
    SHA512 string
}
```

would be nice if tomorrow we want to test other versions

kola/tests/kubeadm/templates.go
tormath1: yeah it runs as root, I took the same way as the current `kubernetes` tests. Do you think it's better to run everything as a regular user and to invoke `sudo` when required ? I don't have a strong opinion on this test

pothos: I think it's more normal to run most things as `core` and add `sudo` where necessary but if in this case we would need to add `sudo` add too many places it's not worth it and one could do it the other way round if wanted, e.g., `sudo -u core kubectl`.

pothos: The `cp -i` is the only change I see which would be good to do before merging, the rest is not so important.

kola/tests/kubeadm/templates.go
```diff
@@ -0,0 +1,384 @@
...
+        [Unit]
+        Description=Unpack CNI plugins to /opt/cni/bin
+        ConditionPathExists=!/opt/cni/bin
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        Restart=on-failure
+        Environment=CNI_VERSION={{ .CNIVersion }}
+        ExecStartPre=/usr/bin/mkdir --parents /opt/cni/bin
+        ExecStartPre=/usr/bin/tar -v --extract --file "/opt/cni-plugins-linux-amd64-${CNI_VERSION}.tgz" --directory /opt/cni/bin --no-same-owner
```
pothos, line 31: bonus: make `amd64` a variable set from kola or maybe dynamically (either using `$(source /etc/os-release; echo $FLATCAR_BOARD | cut -d '"' -f 2 | cut -d - -f 1)` or `$(if [ $(uname -m) = x86_64 ]; then echo amd64; else echo arm64; fi)`)

tormath1:
@pothos thanks for the review - comments have been addressed !

PR #177, opened 2021-06-04
kola/docker: add selinux test

tormath1:
we create a simple scenario where the user would run a docker container
with a shared volume within SELinux context.

## How to use

:warning: can't be run with current Flatcar release as it needs SELinux updates

## Testing done

```
sudo ./bin/kola run -b cl -p qemu --qemu-image flatcar_production_qemu_image.img docker.selinux
```

closes https://github.com/kinvolk/Flatcar/issues/304

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -54,6 +54,14 @@ type simplifiedDockerInfo struct {
 
 func init() {
 	register.Register(&register.Test{
+		Run:           dockerSELinux,
+		ClusterSize:   1,
+		Name:          "docker.selinux",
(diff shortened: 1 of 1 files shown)


tormath1:
@kinvolk/flatcar-maintainers if you have a few minutes to review this - it includes a basic SELinux test for Docker. :) I force pushed to include `beta` to supported versions.

jepio (approved):
Looks good. Would it be possible to directly skip on arm64 as long as we don't have selinux there?

tormath1:
@jepio thanks for the review ! `arm64` is now excluded from the test. Let's wait for the CI, then squash and merge.

PR #178, opened 2021-06-08
tests/cl.filesystem: skip "/run", fix test failure on arm64

wrl:
On arm64, we have a `/run/udev/data/+platform:Fixed MDIO bus.0`, which, due to the presence of spaces in the filename. Since `/run` is an ephemeral directory and we skip other directories (`/proc`, `/sys`) for that reason, it seems like a good idea to skip `/run` as well.


pothos (approved):

pothos:
Maybe fix the sentence by adding _gets flagged_ or what you thought should continue the sentence: _… spaces in the filename[, gets flagged]_

PR #179, opened 2021-06-15
kola/docker: make selinux optional

tormath1:
`selinux` support is going to be disabled on the next shipped Docker.

To keep compatibility with current releases, selinux is made optional:
we just assert that at least `seccomp` or `selinux` or both are
enabled.

this commit will be reverted once SELinux will be back on Flatcar

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -181,6 +180,7 @@ systemd:
       [Service]
       Type=notify
       EnvironmentFile=-/run/flannel/flannel_docker_opts.env
+      Environment=DOCKER_OPTS=--selinux-enabled=false
 
       # the default is not to use systemd for cgroups because the delegate issues still
       # exists and systemd currently does not support the cgroup feature set required
@@ -586,7 +586,7 @@ func testDockerInfo(expectedFs string, c cluster.TestCluster) {
 	}
 
 	// Validations shared by all versions currently
-	if !reflect.DeepEqual(info.SecurityOptions, []string{"seccomp", "selinux"}) {
+	if !hasSecurityOptions(info.SecurityOptions) {
 		c.Errorf("unexpected security options: %+v", info.SecurityOptions)
 	}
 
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -610,3 +610,21 @@ func testDockerInfo(expectedFs string, c cluster.TestCluster) {
 		c.Errorf("runc was not in runtimes: %+v", info.Runtimes)
 	}
 }
+
+// hasSecurityOptions strictly checks that at least one of
+// the docker security option is enabled (seccomp, selinux)
+func hasSecurityOptions(opts []string) bool {
+	var selinux, seccomp bool
```
invidian, line 617: Wouldn't just `len(opts) > 0` work?

tormath1: Good point ! Previous test was strictly checking for `seccomp` and `selinux` in the security options.

So
```go
len(opts) > 0
```

would work but we won't strictly test the enabled security options. (could be `apparmor`, `toto` or whatever). I tried to stay closer from the current test and to just allow some releases to only have `seccomp` as security options.

kola/tests/docker/docker.go
```diff
@@ -610,3 +610,21 @@ func testDockerInfo(expectedFs string, c cluster.TestCluster) {
 		c.Errorf("runc was not in runtimes: %+v", info.Runtimes)
 	}
 }
+
+// hasSecurityOptions strictly checks that at least one of
+// the docker security option is enabled (seccomp, selinux)
```
invidian, line 615:
```suggestion
// the Docker security option is enabled (seccomp, selinux).
```

kola/tests/docker/docker.go
```diff
@@ -610,3 +610,21 @@ func testDockerInfo(expectedFs string, c cluster.TestCluster) {
...
+		case "selinux":
+			selinux = true
+		case "seccomp":
+			seccomp = true
+		default:
+			return false
+		}
+	}
+
+	return selinux || seccomp
```
invidian, line 629: Maybe you can simplify to:
```suggestion
	for _, opt := range opts {
		switch opt {
		case "selinux", "seccomp":
			return true
		}
	}

	return false
```

tormath1: Good idea but I'm not sure we won't strictly check for `selinux`, `seccomp` or both. Let's suppose `opts = ["apparmor", "selinux"]` we will return `true` even if we have `apparmor` in the options. :thinking:

invidian: so maybe:
```go
switch opt {
case "selinux", "seccomp":
default:
  return false
}

return true
```

tormath1:
@invidian thanks for your review ! Comments have been addressed :)

invidian (approved):
Thanks for addressing the feedback @tormath1. Looks good from Go side of things :+1:

sayanchowdhury (approved):

PR #183, opened 2021-07-02
kola/tests/update: add test for idle status

tormath1:
this test assert that there is no regression regarding the status
check when there is no new upload available

Related to: https://github.com/kinvolk/update_engine/pull/10

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -42,6 +42,12 @@ func init() {
 		},
 		Distros: []string{"cl"},
 	})
+	register.Register(&register.Test{
+		Name:        "cl.update.idle",
+		Run:         idle,
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+	})
 }
 
 func Serve() error {
@@ -57,6 +63,22 @@ func Serve() error {
 	}
 
 	return omahawrapper.Serve()
+}
+
+// assert that -update does not fail
+// when there is nothing to do
+// https://github.com/kinvolk/Flatcar/issues/356
+func idle(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	out, err := c.SSH(m, "update_engine_client -update")
+	if err != nil {
+		c.Fatalf("unable to run update_engine_client -update: %v", err)
+	}
+
+	if string(out) == "Update cancelled -- remote version is not newer than the local one" {
+		c.Fatalf("update should be cancelled")
+	}
 }
 
 func payload(c cluster.TestCluster) {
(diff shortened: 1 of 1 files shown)


kola/tests/update/update.go
```diff
@@ -59,6 +65,22 @@ func Serve() error {
...
+// https://github.com/kinvolk/Flatcar/issues/356
+func idle(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	out, err := c.SSH(m, "update_engine_client -update")
+	if err != nil {
+		c.Fatalf("unable to run update_engine_client -update: %v", err)
+	}
+
+	if string(out) == "Update cancelled -- remote version is not newer than the local one" {
```
jepio, line 79: I'm new to this, but just reading the logic I would think that the test should be:
```suggestion
	if string(out) != "Update cancelled -- remote version is not newer than the local one" {
```

tormath1: @jepio that's totally right, it seems my :brain: is already in week-end. :D 

I also added an output redirection in the fixup commit.

kola/tests/update/update.go
```diff
@@ -59,6 +65,22 @@ func Serve() error {
 	return omahawrapper.Serve()
 }
 
+// assert that -update does not fail
+// when there is nothing to do
+// https://github.com/kinvolk/Flatcar/issues/356
+func idle(c cluster.TestCluster) {
```
pothos, line 71: This test is only valid for developer builds or as long as the next release is not published. It may be confusing when running the tests for a published release.

Besides that I'm also a bit concerned about the runtime of the tests because this small test needs its own instance created and that takes multiple minutes depending on the platform.

pothos: I think there is some support in kola for emulating an update service and to perform an update. In combination with this it makes sense to have a test where maybe in one test case an update is handed and in the other test case there is none. But I wouldn't let the behavior depend on the public update service.

tormath1: Ok then we can close this PR for the moment - I'm currently trying to think about a _better_ way to test the update behavior of Flatcar :)

pothos: I see that we are already in the right file: `cl.update.payload` is currently not used in Jenkins, would make sense to set this up.

```
--- SKIP: cl.update.payload (19.31s)
        update.go:93: no update payload provided
```

tormath1: yeah I dug around a bit - we would need to generate an update payload with `delta_generator` and to provide it to the kola command line. I can open a dedicated issue to track this :) That would be a good start !

PR #184, opened 2021-07-05
cork: fix handling of /dev/shm -> /run/shm symlinks

jepio:
# cork: fix handling of /dev/shm -> /run/shm symlinks

In WSL2 /dev/shm is a symlink to /run/shm. cork has code to handle that but that code didn't work correctly and resulted in "permission denied" errors when entering the SDK chroot.

## How to use

`cork create` in a WSL2 terminal.

## Testing done

`cork create`, checked that no errors occur, and then `cork enter` and checked that /run/shm

1. exists
2. has the sticky bit set

by running `ls -ld /run/shm` in the chroot.

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -142,10 +142,10 @@ func (e *enter) MountAPI() error {
 			return fmt.Errorf("Unexpected shm path: %s", shmPath)
 		}
 		newPath := filepath.Join(e.Chroot, shmPath)
-		if err := os.Mkdir(newPath, 01777); err != nil {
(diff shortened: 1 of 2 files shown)


pothos (approved):
Looks like https://golang.org/pkg/io/fs/#FileMode has its own implementation of the bits uncoupled from the OS and only the 0777 bits are directly compatible.

jepio:
@pothos the submitter is the one that presses "merge pull request"?

pothos:
Yes :)

PR #185, opened 2021-07-06
platform/qemu: specify raw backing_fmt explicitly

jepio:
# Silence warnings from qemu-img

Kola output is littered with warnings about auto-detection of backing image format, this PR deals with those.

## How to use

`./build_image` + `kola run`.

## Testing done

`kola run` before and after, and verified that warnings about backing_fmt gone.

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -194,8 +194,12 @@ func setupDiskFromFile(imageFile string) (*os.File, error) {
 			return nil, err
 		}
 	}
+	imgInfo, err := util.GetImageInfo(backingFile)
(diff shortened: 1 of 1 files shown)


pothos (commented):
Looks good in general for the way Jenkins runs the tests but it would be nice to have a flag to test a qcow2 image, too (`flatcar_production_qemu_image.img` and `flatcar_production_qemu_uefi_image.img`)

jepio:
Make sense. Have a look at my second attempt.

pothos (approved):

PR #186, opened 2021-07-07
platform: provide Board access to the runtime

tormath1:
`board` is accessible only for Packet and QEMU machines. It can be helpful to get this value for other machine in order to do some conditional operation in the setup - like enforcing SELinux. See: https://github.com/kinvolk/mantle/pull/180#issuecomment-875497624.

In this PR, we provide a new `Board () string` method to the `Machine` interface.

`Board()` returns value from a new base options implemented in `platform.Options`. We currently need to use the `syncoptions`  method to pass the value of the `board` to each provider - this one is provided by the `--board` flag bound to the `kola.QEMUOptions.Board`

Based on the `options.Board` we now conditionally enable SELinux on the machine.

### DIFF
--- a/platform/machine/aws/machine.go
+++ b/platform/machine/aws/machine.go
@@ -157,3 +157,7 @@ func (am *machine) JournalOutput() string {
 	}
 	return string(data)
 }
+
+func (am *machine) Board() string {
+	return am.cluster.flight.Options().Board
+}
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -80,6 +80,9 @@ type Machine interface {
 	// JournalOutput returns the machine's journal output if available,
 	// or an empty string.  Only expected to be valid after Destroy().
 	JournalOutput() string
+
+	// Board returns the machine's board
+	Board() string
 }
 
 // Cluster represents a cluster of machines within a single Flight.
@@ -158,6 +161,9 @@ type Options struct {
 	// When specified additional files & units will be automatically generated
 	// inside of RenderUserData
 	OSContainer string
+
+	// Board is the board used by the image
+	Board string
 }
 
 // RuntimeConfig contains cluster-specific configuration.
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -170,7 +170,18 @@ func init() {
 
 // Sync up the command line options if there is dependency
 func syncOptions() error {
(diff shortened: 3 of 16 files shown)


platform/machine/aws/machine.go
```diff
@@ -25,10 +25,12 @@ import (
 	"golang.org/x/crypto/ssh"
 
 	"github.com/coreos/mantle/platform"
+	m "github.com/coreos/mantle/platform/machine"
 	"github.com/coreos/mantle/util"
 )
 
 type machine struct {
+	*m.DefaultMachine
```
pothos, line 33: Idea which I didn't validate: maybe a string could be used here and the `Board()` function added to this file. Same for other platforms.
Maybe this could spare the new DefaultMachine interface?

tormath1: I reverted the `DefaultMachine` interface and implemented a `Board()` for each `platform/machine/*/machine`. This one will return the `baseopts` from the current flight where the `Board` value has been passed from the `syncoptions`.

platform/platform.go
```diff
@@ -158,6 +161,9 @@ type Options struct {
 	// When specified additional files & units will be automatically generated
 	// inside of RenderUserData
 	OSContainer string
+
+	// Board is the board used by the image
+	Board string
```
pothos, line 166: I don't follow; when this variable is added here, how should it relate to the existing `Board` variable used in `../platform/api/packet/api.go`?
Above it doesn't get set because only `kola.PacketOptions.Board = board` is assigned. I would recommend to unify this: either use only the new `Board` defined here or add `Board` variables to all platform options instead of here.

tormath1: yeah, it's a bit confusing - basically we are in this use-case for providers != qemu & packet:
```go
type A struct {
	*B
}

type B struct {
	board string
}

a := A{&B{}}
fmt.Printf("%p\n", &a.board) // 0xc00008e000
fmt.Printf("%p\n", &a.B.board) // 0xc00008e000
```

So for `qemu` and `packet` we can simply remove the custom `Board` for each in order to rely on the unique `Board` defined in `platform.Options`.

pothos: Thanks, looks good now - previously I think the wrong variable was used for Packet/QEMU in the SELinux test, now it's more clear what's going on.

pothos (approved):

PR #191, opened 2021-07-15
platform/pkt/machine: delete SSH key after provisionning

tormath1:
we defer the key deletion, no matter the result of the provisionning

## How to use

- run tests on `packet` platform

closes https://github.com/kinvolk/Flatcar/issues/236


pothos:
There already is the `Destroy()` function of the flight which removes the key. Also, doing it after provisioning the machine is not wanted I guess?
The problem of left-over keys seems to be linked to terminating kola, and the solution is to add key removal to the `ore packet gc` command (see `GC` in `platform/api/packet/api.go`).

PR #192, opened 2021-07-16
kola/test: disable tests that won't work with docker 20.10

jepio:
# Disable old tests for alpha channel

docker 1.9 and kubernetes <1.19 are not compatible with docker 20 that we are rolling into alpha. These tests need to be excluded, to be removed later.

## How to use

Fetch the image from http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3035/ and run test (`kola run`)

## Testing done

`kola run`

See also kinvolk/coreos-overlay#931

### DIFF
--- a/kola/tests/kubernetes/basic.go
+++ b/kola/tests/kubernetes/basic.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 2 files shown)


kola/tests/kubernetes/basic.go
```diff
@@ -54,6 +54,8 @@ func init() {
 				ClusterSize: 0,
 				Platforms:   []string{"gce", "do", "aws", "qemu", "azure"}, // TODO: fix packet, esx
 				Distros:     []string{"cl"},
+				// incompatible with docker >=20.10
+				ExcludeChannels: []string{"alpha"},
```
pothos, line 58: With the recent changes we now can do
```suggestion
				MinVersion:  semver.Version{Major: 2956},
```
where 2956 is larger than the currently released Alpha 2955.

(and same for the other test)

pothos (changes requested):

pothos (approved):

kola/tests/kubernetes/basic.go
pothos: Yes, EndVersion ;)

jepio: Yes, that worked better than MinVersion 😄

PR #193, opened 2021-07-22
kola/test/update: reconfigure the instance once rebooted

tormath1:
the instance is currently configured once to be able to use
the `update_payload`.
Once the payload downloaded and the instance rebooted, kola
is going to download again an update: in the case we previously
downloaded and installed an official release, the
/usr/share/update-engine/pub-key will be changed to the official one.
So update-engine will fail to verify the new update payload since this
one appears to be signed only with a dev-key (it's generated from
the SDK.)
So we reconfigure again to inject the dev-pub-key to verify correctly
the downloaded payload

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -83,6 +83,16 @@ func payload(c cluster.TestCluster) {
 
 	tutil.InvalidateUsrPartition(c, m, "USR-A")
 
+	/*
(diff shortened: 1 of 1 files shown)


jepio (approved):
LGTM (maybe add a code comment explaining what you wrote in the commit message?)

tormath1:
@jepio thanks for reviewing. Comment has been added, I'll squash and merge the commit !

PR #198, opened 2021-08-02
kola: support nightly version in version comparisons

pothos:
- kola: support nightly version in version comparions
    
    The nightly version is the build date but the build ID has the branch
    reference. To support the version comparion for nightly builds we need
    to generate a version for the "main" branch and the major version
    branches that is higher than the current release version.
    
    Use 999999.99.99 as version in the semantic version checks for nightly
    builds from the "main" branch or custom dev builds and use MAJOR.99.99
    as version for nightly builds from the release branch.
- kola/tests/misc/verity: run test based on major version

## Testing done

```
$ sudo ./kola run  -k --keys --key ~/.ssh/id_rsa.pub --board=amd64-usr --channel=alpha --parallel=1 --platform=qemu    --qemu-image=flatcar_production_image.bin
2021-08-02T15:25:15Z kola: Using "999999.99.99" as version to filter tests...
=== RUN   cl.network.initramfs.second-boot
[…]
```

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -432,9 +432,25 @@ func getClusterSemver(flight platform.Flight, outputDir string) (*semver.Version
 
 	out, stderr, err := m.SSH("grep ^VERSION_ID= /etc/os-release")
 	if err != nil {
-		return nil, fmt.Errorf("parsing /etc/os-release: %v: %s", err, stderr)
+		return nil, fmt.Errorf("parsing /etc/os-release for VERSION_ID: %v: %s", err, stderr)
 	}
 	ver := strings.Split(string(out), "=")[1]
+
+	out, stderr, err = m.SSH("grep ^BUILD_ID= /etc/os-release | { grep -o 'dev-.*' || true ; } | cut -d - -f 2-  | rev | cut -d - -f 2- | rev")
+	if err != nil {
+		return nil, fmt.Errorf("parsing /etc/os-release for nightly branch in BUILD_ID: %v: %s", err, stderr)
+	}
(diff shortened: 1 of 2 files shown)


kola/harness.go
```diff
@@ -432,10 +432,25 @@ func getClusterSemver(flight platform.Flight, outputDir string) (*semver.Version
...
 
+	out, stderr, err = m.SSH("grep ^BUILD_ID= /etc/os-release | { grep -o 'dev-.*-nightly' || true ; } | cut -d - -f 2-  | rev | cut -d - -f 2- | rev")
+	if err != nil {
+		return nil, fmt.Errorf("parsing /etc/os-release for nightly branch in BUILD_ID: %v: %s", err, stderr)
+	}
+	branch := string(out)
+	if branch == "main" || branch == "flatcar-master" {
+		// "main" is a nightly build of the main branch,
+		// "flatcar-master" refers to the manifest branch where dev builds are started
+		ver = "9999.99.99"
```
krnowak, line 447: I'd make it `999999.99.99`. We don't need to saddle ourselves with "year 2040" problem, right after "year 2038" one. ;)

pothos: Yes, let's trade it for a "year 2286" problem

pothos: Wait, you had one 9 more, "year 4752" - even better!

pothos: My reasoning was to stay < 32767 in case someone swaps out the semver lib to one that is more limited

tormath1 (approved):

PR #199, opened 2021-08-06
kola: support lts version in version comparisons

jepio:
# kola: support lts version in version comparisons

This fixes a small issue in how the lts version number is parsed after the introduction of nightly version number parsing.


jepio:
Superseded by #200

PR #200, opened 2021-08-06
kola/harness: detect LTS major version

pothos:
The LTS branch is flatcar-lts-MAJOR and only flatcar-MAJOR branch names
    were considered which let to an invalid major version "lts" being
    parsed.
    Take the element after either "flatcar" or "flatcar-lts" when splitting
    on "-". The full build ID is "dev-flatcar(-lts)-MAJOR-nightly-XYZ".

## Testing done

Tested with a nightly build for `flatcar-2955`


pothos:
@jepio I saw your PR too late, but took your changes now while simplifying the shell command.

jepio (approved):
Looks good

PR #196, opened 2021-07-30
kola/kubeadm: add kubernetes 1.22 test

tormath1:
this commit brings a new release of kubernetes to test but it also fixes
a `TODO:`. We are now able to provide multiple kubernetes release to
test.
We just need to create a new `map[string]interface{}` holding the
params of the kubernetes release we want to test.

```
$ ./bin/kola list
kubeadm.v1.21.0.calico.base
kubeadm.v1.21.0.cilium.base
kubeadm.v1.21.0.flannel.base
kubeadm.v1.22.0.calico.base
kubeadm.v1.22.0.cilium.base
kubeadm.v1.22.0.flannel.base
```

Signed-off-by: Mathieu Tortuyaux <mathieu@kinvolk.io>

**note for reviewers**:
* I'm a bit concerned by the number of tests we provide now for kubernetes: versions + CNIs are creating big test matrix which can slow down the tests overall... but I guess it's the price to pay in order to identify potential failures with kubernetes workloads

PR #205, opened 2021-08-11
kola/kubeadm: fix CNI selection

tormath1:
register action was keeping the latest occurence of `CNI` so it was
basically always testing the `cilium` CNI.

... it happens.

The issue was already being fixed into #196 but it's better to be merged now to confirm that https://github.com/kinvolk/coreos-overlay/pull/1181 is solving the actual `flannel` issue raised in https://github.com/kinvolk/mantle/pull/196#issuecomment-895358491

### DIFF
--- a/kola/tests/kubeadm/kubeadm.go
+++ b/kola/tests/kubeadm/kubeadm.go
@@ -76,12 +76,17 @@ systemd:
 
 func init() {
 	for _, CNI := range CNIs {
+		// we need to copy the value of `CNI` to pass it
+		// to the `Run` method, otherwise only the last
+		// value of `CNI` (`cilium`) will be used in the registered
+		// test
+		cni := CNI
 		register.Register(&register.Test{
-			Name:             fmt.Sprintf("kubeadm.%s.base", CNI),
+			Name:             fmt.Sprintf("kubeadm.%s.base", cni),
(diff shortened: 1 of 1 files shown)


kola/tests/kubeadm/kubeadm.go
```diff
@@ -76,12 +76,13 @@ systemd:
 
 func init() {
 	for _, CNI := range CNIs {
+		cni := CNI
```
krnowak, line 79: I see now that this is variable capturing in the `Run` function that bit us here. Nasty. Would a comment about the fact be helpful here?

tormath1: Yes for sure. Actually, this section will be replaced by a new one from this PR (#196) - and a comment is already there to explain while we need such a thing (see https://github.com/kinvolk/mantle/pull/196/files#diff-833c725698d5fe10161a1c491a560567f1473d7541658c722022bc45db2d4c3fR88-R95)

krnowak: Ah, cool. Thanks.

tormath1:
@krnowak yes.. Github does not handle yet the `--autosquash` when it detects `--fixup` commits. :/

sayanchowdhury (approved):

PR #208, opened 2021-08-17
sdk: pass '--quiet' to repo sync

jepio:
The messages are long and scroll by too fast to be of any use to users. Adding
quiet turns it into a nice progress indicator.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 
 	"github.com/coreos/go-semver/semver"
+	"github.com/coreos/pkg/capnslog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
@@ -282,7 +283,8 @@ func updateRepo() {
 		}
 	}
 
-	if err := sdk.RepoSync(chrootName, forceSync, useHostDNS); err != nil {
+	verbose := plog.LevelAt(capnslog.INFO)
+	if err := sdk.RepoSync(chrootName, forceSync, verbose, useHostDNS); err != nil {
 		plog.Fatalf("repo sync failed: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


sayanchowdhury (approved):
I was thinking of adding this last night! thanks!

tormath1:
Would it be relevant to bind this `--quiet` option to the `--debug` or `--verbose` global flags ?

jepio:
> Would it be relevant to bind this `--quiet` option to the `--debug` or `--verbose` global flags ?

I could do that, but I would need some help. Do you have a suggestion of how I might get access to the loglevel?

tormath1:
From a first look - it seems that each command (`cork`, `kola`, etc.) inherits from the `main` flags:https://github.com/kinvolk/mantle/blob/927a11b7463d8befc3face602e423e52fbd66953/cli/cli.go#L55-L60

We can confirm by running simple commands:
```shell
$ cork create --help
...

Global Flags:
  -d, --debug             Alias for --log-level=DEBUG
      --log-level value   Set global log level. (default NOTICE)
  -v, --verbose           Alias for --log-level=INFO
```

Since we use the `cobra` framework, it should be easy to get back to the parent command to get the `--debug` / `--verbose` flag from the `runCreate`:
https://github.com/kinvolk/mantle/blob/927a11b7463d8befc3face602e423e52fbd66953/cmd/cork/create.go#L224

Using `cmd.Flags().GetBool("verbose")` for example in order to access the value of `--verbose` - not tested but maybe we can access directly the flag with [command.Flag](https://pkg.go.dev/github.com/spf13/cobra#Command.Flag).

Once we have the value, we can propagate it to the `RepoSync` method in a similar way as the `forceSync` variable:
https://github.com/kinvolk/mantle/blob/927a11b7463d8befc3face602e423e52fbd66953/cmd/cork/create.go#L285-L287

jepio:
@tormath1 I think i found the right way to do this. It also works as intended.

tormath1 (approved):
LGTM, thanks !

PR #214, opened 2021-08-24
update Azure SDK for Gen2 VM support

jepio:
# update Azure SDK for Gen2 VM support

This updates Azure SDK to an up-to-date version. This lets us create Azure Gen2 VMs, and provisioning time seems to be improved. I had an issue with azure-vhd-utils which seems to be abandoned but there's no real replacement for it. It was necessary to update it as well and so I made a personal fork.

See https://github.com/kinvolk/Flatcar/issues/417

## How to use

Build and run.

## Testing done

* `kola spawn` works for both V1 and V2 (if https://github.com/kinvolk/coreos-overlay/pull/1198 is pulled)
* `kola run` works, for as long as I let it run
* `ore azure gc` works


go.mod
```diff
@@ -82,6 +84,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
+replace github.com/Microsoft/azure-vhd-utils => github.com/jepio/azure-vhd-utils v0.0.0-20210818134022-97083698b75f
```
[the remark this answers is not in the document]
jepio: Yes absolutely, I think it should be there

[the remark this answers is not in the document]
jepio: Its there now

platform/api/azure/storage_mit.go
```diff
@@ -91,16 +90,19 @@ func (a *API) ListBlobs(storageaccount, storagekey, container string, params sto
 
 	bsc := sc.GetBlobService()
 
-	resp, err := bsc.ListBlobs(container, params)
+	resp, err := bsc.GetContainerReference(container).ListBlobs(params)
 	if err != nil {
 		return nil, fmt.Errorf("failed listing blobs for %q: %v", container, err)
 	}
 	var res []storage.Blob
+	cont := bsc.GetContainerReference(container)
```
krnowak, line 98: This could be moved above the line 93, where you could then call `resp, err := cont.ListBlobs(params)`.

platform/api/azure/storage_mit.go
```diff
@@ -290,8 +299,11 @@ func (a *API) SignBlob(storageaccount, storagekey, container, blob string) (stri
 	}
 
 	bsc := sc.GetBlobService()
-
-	return bsc.GetBlobSASURI(container, blob, time.Date(2099, time.December, 31, 23, 59, 59, 0, time.UTC), "rl") // TODO: Migrate to new API version
+	b := bsc.GetContainerReference(container).GetBlobReference(blob)
+	blobSASOptions := storage.BlobSASOptions{}
+	blobSASOptions.Read = true
+	blobSASOptions.Expiry = time.Date(2099, time.December, 31, 23, 59, 59, 0, time.UTC)
+	return b.GetSASURI(blobSASOptions) // TODO: Migrate to new API version
```
krnowak, line 306: Is this TODO still valid?

krnowak (approved):
Just one nit and a question. Nothing blocking.

Also, there is a hefty amount of repetition around `future.WaitForCompletionRef(…)` and following-up `_, err := future.Result(…)` but some quick glance at the azure API tells me that this boilerplate rather can't be put into a single function, because each future object is a separate type with a different `Result` member. Some templates in go would be nice to have right now. ;)

platform/api/azure/storage_mit.go
jepio: Funny that you mention it. I was just looking into the Azure marketplace with Sayan and I realized this is wrong. The SAS token needs to be at container level (with read+list permissions) but the URI needs to refer to the blob. So this will get even uglier.

jepio: Pushed a new version of this operation in the latest commit. I don't see a nicer way to do what the marketplace requires :(

jepio:
> Just one nit and a question. Nothing blocking.
> 
> Also, there is a hefty amount of repetition around `future.WaitForCompletionRef(…)` and following-up `_, err := future.Result(…)` but some quick glance at the azure API tells me that this boilerplate rather can't be put into a single function, because each future object is a separate type with a different `Result` member. Some templates in go would be nice to have right now. ;)

Yes, every future is different ;) They replaced this construct in different part of the SDK with something poll based (https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md#long-running-operations) but it looks just as operation specfiic.

krnowak (approved):

PR #222, opened 2021-09-01
plaform/util: enable SELinux for all arch

tormath1:

Since `arm64` is going to support `SELinux`, there is no more need to not enable SELinux for `arm64` in the tests runtime.

Note: all the test will enable SELinux for `arm64`, we need to find a way to prevent channel != alpha to enable SELinux, otherwise all tests will fail until `SELinux/arm64` lands into `stable`.

## How to use

related to: https://github.com/kinvolk/coreos-overlay/pull/1245
## Testing done

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Security
 ### Added
 ### Changed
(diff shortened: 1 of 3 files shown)


pothos:
Thanks, I would say that tests for arm64 on Beta/Stable can fail as they do now already.

pothos (approved):
Would be great to merge this and have it in Jenkins, otherwise I need to run the tests locally to see which ones are still failing.

PR #224, opened 2021-09-03
Jenkinsfile: also build arm64 artifacts

jepio:
With this change we no longer need the job parameter so it has been removed.
The arm64 build requires disabling 'CGO_ENABLED' as otherwise we would need a
cross-compiler in the build container. The arm64 artifacts are now archived at
bin.arm64. The build has also been switched to go1.16, this is what we are
using in the Flatcar SDK as well.


sayanchowdhury (approved):
LGTM

PR #225, opened 2021-09-03
kola/test/docker: enable `docker.selinux` for all arch

tormath1:
SELinux is now supported on ARM64

## Testing done

```
# get image from SELinux PR https://github.com/kinvolk/coreos-overlay/pull/1245
wget https://storage.googleapis.com/flatcar-jenkins/developer/developer/boards/arm64-usr/2021.09.01%2Bdev-flatcar-master-3470/flatcar_production_image.bin.bz2
sudo ./bin/kola run --channel alpha --board arm64-usr --key ${HOME}/.ssh/id_rsa.pub -k -b cl --qemu-bios ./flatcar_production_qemu_uefi_efi_code.fd  -p qemu --qemu-image ./flatcar_production_image.bin docker.selinux
```

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -55,12 +55,11 @@ type simplifiedDockerInfo struct {
 
 func init() {
 	register.Register(&register.Test{
-		Run:           dockerSELinux,
-		ClusterSize:   1,
-		Name:          "docker.selinux",
-		Distros:       []string{"cl"},
-		Channels:      []string{"alpha", "beta"},
-		Architectures: []string{"amd64"},
+		Run:         dockerSELinux,
+		ClusterSize: 1,
+		Name:        "docker.selinux",
+		Distros:     []string{"cl"},
+		MinVersion:  semver.Version{Major: 2942},
 	})
 	register.Register(&register.Test{
 		Run:         dockerNetwork,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Changed
 - Enabled SELinux for ARM64 ([#222](https://github.com/kinvolk/mantle/pull/222/))
+- Enabled `docker.selinux` test for ARM64 ([#225](https://github.com/kinvolk/mantle/pull/225))
 
 ### Removed
 


kola/tests/docker/docker.go
```diff
@@ -55,12 +55,11 @@ type simplifiedDockerInfo struct {
...
-		ClusterSize:   1,
-		Name:          "docker.selinux",
-		Distros:       []string{"cl"},
-		Channels:      []string{"alpha", "beta"},
-		Architectures: []string{"amd64"},
+		Run:         dockerSELinux,
+		ClusterSize: 1,
+		Name:        "docker.selinux",
+		Distros:     []string{"cl"},
+		Channels:    []string{"alpha", "beta"},
```
pothos, line 62: While at it we should switch this to using `MinVersion` or we will forget this when the next channel transition happens

pothos: `MinVersion:       semver.Version{Major: 2942},` should mean everything newer, including the current beta?

tormath1: why not `MinVersion: semver.Version{Major: 2984},` ? SELinux changes will be available only in the next `alpha` :thinking:

pothos: Ah yes, I just thought about the existing `beta` value

tormath1: This is an edge case actually:
  * SELinux `amd64` is available on `alpha`/`beta`
  * SELinux `arm64` is only on `main` atm

so:
* `MinVersion: semver.Version{Major: 2942},`: will make the test fail on `arm64` for both beta / alpha
* `MinVersion: semver.Version{Major: 2984},`: will make the test disappears for `amd64` for both beta / alpha

pothos: Yes, `MinVersion: semver.Version{Major: 2984},` is the right one :D

pothos: Better to skip the tests than have them fail when doing a bugfix release

pothos: ah, wait, didn't get it fully when reading the first time… we have no way to specify the version per architecture, complicated. I think it's ok to let it fail for arm64 on Alpha bugfix releases then and use `2942`.

tormath1: ok done with `2942` :)

pothos (approved):

PR #227, opened 2021-09-06
kola/test: use clc to set `enable_v2`

tormath1:
In 12b5a96ee4099cc3c2dbe4af35ebbfc55b13868a, we enabled the V2 support
for `etcd`. It means that `etcd` starts with support for both datastore
version.

Support has been enabled through a `systemd` dropin - this one will
generate a dropin under the `etcd-member.service` unit but
`container-linux-config` will generate a dropin under
`etcd-member.service` name too.

This is not an issue with `ignition/v2` but it is with `ignition/v3`
(see: https://github.com/kinvolk/ignition/blob/main/docs/configuration-v3_0.md).

So we need to add the `enable_v2` direcly into the `clc` - `enable_v2`
is present in `clc` from `config.Types.Etcd3_2` but by default it
targets `3.0.0` so we need to pin the version to the current tested one
(`3.5.0`) to use `enable_v2` and the correct image tag.

## Testing done
```
sudo ./bin/kola run --key ${HOME}/.ssh/id_rsa.pub -k -b cl -p qemu --qemu-image /home/mathieu/kinvolk/alpha/flatcar_production_qemu_image.img -j 1 cl.locksmith.cluster cl.etcd-member.v2-backup-restore cl.flannel.udp cl.flannel.vxlan
```

## Note for reviewers

CLC needs to be [upgraded](https://github.com/kinvolk/Flatcar/issues/398) to support `etcd>3.3.0` - it's not an issue at the moment for the tests except this warnings:
```
2021-09-06T08:50:24Z platform/conf: parsing Container Linux config: warning at line 4, column 32
Etcd minor version specified is too new, only options available in the previous minor version will be accepted
```

but we won't be able to use particular options from `etcd>3.3.0` in the future tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Enabled SELinux for ARM64 ([#222](https://github.com/kinvolk/mantle/pull/222/))
 - Enabled `docker.selinux` test for ARM64 ([#225](https://github.com/kinvolk/mantle/pull/225))
 - Fixed `amd64` checksums for Kubernetes `v1.21.0` tests ([#226](https://github.com/kinvolk/mantle/pull/226))
+- Used `clc` to set `enable_v2` option ([#227](https://github.com/kinvolk/mantle/pull/227))
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


pothos (approved):
Good to know that a drop-in is not even needed

tormath1:
@pothos actually `clc` hardcodes the config for each `etcd` version (https://github.com/kinvolk/container-linux-config-transpiler/blob/flatcar-master/config/types/etcd.go) - I guess it could be reworked a bit in order to rely exclusively on https://github.com/etcd-io/etcd/blob/main/server/config/config.go

PR #233, opened 2021-09-15
kola/test: add `bpf` tests

tormath1:
In this PR, we kick-off the writing of `bpf` tests.

To use it: 
```
sudo ./bin/kola run -b cl -p qemu --qemu-image ./flatcar_production_qemu_image.img bpf.execsnoop
```

Related to: https://github.com/flatcar-linux/Flatcar/issues/89

### DIFF
--- a/kola/tests/bpf/bpf.go
+++ b/kola/tests/bpf/bpf.go
@@ -1,0 +1,126 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package bpf
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/coreos/pkg/capnslog"
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+	"github.com/flatcar-linux/mantle/util"
+)
+
+// cmdPrefix is a temporary hack to pull `bcc` tools into Flatcar
+const cmdPrefix = "docker run -d --name %s -v /lib/modules:/lib/modules -v /sys/kernel/debug:/sys/kernel/debug -v /sys/fs/cgroup:/sys/fs/cgroup -v /sys/fs/bpf:/sys/fs/bpf --privileged --net host --pid host quay.io/iovisor/bcc %s"
+
+var (
+	plog = capnslog.NewPackageLogger("github.com/flatcar-linux/mantle", "kola/tests/bpf")
+)
+
+// Log defines the standard log format
+// from Docker
+// https://docs.docker.com/config/containers/logging/json-file/
+type Log struct {
+	Log    string `json:"log"`
+	Time   string `json:"time"`
+	Stream string `json:"stream"`
+}
+
+func init() {
+	register.Register(&register.Test{
+		Run:         execsnoopTest,
+		ClusterSize: 1,
+		Name:        `bpf.execsnoop`,
+		Distros:     []string{"cl"},
+		// required while SELinux policy is not correcly updated to support
+		// `bpf` and `perfmon` permission.
+		Flags: []register.Flag{register.NoEnableSelinux},
+	})
+}
+
+func execsnoopTest(c cluster.TestCluster) {
+	c.Run("filter name and args", func(c cluster.TestCluster) {
+		m := c.Machines()[0]
+		containerName := "execsnoop"
+
+		// filter commands with `docker ps`
+		plog.Infof("running %s container", containerName)
+		cmd := fmt.Sprintf(cmdPrefix, containerName, "/usr/share/bcc/tools/execsnoop -n docker -l ps")
+		if _, err := c.SSH(m, cmd); err != nil {
+			c.Fatalf("unable to run SSH command '%s': %v", cmd, err)
+		}
+
+		// wait for the container and the `execsnoop` command to be correctly started before
+		// generating traffic.
+		if err := util.Retry(5, 2*time.Second, func() error {
+			_ = c.MustSSH(m, "docker ps")
+
+			// we first assert that the container is running and then the process too.
+			// it's not possible to use `docker top...` command because it's the execsnoop itself who takes some time to start.
+			logs, err := c.SSH(m, fmt.Sprintf("sudo cat $(docker inspect --format='{{.LogPath}}' %s)", containerName))
+			if err != nil {
+				return fmt.Errorf("getting running process: %w", err)
+			}
+
+			if len(logs) > 0 {
+				return nil
+			}
+
+			return fmt.Errorf("no logs, the service has not started yet properly")
+		}); err != nil {
+			c.Fatalf("unable to get container ready: %v", err)
+		}
+
+		// generate some "traffic"
+		_ = c.MustSSH(m, "docker info")
+		_ = c.MustSSH(m, fmt.Sprintf("docker logs %s", containerName))
+		_ = c.MustSSH(m, fmt.Sprintf("docker top %s", containerName))
+
+		plog.Infof("getting logs from %s container", containerName)
+		logs, err := c.SSH(m, fmt.Sprintf("sudo cat $(docker inspect --format='{{.LogPath}}' %s)", containerName))
+		if err != nil {
+			c.Fatalf("unable to run SSH command: %v", err)
+		}
+
+		l := Log{}
+		dockerLogs := bytes.Split(logs, []byte("\n"))
+		if len(dockerLogs) != 3 {
+			// we have the headers of the table
+			// then 2 lines for docker ps and the torcx call
+			c.Fatalf("docker logs should hold 3 values. Got: %d", len(dockerLogs))
+		}
+
+		execFound := false
+		for _, log := range dockerLogs {
+			if len(log) == 0 {
+				continue
+			}
+
+			if err := json.Unmarshal(log, &l); err != nil {
+				c.Fatalf("unable to unmarshal log: %v", err)
+			}
+			plog.Infof("handling log %v", l)
+
+			if l.Stream == "stderr" {
+				c.Fatal("stream should not log to 'stderr'")
+			}
+
+			if strings.Contains(l.Log, "docker info") || strings.Contains(l.Log, "docker top") || strings.Contains(l.Log, "docker logs") {
+				c.Fatal("log should not contain 'docker info' or 'docker top' or 'docker logs'")
+			}
+
+			if strings.Contains(l.Log, "docker ps") {
+				execFound = true
+			}
+		}
+
+		if !execFound {
+			c.Fatal("did not get 'docker ps' in the logs")
+		}
+	})
+}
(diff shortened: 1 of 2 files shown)


kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/pkg/capnslog"
+)
+
+var (
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "kola/tests/bpf")
+	// cmdPrefix is a temporary hack to pull `bcc` tools into Flatcar
+	cmdPrefix = "docker run -d --name %s -v /lib/modules:/lib/modules -v /sys/kernel/debug:/sys/kernel/debug -v /sys/fs/cgroup:/sys/fs/cgroup -v /sys/fs/bpf:/sys/fs/bpf --privileged --net host --pid host quay.io/iovisor/bcc %s"
```
jepio, line 32: what if this would instead rely on the bcc command aliases added recently?

tormath1: yes - this PR was done before the aliases. Let's try to use them.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,117 @@
...
+
+		// sleep 2s to be sure that the container is correctly started
+		time.Sleep(2 * time.Second)
+
+		// generate some "traffic"
+		_ = c.MustSSH(m, "docker ps")
+		_ = c.MustSSH(m, "docker info")
+
+		plog.Infof("getting logs from %s container", containerName)
+		logs, err := c.SSH(m, fmt.Sprintf("sudo cat $(docker inspect --format='{{.LogPath}}' %s)", containerName))
```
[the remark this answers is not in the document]
tormath1: Using `docker logs` command you don't benefits from the JSON output in order to marshal the content into the `Log` struct right after.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			c.Fatalf("unable to run SSH command: %v", err)
+		}
+
+		l := Log{}
+		dockerLogs := bytes.Split(logs, []byte("\n"))
+		if len(dockerLogs) != 3 {
```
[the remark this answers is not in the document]
tormath1: it's supposed not - since you listen only for `docker ps` commands, we should only get those: https://github.com/flatcar-linux/mantle/pull/233/files#diff-90680d4b953422a18f3c1571bda33d139308ffff8726a956b21e23b489db4aa6R63
In case we have other entries, it means the test is not successful since the command has caught other things than `docker ps` events.

kola/tests/bpf/bpf.go
tormath1: `execsnoop` is not part of the added aliases :/

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,117 @@
...
+
+		// filter commands with `docker ps`
+		plog.Infof("running %s container", containerName)
+		cmd := fmt.Sprintf(cmdPrefix, containerName, "/usr/share/bcc/tools/execsnoop -n docker -l ps")
+		if _, err := c.SSH(m, cmd); err != nil {
+			c.Fatalf("unable to run SSH command '%s': %v", cmd, err)
+		}
+
+		// sleep 2s to be sure that the container is correctly started
+		time.Sleep(2 * time.Second)
```
krnowak, line 69: Meh, I suppose that there isn't any better way here, right?

krnowak (commented):
One question, otherwise looks alright.

kola/tests/bpf/bpf.go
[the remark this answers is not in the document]
jepio: I was just thinking this might not work if you would enable journald log driver, but since we don’t and I see your point I’m resolving.

kola/tests/bpf/bpf.go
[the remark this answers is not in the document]
jepio: Ok, missed that, thanks

kola/tests/bpf/bpf.go
jepio: Oh that’s too bad (and independent of this PR: worth adding)

kola/tests/bpf/bpf.go
tormath1: Not a big fan neither to use `time.Sleep` - I already see the flaky tests in the CI.

Now you say it, I think we can try to use the `util.Retry`: https://github.com/flatcar-linux/mantle/blob/2ad720efa5796c258e1ddf4d5d55ddad7ddc5715/util/retry.go#L22-L25

To assert that the container is correctly started.

krnowak: That certainly would be an improvement, thanks. :)

tormath1: It's done - in the `Retry` method, we run a `docker ps` command if this one does not generate any logs, it means that `execsnoop` is not yet fully started.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,137 @@
...
+		// filter commands with `docker ps`
+		plog.Infof("running %s container", containerName)
+		cmd := fmt.Sprintf(cmdPrefix, containerName, "/usr/share/bcc/tools/execsnoop -n docker -l ps")
+		if _, err := c.SSH(m, cmd); err != nil {
+			c.Fatalf("unable to run SSH command '%s': %v", cmd, err)
+		}
+
+		// wait for the container and the `execsnoop` command to be correctly started before
+		// generating traffic.
+		if err := util.Retry(5, 2*time.Second, func() error {
```
jepio, line 71: Every loop of the retry runs `docker ps` so now its definitely possible to have more than 3 lines in the log (if the retry happens)?

tormath1: Mm, I get your point but it should not:

in the iteration, we run `docker ps` then we check `sudo cat $(docker inspect --format='{{.LogPath}}' %s)`:

- empty: it means `execsnoop` is not correctly started (or does not filter correctly) -> we continue to iterate
- not empty: `execsnoop` is correctly started because it generated some logs based on the `docker ps` "trigger" -> we continue the program

Even in the case we loop 5 times, it means that in the meantime not logs have been generated -> `execsnoop` was not ready. In the end, we should always get 3 lines into the logs.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,137 @@
+// Copyright © Microsoft Corporation
```
dongsupark, line 1: I would prefer to have a neutral term like `Copyright 2021 The mantle Authors` in https://github.com/flatcar-linux/mantle/blob/0282d298e43b0a99e17faaf9a0396d10a6fab4e0/kola/tests/util/template.go#L1 .

dongsupark: Thanks for updating the headers.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/coreos/pkg/capnslog"
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+	"github.com/flatcar-linux/mantle/util"
+)
+
+var (
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "kola/tests/bpf")
```
krnowak, line 19:
```suggestion
	plog = capnslog.NewPackageLogger("github.com/flatcar-linux/mantle", "kola/tests/bpf")
```

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,125 @@
...
+	// cmdPrefix is a temporary hack to pull `bcc` tools into Flatcar
+	cmdPrefix = "docker run -d --name %s -v /lib/modules:/lib/modules -v /sys/kernel/debug:/sys/kernel/debug -v /sys/fs/cgroup:/sys/fs/cgroup -v /sys/fs/bpf:/sys/fs/bpf --privileged --net host --pid host quay.io/iovisor/bcc %s"
+)
+
+// Log defines the standarg log format
```
krnowak, line 24:
```suggestion
// Log defines the standard log format
```

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/coreos/pkg/capnslog"
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+	"github.com/flatcar-linux/mantle/util"
+)
+
+var (
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "kola/tests/bpf")
+	// cmdPrefix is a temporary hack to pull `bcc` tools into Flatcar
+	cmdPrefix = "docker run -d --name %s -v /lib/modules:/lib/modules -v /sys/kernel/debug:/sys/kernel/debug -v /sys/fs/cgroup:/sys/fs/cgroup -v /sys/fs/bpf:/sys/fs/bpf --privileged --net host --pid host quay.io/iovisor/bcc %s"
```
krnowak, line 21: Could you maybe make it `const` instead of `var`?

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,125 @@
...
+				c.Fatalf("unable to unmarshal log: %v", err)
+			}
+			plog.Infof("handling log %v", l)
+
+			if l.Stream == "stderr" {
+				c.Fatal("stream should not log to 'stderr'")
+			}
+
+			if strings.Contains(l.Log, "docker info") || strings.Contains(l.Log, "docker top") || strings.Contains(l.Log, "docker logs") {
+				c.Fatal("log should not container docker info")
```
krnowak, line 113:
```suggestion
				c.Fatal("log should not contain 'docker info' or 'docker top' or 'docker logs'")
```

krnowak (changes requested):

kola/tests/bpf/bpf.go
tormath1: Thanks, missed that during the rebase.

tormath1:
@krnowak thanks for the review - comments have been addressed :)

krnowak (approved):
Cool, looks good.

kola/tests/bpf/bpf.go
```diff
@@ -0,0 +1,126 @@
...
+// from Docker
+// https://docs.docker.com/config/containers/logging/json-file/
+type Log struct {
+	Log    string `json:"log"`
+	Time   string `json:"time"`
+	Stream string `json:"stream"`
+}
+
+func init() {
+	register.Register(&register.Test{
```
jepio, line 35: On arm this fails with:
```
cluster.go:117: WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```
Could you disable this test for arm for the time being?

tormath1: Mm, Docker image has not yet `arm64` support - let's open a tracking issue upstream and exclude ARM64 for this test then.
Thanks for the report !

PR #234, opened 2021-09-17
kola/misc/raid: change CLC to raw ignition config

tormath1:
`size` has been deprecated to `sizeMiB` from `ignition-2.3` and this
parameter is not translatable. See the discussion in here: https://github.com/coreos/ign-converter/pull/29

Since CLC is converted to `ignition-2.2` we can't rely on this to pass
the config so we directly use raw ignition 2.3.0 to provision the
instance.

This test only runs on `qemu` so there is no side effect to pass from
`clc` to `ignition`.

## Testing done

Tested against:
* alpha-2983.0.0
* image with ignition/v3 from: https://github.com/flatcar-linux/coreos-overlay/pull/1249

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -26,37 +26,68 @@ import (
 )
 
 var (
-	raidRootUserData = conf.ContainerLinuxConfig(`storage:
-  disks:
-    - device: "/dev/disk/by-id/virtio-secondary"
-      wipe_table: true
-      partitions:
-       - label: root1
-         number: 1
-         size: 256MiB
-         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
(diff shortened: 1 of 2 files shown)


kola/tests/misc/raid.go
```diff
@@ -26,37 +26,68 @@ import (
...
+  "networkd": {},
+  "storage": {
+    "disks": [
+      {
+        "device": "/dev/disk/by-id/virtio-secondary",
+        "partitions": [
+          {
+            "label": "root1",
+            "number": 1,
+            "sizeMiB": 256,
```
krnowak, line 47: I see that we were lucky using `256MiB` for `size` before, so you didn't need to convert that from sectors to bytes.

krnowak (approved):

PR #236, opened 2021-09-23
kola/ignition/oem: exclude `esx` and `azure` from platforms

tormath1:
`wiping` the OEM file system does not allow the instance to boot. Let's
exclude the platform while we proceed to further investigations.

Related-to: https://github.com/flatcar-linux/Flatcar/issues/514

### DIFF
--- a/kola/tests/ignition/oem.go
+++ b/kola/tests/ignition/oem.go
@@ -67,6 +67,10 @@ func init() {
 		Run:         wipeOEM,
 		MinVersion:  semver.Version{Major: 2983},
 		ClusterSize: 1,
+		// `wiping` the OEM file system does not allow the instance to boot on platforms
+		// different from QEMU.
+		// More details: https://github.com/flatcar-linux/Flatcar/issues/514.
+		Platforms: []string{"qemu", "qemu-unpriv"},
 		UserData: conf.ContainerLinuxConfig(`storage:
   filesystems:
      - name: oem


krnowak:
I would be good to run this test on all platforms that we support to see the fallout. I suppose you could do it by creating your own branch in manifest repo that changes the `main.xml` file, so it has `revision` and `upstream` tags for mantle pointing to `refs/heads/tormath1/exclude-esx-oem`. Then you could use this branch instead of `flatcar-master` as MANIFEST_REF.

tormath1:
@krnowak that would be a great idea - but if I recall correctly, in the CI, we use `mantle` artifacts build from `mantle` jobs: example for AWS: https://github.com/kinvolk/jenkins-os/blob/5bb18e314ac7dab8e71038389fe4dd0bf09ccda3/os/kola/aws.groovy#L85-L88 so I'm not sure we can specify a custom `mantle` branch to use in the CI context.

krnowak:
I must admit that I have yet no idea how this mantle stuff works in jenkins-os, because the master-builder points to the master branch of the mantle repo instead of flatcar-master.

Anyway, it just occurred to me that this test is already in flatcar master, so you could run a toolchains job with all the platforms listed and see for which platform the oem.wipe test failed without doing this manifest dance.

kola/tests/ignition/oem.go
```diff
@@ -67,6 +67,9 @@ func init() {
 		Run:         wipeOEM,
 		MinVersion:  semver.Version{Major: 2983},
 		ClusterSize: 1,
+		// `wiping` the OEM file system does not allow the instance to boot.
+		// More details: https://github.com/flatcar-linux/Flatcar/issues/514.
+		ExcludePlatforms: []string{"esx", "azure"},
```
pothos, line 72: Maybe limiting it to the qemu platform is the best because it doesn't require a working OEM partition. It also helps keeping the testing time down on the cloud platforms.

pothos (approved):
Thanks

PR #237, opened 2021-10-01
kola/test/locksmith: update the cert generation

tormath1:
With OpenSSLv3, there is a tiny regression in the cert generation. The
issuer entry is modified _after_ the check for self-signed certificate.

So it fails with:
```
subject=CN = localhost
Error with certificate - error 20 at depth 0
unable to get local issuer certificate
subject=CN = localhost
Error with certificate - error 21 at depth 0
unable to verify the first certificate
Error with certificate to be certified - should be self-signed
```

workaround is to use intermediate CSR and generate after the cert based
on this CSR and by providing the associated extension in order to not
lost the SANs.

## Testing done

* with OpenSSLv3 image from: https://github.com/flatcar-linux/coreos-overlay/pull/1305
* with alpha 2983.0.0

More details here: https://github.com/openssl/openssl/issues/16720
Related-to: https://github.com/flatcar-linux/coreos-overlay/pull/1305


jepio (approved):
looks good and thanks for reporting upstream

PR #238, opened 2021-10-04
sdk: Updated SDK URL

jonasclaes:
# Updated SDK URL

The URL to the SDK has been updated, since this was moved from Google Cloud to the Flatcar repo server.

## How to use

1. Build the cork SDK using `./build`
2. Create a new chroot development environment using `cork create --verbose --manifest-branch flatcar-alpha-2983.0.0`

## Testing done

I have run the command `cork create --verbose --manifest-branch flatcar-alpha-2983.0.0` and got the correct output. The files are being downloaded, verified and unpacked. The output is too long and irrelevant to post here.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -122,7 +122,7 @@ func init() {
 
 	creationFlags = pflag.NewFlagSet("creation", pflag.ExitOnError)
 	creationFlags.StringVar(&sdkUrlPath,
-		"sdk-url-path", "/flatcar-jenkins/sdk", "SDK URL path")
+		"sdk-url-path", "/sdk", "SDK URL path")
 	creationFlags.StringVar(&sdkVersion,
 		"sdk-version", "", "SDK version. Defaults to the SDK version in version.txt")
 	creationFlags.StringVar(&manifestURL,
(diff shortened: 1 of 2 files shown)


pothos:
Thanks, since it's also used before we publish the files on the mirror, we need to be able to keep GCS working, @tormath1 can chime in with something similar to your approach.

tormath1:
Hi @jonasclaes , many thanks for this contribution - as @pothos suggested, we need to be able to keep GCS working and as you can see, having a hardcoded URL is not really handy for customization purposes.

What do you think about adding a `--sdk-url` flag to the `cork` commands (`create` and `download`) with a default value equals to the one you provided: `mirror.release.flatcar-linux.net` ?

jonasclaes:
@tormath1 I think that would be useful, and prevent service disruptions in the future. Who will implement this? Shall I do it, or will one of the other contributors do that?

tormath1:
@jonasclaes feel free to work it on if you're interested in - besides this repo is now part of the hacktoberfest. :tada: 
I provided some elements to guide you into the codebase but it's mainly in the same place you've already played. Don't hesitate to ping on Matrix if you need anything. :)

PR #241, opened 2021-10-05
*: Change project name from coreos to flatcar-linux

krnowak:
This just renames the project from `github.com/coreos/mantle` to `github.com/flatcar-linux/mantle`.


platform/api/packet/api.go
```diff
@@ -166,7 +166,7 @@ func New(opts *Options) (*API, error) {
 		return nil, fmt.Errorf("connecting to Google Storage bucket: %v", err)
 	}
 
-	client := packngo.NewClientWithAuth("github.com/coreos/mantle", opts.ApiKey, nil)
+	client := packngo.NewClientWithAuth("github.com/flatcar-linux/mantle", opts.ApiKey, nil)
```
krnowak, line 169: I'm not sure if this change can break access to packet API. Should this stay as `github.com/coreos/mantle`?

tormath1: It looks like a token to identify the generated client in order to keep it across the codebase (and avoid generating new client each time?). 
Checking the codebase: https://github.com/packethost/packngo/blob/b666b4ebebb1f0fba2a9ad79428fa79daf0774e2/packngo.go#L334-L342

a regular `NewClient` will populate this field with `packngo lib` so I guess we can put whatever we want inside.

krnowak:
Thanks for the review. Dropped the file and updated the change log.

PR #250, opened 2021-11-10
kola: remove legacy `kubernetes` test

tormath1:
These tests are now deprecated due to the stabilization of Docker >= 20.10.


krnowak:
Is there a way to make those tests compatible with docker 20.10? Or is this pointless?

tormath1:
> Is there a way to make those tests compatible with docker 20.10? Or is this pointless?

@krnowak that's indeed a good question - `kubernetes` tests are now superseded by `kubeadm` one, which rely directly on `docker` container runtime and test more up-to-date `Kubernetes` maintained versions (1.21, 1.22 and soon 1.23)

krnowak:
So these kubernetes tests were testing more-or-less the same thing as the new kubeadm tests? So bumping the kubernetes versions to actively supported ones in kubernetes test would result in a some kind of duplication? If that would be so, then I think it's fine to drop the tests.

tormath1:
@krnowak that's right.

Currently with `kubernetes` tests, we are testing the following versions:
https://github.com/flatcar-linux/mantle/blob/6336d5f143909cc080b95c4c4826b383ff3ac0c0/kola/tests/kubernetes/basic.go#L31-L36

While with `kubadm` we are testing[^1]:
* v1.22.0
* v1.21.0

On top of different tested versions, we can also mention that:
* we test `calico`, `flannel` and `cilium` CNIs with `kubeadm`
* as mentioned in the name `kubeadm` tests deploy Kubernetes using `kubeadm`

[^1]: https://github.com/flatcar-linux/mantle/blob/flatcar-master/kola/tests/kubeadm/kubeadm.go#L46-L69

krnowak:
Alright, let's drop the tests.

Also, I had a look at https://endoflife.date/kubernetes and I was wondering if kubeadm tests should also test version 1.20. But probably no point in it, as its active support expires in one month, and maintenance in 3. Also, different PR for it anyway.

krnowak (approved):

PR #253, opened 2021-11-19
RenderUserData: only incject SELinux enforcing if there is userdata

pothos:
Some tests use machines without userdata and in this case there is no
way to inject it and the test should rather use a minimal dummy
Ignition config to enforce SELinux if it's relevant.

Don't try to inject SELinux enforcement in case no userdata is present
because the function panics in this case.

## How to use

Should fix the panic for `systemd.journal.remote`

## Testing done

`make test`

### DIFF
--- a/platform/cluster.go
+++ b/platform/cluster.go
@@ -173,7 +173,7 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
 		conf.CopyKeys(keys)
 	}
 
-	if !bc.rconf.NoEnableSelinux {
+	if !bc.rconf.NoEnableSelinux && (conf.IsIgnition() || conf.IsCloudInit()) {
 		selinuxConf := `# This file controls the state of SELinux on the system on boot.
 # SELINUX can take one of these three values:
 #	enforcing - SELinux security policy is enforced.
--- a/platform/conf/conf.go
+++ b/platform/conf/conf.go
@@ -896,6 +896,10 @@ func (c *Conf) IsIgnition() bool {
 	return c.ignitionV1 != nil || c.ignitionV2 != nil || c.ignitionV21 != nil || c.ignitionV22 != nil || c.ignitionV23 != nil || c.ignitionV3 != nil
 }
 
+func (c *Conf) IsCloudInit() bool {
+	return c.cloudconfig != nil
+}
+
 func (c *Conf) IsEmpty() bool {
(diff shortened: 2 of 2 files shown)


tormath1 (approved):
Tested with a WIP test without user-data:
```
$ sudo ./bin/kola run --qemu-image ./flatcar_production_qemu_image.img bpf.local-gadget
=== RUN   bpf.local-gadget
=== RUN   bpf.local-gadget/filter_name_and_args
panic: unimplemented case in AddFile

goroutine 137 [running]:
github.com/flatcar-linux/mantle/platform/conf.(*Conf).AddFile(0xc0004e6320, 0x2551d56, 0x13, 0x2537922, 0x4, 0x25dafe9, 0x271, 0x1a4)
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/conf/conf.go:489 +0x546
github.com/flatcar-linux/mantle/platform.(*BaseCluster).RenderUserData(0xc000242540, 0xc00041d020, 0xc000657d50, 0xd, 0xc000657e48, 0x1)
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/cluster.go:192 +0x73e
github.com/flatcar-linux/mantle/platform/machine/qemu.(*Cluster).NewMachineWithOptions(0xc00025a510, 0xc00041d020, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/machine/qemu/cluster.go:61 +0x392
github.com/flatcar-linux/mantle/platform/machine/qemu.(*Cluster).NewMachine(0xc00025a510, 0xc00041d020, 0x0, 0x0, 0x0, 0x0)
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/machine/qemu/cluster.go:44 +0x46
github.com/flatcar-linux/mantle/platform.NewMachines.func1(0xc00024a2f0, 0x28be6f8, 0xc00025a510, 0xc00041d020, 0xc000385020, 0xc000384fc0)
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/platform.go:294 +0x82
created by github.com/flatcar-linux/mantle/platform.NewMachines
	/home/mathieu/go/src/github.com/kinvolk/mantle/platform/platform.go:292 +0x128
```

No error with the patch.

The only thing is that the enforcing of SELinux is conditionally set by the presence of user-data  but I think we can live with it for now.

PR #257, opened 2021-11-22
api/packet: change default ARM server

tormath1:
`c2.large.arm` are deprecated too and CI is failing with:
```
2021-11-22T09:31:56Z kola: creating new machine for semver check: couldn't create device: POST https://api.packet.net/projects/2c850bb4-70b4-4dcd-b60f-2d0be86d0ae4/devices: 422 No available servers with plan c2.large.arm in facility sv15
```

`c2.large.arm` is deprecated in favor of: `c3.large.arm`.

See also: https://metal.equinix.com/developers/docs/servers/server-specs/#c3largearm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - `lsblk --json` output handling ([#244](https://github.com/flatcar-linux/mantle/pull/244))
 - Flannel version to 0.14.0 ([#245](https://github.com/flatcar-linux/mantle/pull/245))
 - Renamed the project name from `github.com/coreos/mantle` to `github.com/flatcar-linux/mantle` ([#241](https://github.com/flatcar-linux/mantle/pull/241))
-- Default server for AMD64 on Equinix Metal ([#256](https://github.com/flatcar-linux/mantle/pull/256))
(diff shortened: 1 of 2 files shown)


pothos:
We also need to switch the region to Dallas or Washington DC for ARM64 in https://github.com/flatcar-linux/scripts/blob/main/jenkins/kola/packet.sh because these two are the only ones where hourly availability is: https://metal.equinix.com/product/servers/c3-large-arm/

dongsupark (approved):

tormath1:
> We also need to switch the region to Dallas or Washington DC for ARM64 in https://github.com/flatcar-linux/scripts/blob/main/jenkins/kola/packet.sh because these two are the only ones where hourly availability is: https://metal.equinix.com/product/servers/c3-large-arm/

@pothos good catch thanks - in this case, let's move everything to Dallas (`da11`) both instance types (`c3.small.x86` and `c3.large.arm`) are available within this facility.

pothos:
> > We also need to switch the region to Dallas or Washington DC for ARM64 in https://github.com/flatcar-linux/scripts/blob/main/jenkins/kola/packet.sh because these two are the only ones where hourly availability is: https://metal.equinix.com/product/servers/c3-large-arm/
> 
> @pothos good catch thanks - in this case, let's move everything to Dallas (`da11`) both instance types (`c3.small.x86` and `c3.large.arm`) are available within this facility.

For PXE booting it helps when the instances are in the same data center or city, I would not move c3.small.x86 to Dallas because it's almost 3000 km away - the exception for arm64 should also annotate that it is a temporary measure until there is c3.large.arm availability in SVJ/SV.

PR #259, opened 2021-11-24
kola/tests/etcd: add `name: ` to one CLC

tormath1:
Using both command line option and environment variable is forbidden
since `etcd/v3`.

It happens quite a lot that user override default `etcd` name
(`machine-id`) using `CLC` through the `name: ` key.
In this case it fails because we will have both `ETCD_NAME`
and `--name` in the runtime.

This patch overrides the name in CLC in order to reproduce this behavior
and prevent regression in the next release.

<hr>

This test is currently failing with current releases - it has to be merged with https://github.com/flatcar-linux/coreos-overlay/pull/1444.

Changelog entry is not required IMHO.

### DIFF
--- a/kola/tests/etcd/discovery.go
+++ b/kola/tests/etcd/discovery.go
@@ -16,8 +16,10 @@ package etcd
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/coreos/pkg/capnslog"
+	"github.com/pborman/uuid"
 
 	"github.com/flatcar-linux/mantle/kola/cluster"
 	"github.com/flatcar-linux/mantle/kola/register"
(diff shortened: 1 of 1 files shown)


krnowak (approved):
Looks good.

tormath1:
cherry-picked to:
* flatcar-3066
* flatcar-3033

krnowak:
@tormath1: Can this be merged or is it waiting for anything else?

tormath1:
@krnowak I was thinking about merging this after the next release. Otherwise nightlies are going to fail for `stable` since the patch has been ported only on current `beta` and `alpha` (since `beta` will be promoted as next `stable`)

krnowak:
Ah, cool. No problem, let's wait then.

PR #260, opened 2021-11-30
test: add bpf test with DNS gadget from local-gadget 

tormath1:
In this PR, we test `IG` and related BPF features on Flatcar with `local-gadget` starting with DNS gadget.

We basically install `local-gadget` and prepare its run using CLC.

It currently runs without SELinux due to this: https://github.com/flatcar-linux/Flatcar/issues/509

## How to use

```
./build kola
./bin/kola list | grep gadget
bpf.local-gadget			[all]								[all]		[cl]		[all]		[all]
```

## Testing done

```
=== RUN   bpf.local-gadget
=== RUN   bpf.local-gadget/dns_gadget
--- PASS: bpf.local-gadget (20.63s)
    --- PASS: bpf.local-gadget/dns_gadget (18.47s)
PASS, output in _kola_temp/qemu-2021-12-01-1126-8837
```

Related-to: https://github.com/flatcar-linux/Flatcar/issues/89

It could be merged with: https://github.com/flatcar-linux/mantle/pull/233 to kickoff BPF tests on Flatcar.

### DIFF
--- a/kola/tests/bpf/local-gadget.go
+++ b/kola/tests/bpf/local-gadget.go
@@ -1,0 +1,149 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package bpf
+
+import (
+	"bytes"
+	"strings"
+	"text/template"
+
+	"github.com/flatcar-linux/mantle/kola"
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+	"github.com/flatcar-linux/mantle/kola/tests/docker"
+	"github.com/flatcar-linux/mantle/platform/conf"
+)
+
+type gadget struct {
+	// Arch holds the gadget architecture (arm or amd).
+	Arch string
+	// Version holds the version of the gadget (v0.4.1 for example).
+	Version string
+	// Sum holds the SHA512 checksum to verify binary.
+	Sum string
+	// Cmd holds the command for the given gadget.
+	Cmd string
+}
+
+var (
+	// binaries holds the map of binaries for each supported
+	// architecture.
+	binaries = map[string]gadget{
+		"arm64": gadget{
+			Version: "0.4.1",
+			Sum:     "4b5761dd08afea378e7e58a5a76b76c727ed59d327e042a42dd72330cac7bcd516da7574f69d22a4076c07dcafa639d08aaced56cf624816c5815226b33a0961",
+		},
+		"amd64": gadget{
+			Version: "0.4.1",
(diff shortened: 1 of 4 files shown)


kola/tests/bpf/local-gadget.go
```diff
@@ -0,0 +1,160 @@
...
+			c.Fatalf("creating node: %v", err)
+		}
+
+		docker.GenDockerImage(c, node, "dig", []string{"dig"})
+
+		if _, err := c.SSH(node, "docker run --rm --name shell01 dig dig flatcar-linux.org"); err != nil {
+			c.Fatalf("unable to run docker cmd: %v", err)
+		}
+
+		out, err := c.SSH(node, "cat /tmp/local-gadget.res  | grep pkt_type | head --lines 1 | jq -r .name")
```
pothos, line 149:
```suggestion
		out, err := c.SSH(node, "grep -m 1 pkt_type /tmp/local-gadget.res | jq -r .name")
```
This makes the processes run without error, not a real problem but if it gets copied later where pipefail is set, grep will fail when head terminates earlier because it does not consume the full input

kola/tests/bpf/local-gadget.go
```diff
@@ -0,0 +1,160 @@
...
+        [Install]
+        WantedBy=multi-user.target`
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:     localGadgetTest,
+		Name:    `bpf.local-gadget`,
+		Distros: []string{"cl"},
+		// required while SELinux policy is not correcly updated to support
```
pothos, line 111: let's see if it works on the current LTS, otherwise we can add a minimum version later

pothos (approved):

tormath1:
@pothos thanks for this quick review :joy: but it's still WIP, currently adding seccomp-gadget test.

kola/tests/bpf/local-gadget.go
tormath1: Thanks for the tip.

kola/tests/bpf/local-gadget.go
```diff
@@ -0,0 +1,160 @@
+// Copyright © Microsoft Corporation
```
dongsupark, line 1: I would prefer to have a neutral term like `Copyright 2021 The mantle Authors` in https://github.com/flatcar-linux/mantle/blob/0282d298e43b0a99e17faaf9a0396d10a6fab4e0/kola/tests/util/template.go#L1 .

tormath1: Added the following headers:
```
// Copyright The Mantle Authors.
// SPDX-License-Identifier: Apache-2.0
```

dongsupark: Thanks for updating the headers. Looks good.

dongsupark (approved):

PR #268, opened 2021-12-14
equinix-metal: recycle existing instances

tormath1:
In this PR, we add the ability to mantle test to reuse existing Equinix Metal instances and to keep them into a pool of devices in the Flight lifecycle.

<hr>

Closes: https://github.com/flatcar-linux/Flatcar/issues/551

### DIFF
--- a/platform/api/packet/api.go
+++ b/platform/api/packet/api.go
@@ -184,7 +184,7 @@ func (a *API) PreflightCheck() error {
 }
 
 // console is optional, and is closed on error or when the device is deleted.
-func (a *API) CreateDevice(hostname string, conf *conf.Conf, console Console) (*packngo.Device, error) {
+func (a *API) CreateOrUpdateDevice(hostname string, conf *conf.Conf, console Console, id string) (*packngo.Device, error) {
 	consoleStarted := false
 	defer func() {
 		if console != nil && !consoleStarted {
@@ -212,7 +212,7 @@ func (a *API) CreateDevice(hostname string, conf *conf.Conf, console Console) (*
 	}
 	defer a.bucket.Delete(context.TODO(), ipxeScriptName)
 
-	device, err := a.createDevice(hostname, ipxeScriptURL)
+	device, err := a.createDevice(hostname, ipxeScriptURL, id)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create device: %v", err)
 	}
@@ -247,6 +247,15 @@ func (a *API) CreateDevice(hostname string, conf *conf.Conf, console Console) (*
 	if err != nil {
 		a.DeleteDevice(deviceID)
 		return nil, fmt.Errorf("timed out waiting for flatcar-install: %v", err)
+	}
+
+	// TCP discard service has been reached so `flatcar-install` is done.
+	// We can deactivate `PXE` boot to avoid bootlooping.
+	alwaysPXE := false
+	if _, _, err = a.c.Devices.Update(deviceID, &packngo.DeviceUpdateRequest{
+		AlwaysPXE: &alwaysPXE,
+	}); err != nil {
+		return nil, fmt.Errorf("unable to deactivate PXE boot: %v", err)
 	}
 
 	plog.Debugf("Finished installation of device: %q", deviceID)
@@ -478,22 +487,58 @@ boot`, a.opts.InstallerImageKernelURL, userdataURL, linuxConsole[a.opts.Board], 
 }
 
 // device creation seems a bit flaky, so try a few times
-func (a *API) createDevice(hostname, ipxeScriptURL string) (device *packngo.Device, err error) {
+func (a *API) createDevice(hostname, ipxeScriptURL, id string) (*packngo.Device, error) {
+	var err error
+
+	// we force a PXE boot in order to fetch the
+	// new configuration and prevent to boot from a mis-installed Flatcar.
+	alwaysPXE := true
+
 	for tries := apiRetries; tries >= 0; tries-- {
-		var response *packngo.Response
-		device, response, err = a.c.Devices.Create(&packngo.DeviceCreateRequest{
-			ProjectID:     a.opts.Project,
-			Facility:      []string{a.opts.Facility},
-			Plan:          a.opts.Plan,
-			BillingCycle:  "hourly",
-			Hostname:      hostname,
-			OS:            "custom_ipxe",
-			IPXEScriptURL: ipxeScriptURL,
-			Tags:          []string{"mantle"},
-		})
-		if err == nil || response.StatusCode != 500 {
-			return
+		var (
+			device   *packngo.Device
+			response *packngo.Response
+		)
+
+		if id != "" {
+			plog.Infof("Recycling instance: %s", id)
+			device, response, err = a.c.Devices.Update(id, &packngo.DeviceUpdateRequest{
+				AlwaysPXE:     &alwaysPXE,
+				IPXEScriptURL: &ipxeScriptURL,
+				Hostname:      &hostname,
+			})
+			if err != nil {
+				err = fmt.Errorf("updating device: %w", err)
+				continue
+			}
+
+			// we reboot the instance to apply the changes.
+			response, err = a.c.Devices.Reboot(id)
+			if err != nil {
+				err = fmt.Errorf("rebooting device: %w", err)
+				continue
+			}
+
+			plog.Infof("device rebooted: %s", id)
+		} else {
+			plog.Infof("Recycling is not possible, creating a new instance")
+			device, response, err = a.c.Devices.Create(&packngo.DeviceCreateRequest{
+				ProjectID:     a.opts.Project,
+				Facility:      []string{a.opts.Facility},
+				Plan:          a.opts.Plan,
+				BillingCycle:  "hourly",
+				Hostname:      hostname,
+				OS:            "custom_ipxe",
+				IPXEScriptURL: ipxeScriptURL,
+				Tags:          []string{"mantle"},
+				AlwaysPXE:     alwaysPXE,
+			})
 		}
+
+		if err == nil || response.StatusCode != 500 {
+			return device, err
+		}
+
 		plog.Debugf("Retrying to create device after failure: %q %q %q \n", device, response, err)
 		if device != nil && device.ID != "" {
 			a.DeleteDevice(device.ID)
@@ -502,7 +547,8 @@ func (a *API) createDevice(hostname, ipxeScriptURL string) (device *packngo.Devi
 			time.Sleep(apiRetryInterval)
 		}
 	}
-	return
+
+	return nil, fmt.Errorf("reached maximum number of retries to create/update a device: %w", err)
 }
 
 func (a *API) startConsole(deviceID string, console Console) error {
@@ -544,6 +590,7 @@ func (a *API) startConsole(deviceID string, console Console) error {
 	}
 	go func() {
 		err := runner()
+
 		if err != nil {
 			ready <- err
 		}
--- a/platform/machine/packet/cluster.go
+++ b/platform/machine/packet/cluster.go
@@ -98,37 +98,40 @@ func (pc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
 		mach.publicIP = pc.flight.api.GetDeviceAddress(device, 4, true)
 		mach.privateIP = pc.flight.api.GetDeviceAddress(device, 4, false)
 		if mach.publicIP == "" || mach.privateIP == "" {
-			mach.Destroy()
+			pc.flight.api.DeleteDevice(mach.ID())
 			err = fmt.Errorf("couldn't find IP addresses for device")
 			continue // provisioning error
 		}
 
+		// Warning: the assumption is that within one test a machine doesn't get reused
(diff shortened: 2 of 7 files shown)


platform/api/packet/api.go
```diff
@@ -478,22 +478,46 @@ boot`, a.opts.InstallerImageKernelURL, userdataURL, linuxConsole[a.opts.Board],
...
+			response *packngo.Response
+			err      error
+		)
+
+		if id != "" {
+			plog.Infof("Recycling instance: %s", id)
+			// we update the device and we force a PXE boot in order to fetch the
+			// new configuration.
+			alwaysPXE := true
+			device, response, err = a.c.Devices.Update(id, &packngo.DeviceUpdateRequest{
```
pothos, line 494: we have to device what to do when the update fails, currently it deletes the device and creates a new one, this is not wanted but instead we want to retry calling this function

pothos: (currently the error is discarded but even if not this is a problem)

platform/api/packet/api.go
```diff
@@ -478,22 +478,46 @@ boot`, a.opts.InstallerImageKernelURL, userdataURL, linuxConsole[a.opts.Board],
...
+				AlwaysPXE:     &alwaysPXE,
+				IPXEScriptURL: &ipxeScriptURL,
+			})
+			plog.Infof("device updated: %s", id)
+
+			// we reboot the instance to apply the changes.
+			response, err = a.c.Devices.Reboot(id)
```
pothos, line 501: same

platform/api/packet/api.go
pothos: An other alternative is to simplify this logic and directly terminate here without retry :)

platform/machine/packet/flight.go
```diff
@@ -94,4 +99,10 @@ func (pf *flight) Destroy() {
 	}
 
 	pf.BaseFlight.Destroy()
+
+	for id := range pf.devicesPool {
```
pothos, line 103:
```suggestion
	close(pf.devicesPool)
	for id := range pf.devicesPool {
```
Otherwise it will deadlock and wait endlessly.
The task here is to delete all devices that got created. This function is called when all tests are done and no new machines get created but we should close the channel to make sure that this is the case (instead of working around the deadlock with a possible select or similar).

platform/machine/packet/flight.go
```diff
@@ -49,8 +53,9 @@ func NewFlight(opts *packet.Options) (platform.Flight, error) {
 	}
 
 	pf := &flight{
-		BaseFlight: bf,
-		api:        api,
+		BaseFlight:  bf,
+		api:         api,
+		devicesPool: make(chan string),
```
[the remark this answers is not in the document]
tormath1: In that case, we could try to estimate the number of instances - "worst scenario" we can take `number of tests x 3`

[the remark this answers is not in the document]
pothos: Some arbitrary high number should also work

platform/api/packet/api.go
```diff
@@ -478,22 +489,46 @@ boot`, a.opts.InstallerImageKernelURL, userdataURL, linuxConsole[a.opts.Board],
...
+
+		if id != "" {
+			plog.Infof("Recycling instance: %s", id)
+			// we update the device and we force a PXE boot in order to fetch the
+			// new configuration.
+			alwaysPXE := true
+			device, response, err = a.c.Devices.Update(id, &packngo.DeviceUpdateRequest{
+				AlwaysPXE:     &alwaysPXE,
+				IPXEScriptURL: &ipxeScriptURL,
+			})
```
pothos, line 508:
```suggestion
			})
			if err != nil {
				return nil, err
			}
```

platform/api/packet/api.go
```diff
@@ -478,22 +489,46 @@ boot`, a.opts.InstallerImageKernelURL, userdataURL, linuxConsole[a.opts.Board],
...
+			plog.Infof("device updated: %s", id)
+
+			// we reboot the instance to apply the changes.
+			response, err = a.c.Devices.Reboot(id)
```
pothos, line 512:
```suggestion
			response, err = a.c.Devices.Reboot(id)
			if err != nil {
				return nil, err
			}
```

platform/api/packet/api.go
```diff
@@ -249,6 +249,17 @@ func (a *API) CreateDevice(hostname string, conf *conf.Conf, console Console) (*
 		return nil, fmt.Errorf("timed out waiting for flatcar-install: %v", err)
 	}
 
+	if id != "" {
```
pothos, line 252: For consistency and easier debugging if things go wrong depending on first use vs reuse, we could set AlwaysPXE from the beginning and rely on this code path here to unconditionally disable it again.

pothos: (not a blocker)

platform/api/packet/api.go
tormath1: `return` at this point will make leave the `createDevice` method direct and we will have some remaining instances into EM. They won't be deleted when the flight will leave since the instance would have been consumed from the `devicesPool` channel.

I think we can keep in this way, if we fail to `update` and / or `reboot` the existing device, we just loop over. However, we could provide a bit more of context into the error:
```suggestion
			})
			if err != nil {
			    err = errors.New("updating instance: %w", err)
			}
```

What do you think ?

pothos: ah, right, but to loop over you need to add a `continue` here

pothos: otherwise later `a.DeleteDevice(device.ID)` gets called

platform/api/packet/api.go
```diff
@@ -249,6 +249,17 @@ func (a *API) CreateDevice(hostname string, conf *conf.Conf, console Console) (*
...
+		// TCP discard service has been reached so `flatcar-install` is done.
+		// We can deactivate `PXE` boot to avoid bootlooping.
+		alwaysPXE := false
+		if _, _, err = a.c.Devices.Update(id, &packngo.DeviceUpdateRequest{
+			AlwaysPXE: &alwaysPXE,
+		}); err != nil {
+			plog.Errorf("unable to deactivate PXE boot: %v", err)
```
pothos, line 259:
```suggestion
			return nil, fmt.Errorf("unable to deactivate PXE boot: %v", err)
```
With PXE boot enabled it makes more sense to stop here or we might get strange errors later

platform/machine/packet/machine.go
```diff
@@ -65,8 +65,17 @@ func (pm *machine) Reboot() error {
...
+	// Instead of actually deleting the device.
+	// We add it to the devices pool in order to mark it
+	// as "ready to be used" by other tests.
+	id := pm.ID()
+	pm.cluster.flight.devicesPool <- id
+	plog.Infof("device %s added to the pool", id)
+
+	// console needs to be manually closed in order to prevent program from
+	// freezing on `done` channel in the `Output()` console's method.
+	if err := pm.console.Close(); err != nil {
```
pothos, line 77: I think we should close the session here instead of the console. The `startConsole` function waits on `session.Wait()` and will take care of closing the file writer and the console.

platform/machine/packet/console.go
```diff
@@ -38,7 +39,10 @@ func (c *console) Write(p []byte) (int, error) {
 }
 
 func (c *console) Close() error {
-	close(c.done)
+	if !c.closed {
```
pothos, line 42: see https://github.com/flatcar-linux/mantle/pull/268/files#r770517411 on why this here is not really needed

platform/machine/packet/machine.go
pothos: I realize that this needs a bit of rework in `startConsole` (`api/packet/api.go`) to expose a way of closing the session

platform/machine/packet/cluster.go
```diff
@@ -90,37 +98,37 @@ func (pc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
 		mach.publicIP = pc.flight.api.GetDeviceAddress(device, 4, true)
 		mach.privateIP = pc.flight.api.GetDeviceAddress(device, 4, false)
 		if mach.publicIP == "" || mach.privateIP == "" {
-			mach.Destroy()
+			pc.flight.api.DeleteDevice(mach.ID())
 			err = fmt.Errorf("couldn't find IP addresses for device")
 			continue // provisioning error
 		}
 
 		dir := filepath.Join(pc.RuntimeConf().OutputDir, mach.ID())
```
pothos, line 106:
```suggestion
		dir := filepath.Join(pc.RuntimeConf().OutputDir, mach.ID())
```
This here will overwrite the existing file, we should not use the machine ID here and rather need a suffix for it. I would like to see this directly be part of the `.ID()` function to be able to distinguish the machines reuse when tests log their ID. The `vmname` above already is a random string that we can append in the machine's `ID()` function.

Below for the deletion the real machine ID needs to be used and I suggest that this becomes a function on the machine struct.

tormath1: I think it's alright - indeed instances will have the same ID but they won't be in the same folder:
```
$ tree _kola_temp/packet-latest
_kola_temp/packet-latest
├── cl.basic
│   └── a279e8b2-14f4-451c-acd5-8705df9e70b5
│       ├── console.txt
│       ├── journal-raw.txt.gz
│       ├── journal.txt
│       └── user-data
├── cl.internet
│   └── a279e8b2-14f4-451c-acd5-8705df9e70b5
│       ├── console.txt
│       ├── journal-raw.txt.gz
│       ├── journal.txt
│       └── user-data
```

There is no chance from the same flight to have instances recycled under the same cluster I think.

Or I miss a usecase ?

pothos: Good, I don't see a test that destroys the machine/cluster it created and would create a new one afterwards. I think we are safe as long as this is the case.

platform/machine/packet/machine.go
tormath1: So I rewrote a bit in order to keep a map of opened SSH clients and we close the associated client when we mark the device as available.

platform/api/packet/api.go
pothos: If the first PXE boot fails for what ever reason (ipxe or initramfs network error), on the second boot the instance would try to boot from disk. We can make this more robust if we start with `AlwaysPXE: &true` and disable it here again in all cases.

platform/machine/packet/console.go
```diff
@@ -16,20 +16,44 @@ package packet
...
 type console struct {
-	pc   *cluster
-	f    *os.File
-	buf  bytes.Buffer
-	done chan interface{}
+	pc      *cluster
+	f       *os.File
+	buf     bytes.Buffer
+	done    chan interface{}
+	clients map[string]*ssh.Client
```
pothos, line 30: Why the map? - There is only one SSH client per serial console, this is done in `startConsole`.

tormath1: right - for one moment I thought the console was created _before_ the cluster - but it's actually created for each machine :)

platform/machine/packet/machine.go
```diff
@@ -65,8 +65,18 @@ func (pm *machine) Reboot() error {
...
+	// Instead of actually deleting the device.
+	// We add it to the devices pool in order to mark it
+	// as "ready to be used" by other tests.
+	id := pm.ID()
+	pm.cluster.flight.devicesPool <- id
+	plog.Infof("device %s added to the pool", id)
+
+	// SSH client needs to be manually closed in order to prevent program from
+	// freezing on `done` channel in the `Output()` console's method.
+	plog.Infof("closing %s SSH client", id)
```
pothos, line 77:
```suggestion
	plog.Infof("closing %s serial console SSH client", id)
```

platform/machine/packet/machine.go
```diff
@@ -65,8 +65,18 @@ func (pm *machine) Reboot() error {
...
+	if err := pm.console.CloseSSH(); err != nil {
+		plog.Errorf("closing SSH client: %v", err)
```
pothos, line 79:
```suggestion
		plog.Errorf("closing serial console SSH client: %v", err)
```

platform/machine/packet/machine.go
```diff
@@ -65,8 +65,18 @@ func (pm *machine) Reboot() error {
...
-	if err := pm.cluster.flight.api.DeleteDevice(pm.ID()); err != nil {
-		plog.Errorf("Error terminating device %v: %v", pm.ID(), err)
+	// Instead of actually deleting the device.
+	// We add it to the devices pool in order to mark it
+	// as "ready to be used" by other tests.
+	id := pm.ID()
+	pm.cluster.flight.devicesPool <- id
+	plog.Infof("device %s added to the pool", id)
+
+	// SSH client needs to be manually closed in order to prevent program from
```
pothos, line 75:
```suggestion
	// Ther serial console SSH client needs to be manually closed in order to prevent program from
```

pothos (approved):
Looks good now, didn't test myself ;)

tormath1:
@pothos 

I just started a final test:

```
--parallel=2 cl.basic kubeadm.v1.22.0.calico.base cl.internet kubeadm.v1.22.0.flannel.base
```

we should have less than 8 devices created.

PR #271, opened 2022-01-07
rework resource cleanup on Azure

jepio:
# rework resource cleanup on Azure

Rework cleanup of NIC/IP/disk resources on Azure to make it more reliable and a responsibility of the Azure Resource Manager. Thanks to this, OS disks are now deleted immediately with the VM. Also implement correct architecture detection on Azure.

## How to use

`kola run`

## Testing done

`kola run` completes slightly faster and tests find the correct `kolet` binary.

- [X] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)


jepio:
@tormath1 this is a followup to #270.

platform/api/azure/instance.go
```diff
@@ -29,6 +29,8 @@ import (
 	"github.com/flatcar-linux/mantle/util"
 )
 
+var forceDelete = true
```
tormath1, line 32: I think it can be safely defined as a constant:
```suggestion
const forceDelete = true
```

jepio: Tried that, it's not legal to take the address of a constant this way

tormath1 (approved):
thanks for the cleanup :racing_car:

kola/harness.go
```diff
@@ -577,6 +577,9 @@ func architecture(pltfrm string) string {
 	if pltfrm == "aws" && AWSOptions.Board != "" {
 		nativeArch = boardToArch(AWSOptions.Board)
 	}
+	if pltfrm == "azure" && AzureOptions.Board != "" {
```
krnowak, line 580: Oh, the saving of keystrokes. Extending lifetime of our dear keyboards. Heh.

(No need to change anything, just being amused here. :) )

krnowak (approved):

kola/harness.go
jepio: I think it's still too long, the `r` is a waste of space

PR #274, opened 2022-01-12
plume: Add new AWS regions, af-south-1, ap-southeast-3, eu-south-1

sayanchowdhury:
# plume: Add new AWS region, af-south-1

## Testing done
Locally built plume


cmd/plume/containerlinux.go
```diff
@@ -62,6 +62,7 @@ var (
 				"ap-southeast-2",
 				"ap-northeast-1",
 				"ap-northeast-2",
+				"af-south-1",
```
krnowak, line 65: Like in https://github.com/flatcar-linux/flatcar-build-scripts/pull/93, maybe add also "ap-southeast-3" and "eu-south-1"?

sayanchowdhury: Good catch, Done!

krnowak (approved):

PR #276, opened 2022-01-13
ore: prevent reusing the amd64 EBS snapshot

pothos:
The ore image name was defaulting to the version in contrast to plume
where we assemble a more complete name. This led to a reuse of the
amd64 EBS snapshot for arm64, or vice versa.

Add an arch suffix as done in plume (i.e., only for arm64 to keep the
existing name for amd64).

## Testing done

None


iaguis (approved):
lgtm

PR #280, opened 2022-01-21
kola: Ignore dead links in /run/udev/links

krnowak:
Previously the entries in the directory were like `rtc/c252:0`, where
`c252:0` was just an empty file. Starting from systemd 250 it became a
dead symlink which points to `-100:/dev/rtc0`. This has caused the
`cl.filesystem` test to fail. To fix it, we add another entry to the
ignore list.

Should fix an issue like follows:

```
  Error: "--- FAIL: cl.filesystem/deadlinks (0.42s)
            files.go:89: Dead symbolic links found: [
/run/udev/links/disk\\x2fby-id\\x2fata-QEMU_DVD-ROM_QM00003/b11:0
/run/udev/links/disk\\x2fby-path\\x2fpci-0000:00:01.1-ata-2.0/b11:0
/run/udev/links/disk\\x2fby-path\\x2fpci-0000:00:01.1-ata-2/b11:0
…
yadda yadda yadda
]"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Changed
 - removed `packet` occurrences in favor of `equinixmetal` ([#277](https://github.com/flatcar-linux/mantle/pull/277))
+- kola: fixed cl.filesystem test for systemd 250 and newer ([#280](https://github.com/flatcar-linux/mantle/pull/280))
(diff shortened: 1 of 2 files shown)


jepio (approved):

krnowak:
Currently running this PR in CI: http://localhost:9091/job/os/job/manifest/4652/cldsv/, let see if the fix works.

krnowak:
The fix seems to work - no qemu test failed.

PR #281, opened 2022-01-25
kola: add `Metro` attribute to Equinix-Metal device creation

tormath1:
In this PR, we upgrade `packgno` library and add the `Metro` attribute to the device creation - this should decrease the pressure on the cloud provider since it will dispatch the load all along the facilities availability.

See also: https://metal.equinix.com/developers/docs/locations/metros/ 

## Testing done

Tested using `--equinixmetal-facility` and `--equinixmetal-metro` both at the same time and tested with `--equinixmetal-metro DA` only.


jepio (approved):
LGTM - the diff is much smaller than I expected, nice.

PR #284, opened 2022-01-31
kola/test/ignition/fs: add swap_activation test

tormath1:
This test aims to reproduce swap activation behavior on FCL through CLC

<hr>

Related to:
* https://github.com/flatcar-linux/Flatcar/issues/133
* https://www.flatcar.org/docs/latest/setup/storage/adding-swap/#adding-swap-with-a-container-linux-config

## Testing done

Tested locally with Stable-3033.2.1 and Alpha-3127.0.0

### DIFF
--- a/kola/tests/ignition/filesystem.go
+++ b/kola/tests/ignition/filesystem.go
@@ -15,8 +15,10 @@
 package ignition
 
 import (
+	"fmt"
 	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/flatcar-linux/mantle/kola/cluster"
 	"github.com/flatcar-linux/mantle/kola/register"
 	"github.com/flatcar-linux/mantle/platform/conf"
@@ -25,6 +27,8 @@ import (
 const (
 	targetUUID   = "9aa5237a-ab6b-458b-a7e8-f25e2baef1a3"
 	targetVfatID = "1A37-8FA3"
+	swapFileName = "/var/vm/swapfile1"
+	swapSize     = "1024M"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


kola/tests/ignition/filesystem.go
```diff
@@ -253,6 +256,52 @@ func init() {
...
+        RequiresMountsFor=/var
+        ConditionPathExists=!%s
+
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/bin/mkdir -p /var/vm
+        ExecStart=/usr/bin/fallocate -l %s %s
+        ExecStart=/usr/bin/chmod 600 %s
+        ExecStart=/usr/sbin/mkswap %s
+        RemainAfterExit=true`, swapFileName, swapFileName, swapSize, swapFileName, swapFileName, swapFileName))
```
krnowak, line 296: How about replacing all `%s` with `%[1]s` and `%[2]s`, so the parameters passed to `Sprintf` would be just `swapFileName, swapSize`? (See [Explicit argument indexes](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes))

```suggestion
systemd:
  units:
    - name: var-vm-swapfile1.swap
      enabled: true
      contents: |
        [Unit]
        Description=Turn on swap
        Requires=create-swapfile.service
        After=create-swapfile.service

        [Swap]
        What=%[1]s

        [Install]
        WantedBy=multi-user.target
    - name: create-swapfile.service
      contents: |
        [Unit]
        Description=Create a swapfile
        RequiresMountsFor=/var
        ConditionPathExists=!%[1]s

        [Service]
        Type=oneshot
        ExecStart=/usr/bin/mkdir -p /var/vm
        ExecStart=/usr/bin/fallocate -l %[2]s %[1]s
        ExecStart=/usr/bin/chmod 600 %[1]s
        ExecStart=/usr/sbin/mkswap %[1]s
        RemainAfterExit=true`, swapFileName, swapSize))
```

kola/tests/ignition/filesystem.go
```diff
@@ -253,6 +256,52 @@ func init() {
...
+
+	register.Register(&register.Test{
+		Name:        "cl.swap_activation",
+		Run:         testSwapActivation,
+		ClusterSize: 1,
+		UserData:    swapActivation,
+		Distros:     []string{"cl"},
```
dongsupark, line 303: If I understand the [discussion](https://github.com/flatcar-linux/Flatcar/issues/133#issuecomment-1025515358) correctly, the systemd unit described on the doc started working out of the box since recent versions of systemd, like v249, right?
In that case, adding a test like that would fail with LTS 2605.x.y, as LTS still has systemd v246.
Maybe restrict the test like `MinVersion:  semver.Version{Major: 3033}`?

kola/tests/ignition/filesystem.go
tormath1: Ouh nice, did not about this. Thanks !

kola/tests/ignition/filesystem.go
tormath1: Good catch thanks, it should fail indeed with LTS so let's add a minimum version.

tormath1:
@krnowak @dongsupark thanks for your reviews. Suggestion have been addressed in a `fixup!` commit.

dongsupark (approved):

krnowak (approved):

kola/tests/ignition/filesystem.go
```diff
@@ -253,6 +257,53 @@ func init() {
...
+        What=%[1]s
+
+        [Install]
+        WantedBy=multi-user.target
+    - name: create-swapfile.service
+      contents: |
+        [Unit]
+        Description=Create a swapfile
+        RequiresMountsFor=/var
+        ConditionPathExists=!%[1]s
```
pothos, line 289: This makes this required unit fail to start which, e.g., could happen on reboots, I would move this check into `ExecStart`

pothos: (or just unconditionally run `fallocate` and `mkswap` again if it's idempotent and will succeed)

tormath1: `fallocate` and `mkswap` look idempotent - will remove the condition:
```
$ fallocate -l 1024m ./test
$ echo $?
0
$ chmod 600 ./test
$ mkswap ./test
Setting up swapspace version 1, size = 1024 MiB (1073737728 bytes)
no label, UUID=1b298276-fde1-4320-b4ee-092e04548db6
$ fallocate -l 1024m ./test
$ echo $?
0
$ mkswap ./test
mkswap: ./test: warning: wiping old swap signature.
Setting up swapspace version 1, size = 1024 MiB (1073737728 bytes)
no label, UUID=9e988d79-78cb-4e4e-9ee0-9535bc1a59f5
$ echo $?
0
```

kola/tests/ignition/filesystem.go
```diff
@@ -253,6 +257,53 @@ func init() {
...
+
+        [Swap]
+        What=%[1]s
+
+        [Install]
+        WantedBy=multi-user.target
+    - name: create-swapfile.service
+      contents: |
+        [Unit]
+        Description=Create a swapfile
```
pothos, line 287: I would add `DefaultDependencies=no` here (not on the .swap unit as suggested in the issue)

PR #286, opened 2022-02-04
platform/api/aws: use env credentials when no other creds are provided

jepio:
We updated the AWS SDK and hit by a change which caused environment
credentials to be ignored when a profile is passed. Details are here[1]
Explicily set credentials to NewEnvCredentials() to make this work
again. Without this, tests fail in kola:

  2022-02-04T01:19:43Z kola: creating flight for RunTests failed: NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors

[1]: https://github.com/aws/aws-sdk-go/issues/3382


tormath1 (approved):
Thanks ! Hoping it will be the only side effect. :)

PR #288, opened 2022-02-09
platform/api/equinixmetal: use httpS for PXE URL

tormath1:
PXE boot over httpS is now supported since this PR: https://github.com/flatcar-linux/Flatcar/issues/527

> We switched to ISRG Root X1 preferred chain, the handshake data is now short enough that iPXE can digest it.

## Testing done

* Run an amd64 test in Equinix Metal


jepio (commented):
Lines 56-60 at the top of `platform/api/equinixmetal/api.go also mention "HTTPS causes iPXE to fail"

pothos:
The URL used for testing is passed from the script environment and uses GCS (where the image to test is stored)

sayanchowdhury (approved):

PR #290, opened 2022-02-11
improve aws test runtime and logging

jepio:
# improve aws test speed and logging

Several improvements to the test runtime and logging:
* instances are force stopped before terminating
* we fetch "latest" console output, which is supported on Nitro. Without the "latest" option, console output is only updated on instance lifecycle events
* reduce the timeout for new console data - the 300s timeout for updated console output  stalls almost every test case on AWS. 30s wait should be enough
* rework retry handling when establishing a journalctl connection: 120 retries, 10s delay between retries, and the ssh timeout itself meant that some failing test cases spent 6000s (!!!) before timing out. Rework with to 60 retries with a bounded timeout of 600s.

## How to use

## Testing done

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	sshRetries = 120
+	sshRetries = 60
 	sshTimeout = 10 * time.Second
 )
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,10 +6,15 @@ properties([
     [$class: 'CopyArtifactPermissionProperty',
      projectNames: '*'],
 
-    pipelineTriggers([pollSCM('H/15 * * * *')])
+    pipelineTriggers([pollSCM('H/15 * * * *')]),
+    parameters([
+        booleanParam(name: 'ARCHIVE_ARTIFACTS', defaultValue: false),
+        booleanParam(name: 'CLEAN', defaultValue: true)
+    ])
 ])
 
 node('amd64 && docker') {
+  try {
     stage('SCM') {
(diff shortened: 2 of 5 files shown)


platform/platform.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	sshRetries = 120
+	sshRetries = 60
```
pothos, line 35: Is this still enough for Equinix Metal? I think yes because it starts from the reboot of the PXE installer (there is a waiter first to check for the end of the PXE install) and I expect a single reboot to be done in less than 10 minutes.

jepio: Im curious as well - I think it should be on the safe side, but I'm willing to increase it to 90 retries? The important thing is that we include the ssh timeout in the calculation, 100 minutes timeout for a test case where the whole suite takes <1hr is madness. We'll know more tomorrow after all tests have run.

jepio:
This is being tested here: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4842/cldsv/ (all builds reference a mantle branch that includes these patches, jepio/lbunzip2).

platform/platform.go
pothos: Let's leave it small then :)

jepio:
Test run was successful with a normal amount of flakes. I'm ready to merge this.

jepio:
Made one more change to how the `lastErr` is extracted from the `WaitUntilReady` callback, that wasn't returning the error as intended (though this is only informational).

pothos (approved):

PR #293, opened 2022-02-15
platform/local/etcd: bump dependency 

tormath1:
In this PR, we bump `etcd` dependency to pull protobuf 1.5.2 which should solve the issue raised in the CI:
```
panic: field [etcdserverpb.Request.ID](http://etcdserverpb.request.id/) has invalid type: got uint64, want pointer
```

## Note for reviewers

* Cobra has been upgraded by the same time to a tagged version
* `github.com/coreos/etcd` has been sed to `go.etcd.io/etcd`
* Nothing relevant about the codebase updating, I just have one doubt regarding the logging of etcdserver: currently set to error level which seems to me the most relevant (it was previously set by the capsnlog package). 

## Testing done

* tested locally with `coreos.tls.fetch-urls` but let's give a try to the new Jenkins feature to run CI with custom mantle :D
* CI (:large_blue_circle:): http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4860/cldsv/


jepio (approved):
Haven't reviewed the diff in detail but it seems to fix the test failures so I'm ok with merging.

PR #295, opened 2022-02-18
azure private networking

jepio:
# azure private networking

Add kola support for creating instances in an existing virtual network/subnet, and addressing them using private IP addresses. This assumes that we are either also located in the specified virtual network or can otherwise access it via VPN.

The `--azure-use-private-ips` option only makes sense when used together with the `--azure-vnet-subnet-name`, as otherwise kola creates the vnet and it won't be peered or have a VPN configured.

## How to use

```
./bin/kola spawn --platform azure --verbose --azure-location <location> --azure-publisher kinvolk --azure-offer flatcar-container-linux-free  --azure-sku alpha --azure-use-private-ips=true --azure-vnet-subnet-name <vnet-name>/<subnet-name>
```

## Testing done

Ran the command from the "how-to-use" section.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -107,7 +107,9 @@ func init() {
 	sv(&kola.AzureOptions.Location, "azure-location", "westus", "Azure location (default \"westus\"")
 	sv(&kola.AzureOptions.Size, "azure-size", "Standard_DS2_v2", "Azure machine size (default \"Standard_DS2_v2\")")
 	sv(&kola.AzureOptions.HyperVGeneration, "azure-hyper-v-generation", "V1", "Azure Hyper-V Generation (\"V1\" or \"V2\")")
+	sv(&kola.AzureOptions.VnetSubnetName, "azure-vnet-subnet-name", "", "Use a pre-existing virtual network for created instances. Specify as vnet-name/subnet-name. Subnet name may be omitted \"default\" is assumed")
 	bv(&kola.AzureOptions.UseGallery, "azure-use-gallery", false, "Use gallery image instead of managed image")
+	bv(&kola.AzureOptions.UsePrivateIps, "azure-use-private-ips", false, "Assume nodes are reachable using private ip addresses")
 
 	// do-specific options
 	sv(&kola.DOOptions.ConfigPath, "do-config-file", "", "DigitalOcean config file (default \"~/"+auth.DOConfigPath+"\")")
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -107,9 +107,9 @@ func init() {
 	sv(&kola.AzureOptions.Location, "azure-location", "westus", "Azure location (default \"westus\"")
 	sv(&kola.AzureOptions.Size, "azure-size", "Standard_DS2_v2", "Azure machine size (default \"Standard_DS2_v2\")")
 	sv(&kola.AzureOptions.HyperVGeneration, "azure-hyper-v-generation", "V1", "Azure Hyper-V Generation (\"V1\" or \"V2\")")
-	sv(&kola.AzureOptions.VnetSubnetName, "azure-vnet-subnet-name", "", "Use a pre-existing virtual network for created instances. Specify as vnet-name/subnet-name. Subnet name may be omitted \"default\" is assumed")
+	sv(&kola.AzureOptions.VnetSubnetName, "azure-vnet-subnet-name", "", "Use a pre-existing virtual network for created instances. Specify as vnet-name/subnet-name. If subnet name is omitted then \"default\" is assumed")
 	bv(&kola.AzureOptions.UseGallery, "azure-use-gallery", false, "Use gallery image instead of managed image")
-	bv(&kola.AzureOptions.UsePrivateIps, "azure-use-private-ips", false, "Assume nodes are reachable using private ip addresses")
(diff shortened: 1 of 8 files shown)


platform/machine/azure/flight.go
```diff
@@ -93,8 +94,9 @@ func NewFlight(opts *azure.Options) (platform.Flight, error) {
 			return nil, err
 		}
 
-		_, err = af.api.PrepareNetworkResources(af.ImageResourceGroup)
+		af.Network, err = af.api.PrepareNetworkResources(af.ImageResourceGroup)
 		if err != nil {
+			af.Destroy()
```
krnowak, line 99: Do you still need to destroy the network if getting it resulted in failure? I'd say that PrepareNetworkResources should take care of it.

platform/api/azure/network.go
```diff
@@ -26,19 +27,65 @@ import (
...
-		return network.Subnet{}, err
+		return Network{}, err
 	}
 
-	return a.createSubnet(resourceGroup)
+	_, err := a.createSubnet(resourceGroup)
+	if err != nil {
+		return Network{}, err
+	}
+	subnet, err := a.getSubnet(resourceGroup, kolaVnet, kolaSubnet)
```
pothos, line 82:
```suggestion
	subnet, err := a.createSubnet(resourceGroup)
	if err != nil {
		return Network{}, err
	}
```
Would this here work, too?

platform/machine/azure/flight.go
jepio: I should expand the commit message: if `PrepareNetworkResources` fails because you passed a wrong vnet/subnet name, then you want the resource group + storage account to be deleted, that's why I added the `Destroy()` here.

platform/api/azure/instance.go
```diff
@@ -70,13 +71,19 @@ func (a *API) getVMParameters(name, userdata, sshkey, storageAccountURI string,
 			Sku:       &a.opts.Sku,
 			Version:   &a.opts.Version,
 		}
+		plan = &compute.Plan{
+			Publisher: imgRef.Publisher,
```
pothos, line 75: Is this always valid given any disk URI or should it only be set when the else case above is hit?

platform/api/azure/network.go
jepio: Yes that should work and it's much better, thanks!

platform/api/azure/instance.go
jepio: It's definitely valid here, but would also be valid if passed an uri to a gallery image disk. For images/VHD uris there is no attached plan information.

I don't know if there is a way to lookup the plan from an arbitrary disk uri.

pothos: Above imfRef is set by a.opts.DiskURI and the plan values are empty then? Maybe it makes sense to have plan be nil in that case. If it's not a problem for the API, we can leave it as is.

platform/machine/azure/flight.go
krnowak: Ah, for some reason I also thought that it was calling Destroy on the just returned network. Nevermind.

platform/api/azure/network.go
```diff
@@ -26,19 +27,64 @@ import (
...
-		return network.Subnet{}, err
+		return Network{}, err
 	}
 
-	return a.createSubnet(resourceGroup)
+	subnet, err := a.createSubnet(resourceGroup)
+	if err != nil {
+		return Network{}, err
+	}
+	return Network{subnet}, err
```
krnowak, line 82:
```suggestion
	return Network{subnet}, nil
```

platform/api/azure/options.go
```diff
@@ -35,7 +35,9 @@ type Options struct {
 	Size             string
 	Location         string
 	HyperVGeneration string
+	VnetSubnetName   string
 	UseGallery       bool
+	UsePrivateIps    bool
```
krnowak, line 40:
```suggestion
	UsePrivateIPs    bool
```

krnowak (approved):
Looks good, I only had nitpicks around capitalization.

platform/api/azure/instance.go
pothos: Ok, was just confused by the alignment here in GitHub^^

cmd/kola/options.go
```diff
@@ -107,7 +107,9 @@ func init() {
 	sv(&kola.AzureOptions.Location, "azure-location", "westus", "Azure location (default \"westus\"")
 	sv(&kola.AzureOptions.Size, "azure-size", "Standard_DS2_v2", "Azure machine size (default \"Standard_DS2_v2\")")
 	sv(&kola.AzureOptions.HyperVGeneration, "azure-hyper-v-generation", "V1", "Azure Hyper-V Generation (\"V1\" or \"V2\")")
+	sv(&kola.AzureOptions.VnetSubnetName, "azure-vnet-subnet-name", "", "Use a pre-existing virtual network for created instances. Specify as vnet-name/subnet-name. Subnet name may be omitted \"default\" is assumed")
```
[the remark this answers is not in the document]
jepio: I phased it like this: `If subnet name is omitted then \"default\" is assumed`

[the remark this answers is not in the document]
krnowak: Sounds good.

pothos:
> @pothos any further suggestions before merging?

no, all good

PR #297, opened 2022-02-21
kubeadm: bump versions

tormath1:
in this PR, we officially bump Kubernetes tested versions. From what I understood, Kubernetes was already pulling stable version for the images, so it's only an update of the binaries (`kubeadm`, `kubelet`, etc.).

CNI version has been upgraded to 1.0.1 - since 1.0.0 Flannel plugin has been moved to its own repository and it's being installed by an init-container in the `kube-flannel.yml` plugin.

This leads to an issue when trying to copy CNI plugin from the container to `/opt/cni` because from a SELinux PoV labels are mismatching between hosts and container and this can be added to Flatcar virt policy in a second time. (see also: https://github.com/flatcar-linux/Flatcar/issues/635) 

The latest commit `kubeadm/template: update /opt/cni SELinux label type` can be dropped once the patch has been applied on every channel.

## Testing done

- [x] Jenkins :large_blue_circle: (http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4911/cldsv/)
- [x] Tested manually on EM with `kubeadm.v1.23.4.flannel.base`
<hr>


jepio (approved):
The change itself is good, but I wonder about `/usr/bin/chcon -R /opt/cni -t svirt_lxc_file_t`: is this likely something users might also need to set when they deploy with kubeadm? If so then we should add it to our selinux docs (if we have some, otherise let's start some :) )

tormath1:
@jepio - thanks for the review ! There is a SELinux doc (a bit outdated but not on the commands: https://www.flatcar.org/docs/latest/setup/security/selinux/).
 >  is this likely something users might also need to set when they deploy with kubeadm.

I would say "Yes" under these two conditions:
- SELinux in enforcing mode (Flatcar is shipped in permissive mode)
- Flannel CNI >= 0.15.0

> If so then we should add it to our selinux docs

I opened a GH issue to track that: https://github.com/flatcar-linux/Flatcar/issues/635, this is something we should fix in the OS. The issue also mentions the mitigation, so I'm not sure it requires a dedicated documentation. What do you think ?

jepio:
Then atleast link the issue from the limitations section?

tormath1:
@jepio done in: https://github.com/flatcar-linux/flatcar-docs/pull/212. Merging this PR.

PR #298, opened 2022-02-23
kola/tests: Add test for runtime switching to cgroupv1

jepio:
# kola/tests: Add test for runtime switching to cgroupv1

Add test illustrating switching to cgroupv1 with no reboot.

## How to use

Run test with this PR: https://github.com/flatcar-linux/coreos-overlay/pull/1666

## Testing done

CI running

### DIFF
--- a/kola/tests/kubeadm/kubeadm.go
+++ b/kola/tests/kubeadm/kubeadm.go
@@ -21,6 +21,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/pkg/capnslog"
 
 	"github.com/flatcar-linux/mantle/kola"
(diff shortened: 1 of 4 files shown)


pothos:
I think we need a more high-level test, too. I think a variant of the kubeadm tests (maybe just one specific configuration) would be helpful to assert that a system booted up this way is actually able to run Kubernetes. That's the only concern I have, otherwise the changes look good together.

tormath1:
@pothos good idea for the kubadm test. I think we can easily add a `cgroupv1` boolean field to one of the tested version to conditionally add the `/etc/flatcar-cgroupv1` file to the provisioning.

tormath1:
@jepio test is looking good to me. Can you maybe add a changelog entry in the `added` section ?

kola/tests/kubeadm/kubeadm.go
```diff
@@ -124,27 +127,42 @@ etcd:
...
+					cgroupSuffix = ".cgroupv1"
+				}
+
+				register.Register(&register.Test{
+					Name:             fmt.Sprintf("kubeadm.%s.%s%s.base", version, CNI, cgroupSuffix),
+					Distros:          []string{"cl"},
+					ExcludePlatforms: []string{"esx"},
+					Run: func(c cluster.TestCluster) {
+						kubeadmBaseTest(c, testParams)
+					},
```
tormath1, line 159: As in previous commit, I think we need a minimum version ?

jepio: I've added a minversion as well.

jepio:
@tormath1 @pothos I've added tests for cgroupv1+k8s v1.21.*. Search for cgroupfs in the console output here: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu/3761/console.

pothos (approved):

dongsupark (approved):

PR #300, opened 2022-03-02
Jenkinsfile: keep artifacts from 3 latest builds

jepio:
Artifacts take space and we only use the latest artifacts when running tests so there is
no need for us to store more than the last 3.


pothos (approved):
Thanks!

PR #301, opened 2022-03-03
add some ignition/v3 test

tormath1:
In this PR, we add some Ignition V3 tests from `coreos/coreos-assembler` project in order to be closer from the upstream. It also brings new test for IgnitionV3 on top of `UserDataV3` ones.

Tested with: http://localhost:9091/job/os/job/manifest/5049/cldsv/

_NOTE_: `kargs` test could be upstreamed


kola/tests/ignition/kernel.go
```diff
@@ -0,0 +1,39 @@
...
+			"contents": {
+			  "source": "data:,helloworld%0A",
+			  "verification": {}
+			},
+			"mode": 420
+		      }
+		    ]
+		  },
+		  "kernelArguments": {
+		    "shouldExist": ["flatcar.first_boot=detected"]
```
pothos, line 33:
```suggestion
		    "shouldExist": ["flatcar.first_boot=detected"]
```
I would use something like `flatcar.autologin` because it is harmless compared to forcing to run ignition this way which can maybe cause a reboot to fail due to not being able to remove the flag file

platform/conf/conf.go
```diff
@@ -893,7 +1022,7 @@ func keysToStrings(keys []*agent.Key) (keyStrs []string) {
 // Returns false in the case of empty configs as on most platforms,
 // this will default back to cloudconfig
 func (c *Conf) IsIgnition() bool {
-	return c.ignitionV1 != nil || c.ignitionV2 != nil || c.ignitionV21 != nil || c.ignitionV22 != nil || c.ignitionV23 != nil || c.ignitionV3 != nil
+	return c.ignitionV1 != nil || c.ignitionV2 != nil || c.ignitionV21 != nil || c.ignitionV22 != nil || c.ignitionV23 != nil || c.ignitionV3 != nil || c.ignitionV31 != nil || c.ignitionV32 != nil || c.ignitionV33 != nil
```
pothos, line 1025: if this here needs to be extended with V31, V32, V33 I think we also have to do it for `AddSystemdUnitDropin`, `AddFile` and `getIgnitionValidateValue`

tormath1: ok like we did for `copyKeys` and `merge` function... will try to see if we can avoid copy paste

kola/tests/ignition/kernel.go
pothos: Or maybe more generic like `quiet` if you want to upstream it

platform/conf/conf.go
pothos: sure, just wanted to make sure that this is known because otherwise I think at least vmware tests will fail

kola/tests/ignition/kernel.go
tormath1: but from what I understood, it's up to the distro maintainer to provide the helper: `ignition-kargs-helper`

so not sure if it's something upstreamable.

pothos: I think Fedora CoreOS has this, so the coreos-assembler repo's kola would be able to run it

platform/conf/conf.go
tormath1: added missing helpers in a following commit.

kola/tests/ignition/kernel.go
```diff
@@ -0,0 +1,29 @@
...
+		  },
+		  "kernelArguments": {
+		    "shouldExist": ["quiet"]
+		  }
+		}`),
+		MinVersion: semver.Version{Major: 3178},
+	})
+}
+
+func check(c cluster.TestCluster) {}
```
pothos, line 29:
```suggestion
func check(c cluster.TestCluster) {
	m := c.Machines()[0]

	c.AssertCmdOutputContains(m, "cat /proc/cmdline", " quiet") // assuming space for word separation
}
```

go.mod
```diff
@@ -19,8 +19,8 @@ require (
 	github.com/coreos/yaml v0.0.0-20141224210557-6b16a5714269 // indirect
 	github.com/digitalocean/godo v1.45.0
 	github.com/flatcar-linux/container-linux-config-transpiler v0.9.3-0.20220208152502-6e8303479682
-	github.com/flatcar-linux/ignition v0.36.1
-	github.com/flatcar-linux/ignition/v2 v2.2.1-0.20220107090316-32908ec8bade
+	github.com/flatcar-linux/ignition v0.36.2-0.20220107144332-c956607ca6e9
+	github.com/flatcar-linux/ignition/v2 v2.2.1-0.20220302151613-dcd27a7791fb
```
pothos, line 23: This is a reference to the temporary branch, or?

tormath1: yes, this is the current reference for https://github.com/flatcar-linux/ignition/pull/23 - should be updated once merged and tagged.

pothos: The last commit on our `main` is now `cb95c51122f5a95cea4ed042677933c588fded18`

kola/tests/ignition/kernel.go
pothos: Can we add this?

tormath1: Oh yes definitely - I totally missed this suggestion. :)

kola/tests/ignition/units.go
```diff
@@ -0,0 +1,56 @@
...
+		  "name": "echo@foo.service"
+		},
+		{
+			"enabled": true,
+			"name": "echo@foo.timer"
+		}]
+    }
+}`),
+		Distros: []string{"cl"},
+	})
```
jepio, line 43: @tormath1 this need a MinVersion doesn't it?

tormath1: Oh yes, version 3.0.0... Totally missed this one.

PR #302, opened 2022-03-07
util/bunzip: use lbunzip2 if available

jepio:
# util/bunzip: use lbunzip2 if available

Use lbunzip2 if the binary is available on the system. RIght now the os/prerelease/azure job takes much longer since we switched to fixed size VHDs.

## How to use

Run an os/prerelease/azure job with MANTLE_JOB set to a build containing this commit.

## Testing done

Ran an os/prerelease/azure job with MANTLE_JOB set to a build containing this commit.

### DIFF
--- a/util/bunzip.go
+++ b/util/bunzip.go
@@ -16,8 +16,10 @@ package util
 
 import (
 	"compress/bzip2"
+	"fmt"
 	"io"
 	"os"
+	"os/exec"
 )
 
 // Bunzip2 does bunzip2 decompression from src to dst.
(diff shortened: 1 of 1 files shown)


util/bunzip.go
```diff
@@ -48,3 +49,42 @@ func Bunzip2File(dst, src string) error {
...
+	if err != nil {
+		return fmt.Errorf("setup stdout pipe: %w", err)
+	}
+
+	out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err != nil {
```
pothos, line 72: What `err` should this refer to?

pothos: Ok, so this is any of the `err` variables following. I think it's a bit fragile if someone touches this a later line to be `val, err := something` because it will shadow. I think either a comment that one should not use shadowing would be good or move the `os.Remove` call before each `return fmt.Errorf`.

pothos (approved):

util/bunzip.go
jepio: Ill add the os.Remove calls

PR #303, opened 2022-03-09
Jenkinsfile: flip default for archive artifacts to "true"

jepio:
# Jenkinsfile: flip default for archive artifacts to "true"

Now that we keep only the 3 latest build aritfacts it is perfectly fine
to enable artifact archiving for all branch/PR builds.

This will mean people no longer need to manually trigger a rebuild to test mantle changes.

## How to use

## Testing done

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ properties([
 
     pipelineTriggers([pollSCM('H/15 * * * *')]),
     parameters([
-        booleanParam(name: 'ARCHIVE_ARTIFACTS', defaultValue: false),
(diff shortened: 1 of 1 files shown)


tormath1 (approved):

jepio:
@tormath1 added some more tweaks to the jenkinsfile

tormath1 (approved):
Changes still look good.
I just have one feedback - why do we have a cron task (`H/15 * * * *`) ? To me we should build artifacts / run CI only when something is pushed on the branch no ?

jepio:
That's a fallback to poll scm should the webhook not be delivered.

PR #304, opened 2022-03-09
platform/api/equinixmetal: wipe disks before installing

jepio:
# platform/api/equinixmetal: wipe disks before installing

On EM disk names are not stable and may change across reboots. This is
generally not a problem, but since we started reusing machines, it is
much more likely that there will be a leftover flatcar installation on
one disk and we are now installing to the second. This messes with some
tests because partitions from both disks may be mixed up when mounting
by label. Fix this by wiping the partition tables from disks on
reinstallation.

This should resolve a flake we've seen with cl.ignition.oem.regular, and others where "expected /dev/sda found /dev/sdb" show up.

## How to use

## Testing done

### DIFF
--- a/platform/api/equinixmetal/api.go
+++ b/platform/api/equinixmetal/api.go
@@ -362,6 +362,7 @@ ExecStart=/usr/bin/curl --retry-delay 1 --retry 120 --retry-connrefused --retry-
 # We don't verify signatures because the iPXE script isn't verified either
 # (and, in fact, is transferred over HTTP)
 
+ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs --all --force ${disk}; done'
 ExecStart=/usr/bin/flatcar-install -s -f image.bin.bz2 %v /userdata
 
 ExecStart=/usr/bin/systemctl --no-block isolate reboot.target


platform/api/equinixmetal/api.go
```diff
@@ -362,6 +362,7 @@ ExecStart=/usr/bin/curl --retry-delay 1 --retry 120 --retry-connrefused --retry-
 # We don't verify signatures because the iPXE script isn't verified either
 # (and, in fact, is transferred over HTTP)
 
+ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do echo wipefs -af ${disk}; done'
```
tormath1, line 365: Maybe, I'm missing something - but:
```suggestion
ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs -af ${disk}; done'
```

tormath1, line 365: For this kind of command, we should take the habit to have long options instead of short ones:
```suggestion
ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do echo wipefs --all --force ${disk}; done'
```

What do you think ? It helps for later reviewing - or if you're not familiar with the command. :)

tormath1: Yes I thought that was the main reason haha.

jepio:
Let's keep the commit around, in case we completely misread the situation.

platform/api/equinixmetal/api.go
```diff
@@ -362,6 +362,7 @@ ExecStart=/usr/bin/curl --retry-delay 1 --retry 120 --retry-connrefused --retry-
 # We don't verify signatures because the iPXE script isn't verified either
 # (and, in fact, is transferred over HTTP)
 
+ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs --all --force ${disk}; done'
```
pothos, line 365: In case LVM volumes are left over, it also makes sense to run `lvchange -an /dev/mapper/*` before `flatcar-install`.

pothos: Also, it must be `$${disk}` or it will be substituted as argument for bash itself instead of being evaluated inside the shell

pothos: I'll file a PR

pothos: https://github.com/flatcar-linux/mantle/pull/305

PR #305, opened 2022-03-09
platform/api/equinixmetal: fix disk wipe

pothos:
- platform/api/equinixmetal: fix disk wipe
    
    The "disk" variable got substituted as argument for bash from the unit
    instead of inside bash as shell variable.
    Use escaping to prevent env var substitution inside the unit.
- platform/api/equinixmetal: disable LVM volumes before wiping the disks
    
    The autodetection of LVM volumes can happen now that we recycle
    baremetal instances through a PXE install. For the wipe to work and the
    flatcar-install free disk detection we need to disable any LVM volume
    first.
- platform/api/equinixmetal: delete any /dev/XdX devices
    
    We don't want to only wipe the partitions but the full device.
    Also, we have to look for vda, sdb, and xvda.

## Testing done

none except that the quoting is now right

### DIFF
--- a/platform/api/equinixmetal/api.go
+++ b/platform/api/equinixmetal/api.go
@@ -363,7 +363,7 @@ ExecStart=/usr/bin/curl --retry-delay 1 --retry 120 --retry-connrefused --retry-
 # (and, in fact, is transferred over HTTP)
 
 ExecStartPre=-/bin/bash -c 'lvchange -an /dev/mapper/*'
-ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs --all --force $${disk}; done'
+ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*d? /dev/nvme?n1; do wipefs --all --force $${disk}; done'
 ExecStart=/usr/bin/flatcar-install -s -f image.bin.bz2 %v /userdata
 
 ExecStart=/usr/bin/systemctl --no-block isolate reboot.target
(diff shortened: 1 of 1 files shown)


platform/api/equinixmetal/api.go
```diff
@@ -362,7 +362,8 @@ ExecStart=/usr/bin/curl --retry-delay 1 --retry 120 --retry-connrefused --retry-
 # We don't verify signatures because the iPXE script isn't verified either
 # (and, in fact, is transferred over HTTP)
 
-ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs --all --force ${disk}; done'
+ExecStartPre=-/bin/bash -c 'lvchange -an /dev/mapper/*'
+ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*da? /dev/nvme?n1; do wipefs --all --force $${disk}; done'
```
jepio, line 366:
```suggestion
ExecStartPre=-/bin/bash -c 'shopt -s nullglob; for disk in /dev/*d? /dev/nvme?n1; do wipefs --all --force $${disk}; done'
```

jepio: This made me realize the previous approach was slightly flawed

pothos: Related, do you know when there would be a `n2` suffix for the nvme device?

jepio: Those are nvme namespaces, but I think they are only supported by advanced drives, which we won't come across.

pothos: Should we use `?d?`?

pothos: In case this runs in other other environment there could be `/dev/xda`, or?

jepio (approved):

platform/api/equinixmetal/api.go
pothos: Changed it back to `*d` as prefix to capture `xvda`, too

PR #307, opened 2022-03-17
Add private network support to qemu-unpriv platform

pothos:
The "qemu" platform has the ability for VMs to connect to each other
    but the "qemu-unpriv" platform lacked it, and thus would skip many
    tests.
    Make the qemu-unpriv platform a viable mode for running kola tests
    by adding a private network. This private network only works through
    static or link-local IPv4 addresses (IPv6 would work though) and it
    seems that static addresses make things easier, thus they were chosen.
    Yet since the required setup can't be done in the initramfs, those
    tests that require initramfs private network access have to be skipped.
    The tests which now can run got marked so and those that didn't work
    before either but were forgotten to be excluded are now marked to be
    excluded. Somehow Docker containers can't yet reach the other VMs and
    the kubeadm tests have network config issues.

## How to use

```
kola run -d -k --parallel=4 --platform=qemu-unpriv --qemu-image flatcar_production_image.bin
```

While at it I corrected one esx exclusion which wasn't valid

## Testing done

All tests passed.

Ran `cl.etcd-member.v2-backup-restore` on esx with success

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -15,8 +15,10 @@
 package unprivqemu
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -39,9 +41,12 @@ import (
 // through type assertions.
 type Cluster struct {
 	*platform.BaseCluster
-	flight *flight
+	flight          *flight
+	mcastPortHolder net.Listener
 
 	mu sync.Mutex
+
+	counter uint64
 }
 
 func (qc *Cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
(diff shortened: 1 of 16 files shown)


platform/machine/unprivqemu/cluster.go
```diff
@@ -101,7 +139,9 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 
 	qc.mu.Lock()
 
-	qmCmd = append(qmCmd, "-netdev", "user,id=eth0,hostfwd=tcp:127.0.0.1:0-:22", "-device", platform.Virtio(qc.flight.opts.Board, "net", "netdev=eth0"))
+	sharedNetDev := "socket,id=shared0,mcast=230.0.0.1:" + strings.ReplaceAll(qc.mcastPortHolder.Addr().String(), "127.0.0.99:", "")
```
tormath1, line 142: We could split a bit this line to get a better reading:
```suggestion
        _, port, err := net.SplitHostPort(listener.Addr().String())
	if err != nil { ... }
	sharedNetDev := "socket,id=shared0,mcast=230.0.0.1:" + port
```

platform/machine/unprivqemu/flight.go
```diff
@@ -33,14 +36,15 @@ type flight struct {
 
 	diskImagePath string
 	diskImageFile *os.File
+	nextPort      uint16
```
tormath1, line 39: what's the purpose of this value ?

platform/machine/unprivqemu/flight.go
```diff
@@ -62,9 +66,14 @@ func (qf *flight) NewCluster(rconf *platform.RuntimeConfig) (platform.Cluster, e
 		return nil, err
 	}
 
+	l, err := net.Listen("tcp", "127.0.0.99:0")
```
tormath1, line 69: should we close this listener ? In the flight's `Destroy` maybe ?

platform/machine/unprivqemu/cluster.go
pothos: I've simplified it a bit now

platform/machine/unprivqemu/flight.go
pothos: Thanks, it was a leftover

platform/machine/unprivqemu/flight.go
pothos: Right, I thought about doing it like twice and forgot it again every time^^

platform/machine/unprivqemu/cluster.go
```diff
@@ -60,15 +65,47 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
...
+Description=QEMU metadata agent
+After=nss-lookup.target
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+Environment=OUTPUT=/run/metadata/flatcar
+ExecStart=/usr/bin/mkdir --parent /run/metadata
+ExecStart=/usr/bin/bash -c 'echo "COREOS_CUSTOM_PRIVATE_IPV4=`+privateAddr+`\nCOREOS_CUSTOM_PUBLIC_IPV4=`+privateAddr+`\n" > ${OUTPUT}'
```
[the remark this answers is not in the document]
tormath1: Yes, folder still needs to be created previously - but it avoid using redirection symbols.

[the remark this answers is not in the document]
pothos: The mkdir command can't run as pre-start command because it also wants to write its stdout to the nonexisting folder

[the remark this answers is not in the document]
tormath1: Ok I see - let's keep it in this way so. :)

tormath1 (approved):
We could add a changelog entry in the "changed" or "added" section :) To me it's a high level change

PR #308, opened 2022-03-18
platform/azure: switch to PremiumSSD for OS disk

jepio:
# platform/azure: switch to PremiumSSD for OS disk

bpf.execsnoop times out frequently on Azure, it also takes longer than on other
platforms. Looking deeper suggests that slow disk performance may be the cause.
Switch to premium SSD for disk storage to improve disk performance and reduce
test flakiness.

## How to use

## Testing done


jepio:
Tested here: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/azure/1884/

tormath1 (approved):

sayanchowdhury (approved):

PR #309, opened 2022-03-23
kola/tests/misc/grub.go: Try to fix grubnop test

krnowak:
The mask clause in coreos-cloudinit was masking the unit and then
telling systemd to reload. With that, systemd noticed that update
engine is masked and killed it, but for some reason marked it as
failed.

Current approach is to have a unit that conflicts with update engine
and masks it. That way, systemd should stop update engine without
marking it as failed and then mask it.

CI passed: http://localhost:9091/job/os/job/kola/job/qemu/3993/

### DIFF
--- a/kola/tests/misc/grub.go
+++ b/kola/tests/misc/grub.go
@@ -140,8 +140,21 @@ var (
 
 coreos:
   units:
-    - name: update-engine.service
-      mask: true
+    - name: stop-and-mask-update-engine.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Stop Update Engine
+        Conflicts=update-engine.service
+        Before=update-engine.service
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/usr/bin/systemctl mask --now update-engine.service
+
+        [Install]
+        WantedBy=multi-user.target
 `)
 )
 


kola/tests/misc/grub.go
```diff
@@ -140,8 +140,22 @@ var (
...
+        [Unit]
+        Description=Stop Update Engine
+        Conflicts=update-engine.service
+        Before=update-engine.service
+
+        [Service]
+        # This is probably redundant, since Conflict clause above
+        # should stop update engine, but eh, better safe than sorry.
+        ExecStartPre=-/usr/bin/systemctl stop update-engine.service
+        ExecStart=/usr/bin/systemctl mask update-engine.service
```
pothos, line 155:
```suggestion
        ExecStart=/usr/bin/systemctl mask --now update-engine.service
```
This also stops the service if running

kola/tests/misc/grub.go
```diff
@@ -140,8 +140,22 @@ var (
...
-      mask: true
+    - name: stop-and-mask-update-engine.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Stop Update Engine
+        Conflicts=update-engine.service
+        Before=update-engine.service
+
+        [Service]
```
pothos, line 151:
```suggestion
        [Service]
        Type=oneshot
        RemainAfterExit=true
```
Not really required but in case this gets copied elsewhere I think it's better to have this set to avoid that the service gets started multiple times

pothos (approved):
Thanks, one less

PR #310, opened 2022-03-23
kola/tests/misc: ignore /var/lib/flatcar-oem-gce in some tests

jepio:
# kola/tests/misc: ignore /var/lib/flatcar-oem-gce in some tests

The test execution races against the OEM GCE services, which unpack a container
at /var/lib/flatcar-oem-gce. This folder contains sticky and suid/sgid files,
which cause the tests to fail a lot of the time on GCE. Add
/var/lib/flatcar-oem-gce to the ignore list for relevant tests.

## How to use

## Testing done

Going to test this in CI.


pothos (approved):

pothos:
Thanks :)

PR #314, opened 2022-03-29
update tests

jepio:
# update tests
* bpf.execsnoop is still failing on Azure most of the time - double the launch timeout from ~10s to ~20s to see if that improves the situation
* DO image creation often takes around ~10m, even when switched to gzip compression, which is too tight with the selected timeout of 10m, change to 15m.
* cgroupv1 test can now run on flatcar stable 3033.

## How to use

## Testing done

### DIFF
--- a/kola/tests/bpf/bpf.go
+++ b/kola/tests/bpf/bpf.go
@@ -59,7 +59,7 @@ func execsnoopTest(c cluster.TestCluster) {
 
 		// wait for the container and the `execsnoop` command to be correctly started before
 		// generating traffic.
-		if err := util.Retry(5, 2*time.Second, func() error {
+		if err := util.Retry(10, 2*time.Second, func() error {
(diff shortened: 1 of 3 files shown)


kola/tests/bpf/bpf.go
```diff
@@ -59,7 +59,7 @@ func execsnoopTest(c cluster.TestCluster) {
 
 		// wait for the container and the `execsnoop` command to be correctly started before
 		// generating traffic.
-		if err := util.Retry(5, 2*time.Second, func() error {
+		if err := util.Retry(10, 2*time.Second, func() error {
```
tormath1, line 62: no clue about what's going on ? would be interesting to see a `systemd-analyze blame` but since every other tests are looking fine I guess it's really it's the pulling of the image ? 
We could try to pull a docker or a GHCR image to see if it changes something.

tormath1 (approved):

PR #318, opened 2022-04-11
platform/conf: add `Butane` config support

tormath1:
In this PR, we provide `conf.Butane` to define userdata:
```go
		UserData: conf.Butane(`---
variant: flatcar
version: 1.0.0
kernel_arguments:
  should_exist:
    - quiet`),
```

it should be used whenever we want to generate Ignition configuration, in place of CLC - except if we want to test specific CLC  of course. It should support other `version` of Butane Flatcar's variant as long as we keep the vendor in sync with the upstream dependency since it's butane who takes care of finding the right translator for a given configuration.

`1.1.0-experimental` does not work currently because it generates Ignition-3.4.0 which is not supported by Ignition neither Flatcar.

## Note for reviewers

* The following commit: ` platform/conf: use coreos/ignition for ignition/v3` is required because CLC generates an Ignition `types` defined in `flatcar-linux/ignition`, so we need to keep `flatcar-linux/ignition`. 

- [x] Tested with updated tests: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu_uefi/145/


tormath1:
CI passed.

pothos (approved):

PR #320, opened 2022-04-12
kola: skip Ignition kargs test on Equinix Metal

pothos:
The Ignition kargs test hits mantle's provisioning timeout due to the
additional reboot.
Since we don't really need this test to run on Equinix Metal since
running it on QEMU is equally good, we can skip it.

### DIFF
--- a/kola/tests/ignition/kernel.go
+++ b/kola/tests/ignition/kernel.go
@@ -23,6 +23,9 @@ func init() {
 		  }
 		}`),
 		MinVersion: semver.Version{Major: 3185},
+		// The additional reboot causes a large waiting time
+		// and it's enough to test this on QEMU
+		ExcludePlatforms: []string{"equinixmetal"},
 	})
 }
 


dongsupark (approved):
Funny, @tormath1 and I talked about this issue minutes ago.

kola/tests/ignition/kernel.go
```diff
@@ -23,6 +23,9 @@ func init() {
 		  }
 		}`),
 		MinVersion: semver.Version{Major: 3185},
+		// The additional reboot causes a large waiting time
+		// and it's enough to test this on QEMU
+		ExcludePlatforms: []string{"equinixmetal"},
```
tormath1, line 28:
```suggestion
		Platforms: []string{"qemu"},
```

if we consider QEMU platform as enough ?

pothos: Should we do this for more tests now or is this for the restructuring later?

pothos: I didn't want to forbid running on other platforms in case it makes sense there. How about a flag that a test is "generic" and when running the QEMU tests in the CI we can enable "generic" tests, but when we run other cloud tests we can disable "generic" tests? This would allow the generic tests to run on a any platform if wanted.

tormath1: Ok, in that case let's keep `ExcludePlatforms: []string{"equinixmetal"},` to not block the PR.

Tell me more about this `generic` flag. What would be the difference between a test flagged as "generic" and `Platforms: []string{"all"}` ?

pothos: *looking into the future* A new generic flag would allow to filter which tests to run: we can use `kola run --no-generic` when doing cloud platform testing to skip those tests which are not interesting because they do the same as when executed under QEMU. The extra flag would still allow these tests to run for a full validation or when reproducing an issue, without having to change the kola source code.

pothos: Should I create an issue for this?

tormath1: Yes - that sounds like a good feature. We would just need to define based on which criteria a test can be defined as "generic" or not.

pothos: Added it to the existing issue: https://github.com/flatcar-linux/Flatcar/issues/654#issuecomment-1097501973

PR #321, opened 2022-04-12
azure: set user-data on top of custom data

tormath1:
`userData` should be the new way to set user-data.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
@@ -367,9 +368,23 @@ func RunTests(patterns []string, channel, offering, pltfrm, outputDir string, ss
 
 	if !skipGetVersion {
 		plog.Info("Creating cluster to check semver...")
+
 		version, err := getClusterSemver(flight, outputDir)
 		if err != nil {
 			plog.Fatal(err)
+		}
+
+		// If the version is > 3033, we can safely use user-data instead of custom-data for
+		// provisioning the instance on Azure.
+		if !version.LessThan(semver.Version{Major: 3034}) && pltfrm == "azure" {
+			// Using reflection is a bit hacky, but it seems to be the only way to
+			// access the field we want to set.
+			f := reflect.ValueOf(flight).Elem()
+			api := f.FieldByName("Api")
+			opts := api.Elem().FieldByName("Opts")
+			userData := opts.Elem().FieldByName("UseUserData")
+			// At this point, this field can be set.
+			userData.SetBool(true)
 		}
 
 		versionStr = version.String()
--- a/platform/api/azure/api.go
+++ b/platform/api/azure/api.go
@@ -50,7 +50,7 @@ type API struct {
 	ipClient   network.PublicIPAddressesClient
 	intClient  network.InterfacesClient
 	accClient  armStorage.AccountsClient
-	opts       *Options
+	Opts       *Options
 }
 
 type Network struct {
@@ -125,7 +125,7 @@ func New(opts *Options) (*API, error) {
 
 	api := &API{
 		client: client,
-		opts:   opts,
+		Opts:   opts,
 	}
 
 	err = api.resolveImage()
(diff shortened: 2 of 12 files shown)


jepio:
If we set both, then we won't be properly testing what happens when only one of two is set.

tormath1:
> If we set both, then we won't be properly testing what happens when only one of two is set.

Yeah - I was thinking checking the ignition logs to see if we have the "fail to get user-data - fallback to custom-data" line.

pothos:
We don't have a way to set this based on the Flatcar major version but adding this would help with both the transition and proper testing

tormath1:
we could introduce an `--azure-imds` flag to conditionally supply `userData` to `customData` or `userData` instance parameter (in the same way as @jepio did for GCE GVNIC)

pothos:
That would serve as a "unit test" for Ignition but I think it's not really something we should test for each release. In addition, it does not solve the transition problem where we need to decide based on the major version whether userData is supported or not.

pothos:
We could add a parameter to the VM creation to use either CustomData or UserData. The flight could carry the setting based on the result of the semver machine boot's version gathering where the Alpha version gets hardcoded.

kola/harness.go
```diff
@@ -371,6 +371,21 @@ func RunTests(patterns []string, channel, offering, pltfrm, outputDir string, ss
 			plog.Fatal(err)
 		}
 
+		if version.LessThan(semver.Version{Major: 3034}) && pltfrm == "azure" {
+			AzureOptions.UseCustomData = true
+
+			// We need to recreate the flight to apply this modification.
+			flight, err = NewFlight(pltfrm)
```
pothos, line 378: Creating a new flight is not so good, it creates a new resource group and would also upload the image again. Since we know that `flight` is an Azure flight, can we access `flight.api.GetOpts()` and set it there? Maybe we need to add a getter to the flight or do some reflection?

tormath1: So I've been for reflection - it's far from being perfect and I had to export the `Api` and the `Opts` fields (otherwise the field that we want to update can't be set).
`api` can't be accessed as it's not part of the initial `platform.Flight` struct (and some platforms do not have this field neither). The whole thing would need a bit of abstraction.

kola/harness.go
```diff
@@ -367,11 +368,26 @@ func RunTests(patterns []string, channel, offering, pltfrm, outputDir string, ss
...
+		if version.Compare(semver.Version{Major: 3034}) == 1 && pltfrm == "azure" {
+			// Using reflection is a bit hacky, but it seems to be the only way to
+			// access the field we want to set.
+			f := reflect.ValueOf(flight).Elem()
+			api := f.FieldByName("Api")
+			opts := api.Elem().FieldByName("Opts")
+			userData := opts.Elem().FieldByName("UseUserData")
+			if userData.CanSet() {
+				userData.SetBool(true)
+			}
```
pothos, line 388:
```suggestion
			} else { some error }
```
When would `CanSet` be false?

kola/harness.go
```diff
@@ -367,11 +368,26 @@ func RunTests(patterns []string, channel, offering, pltfrm, outputDir string, ss
...
 		plog.Info("Creating cluster to check semver...")
+
 		version, err := getClusterSemver(flight, outputDir)
 		if err != nil {
 			plog.Fatal(err)
 		}
 
+		// If the version is > 3033, we can safely use user-data instead of custom-data for
+		// provisioning the instance on Azure.
+		if version.Compare(semver.Version{Major: 3034}) == 1 && pltfrm == "azure" {
```
pothos, line 379:
```suggestion
		if version.GreaterThanOrEqual(semver.Version{Major: 3034}) && pltfrm == "azure" {
```

tormath1: Where did you find it? There is no mention of this in the godoc (https://pkg.go.dev/github.com/coreos/go-semver/semver) neither in the vendors 🤔 
I only found `LessThan`, that's why I used `Compare`

pothos: Ah, then it's the wrong go version library I looked up^^ Then how about `!LessThan`?

kola/harness.go
tormath1: In theory never as the fields are known and exported. Since there is not even an error handling, let's drop the if condition.

PR #322, opened 2022-04-13
gcp image features (UEFI/GVNIC)

jepio:
# gcp image features

This PR marks published GCP images as supporting UEFI and GVNIC.

Marking the images as UEFI automatically makes instances boot with UEFI and have access to a VTPM. 

Marking the image as support GVNIC does not automatically make it active but allows users to choose GVNIC. Therefore I intend to backport https://github.com/flatcar-linux/coreos-overlay/pull/1802 to all releases. For kola testing a new flat is added that enables GVNIC on tested instances.

## How to use

## Testing done

Tested here: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5373/cldsv/, fetch the latest os/kola/gce console logs and look for gve/gvnic/EFI.


jepio:
Base is chosen intentionally - when tormath1/bump-go is merged this PR will be re-targeted at flatcar-master.

platform/api/gcloud/compute.go
```diff
@@ -52,6 +52,10 @@ func (a *API) mkinstance(userdata, name string, keys []*agent.Key) *compute.Inst
 			Value: &sshKeys,
 		})
 	}
+	nicType := ""
```
tormath1, line 55:
```suggestion
	nicType := "VIRTIO_NET"
```

What do you think ? 

Doc: https://pkg.go.dev/google.golang.org/api/compute/v1#Instance
```
        // NicType: The type of vNIC to be used on this interface. This may be
	// gVNIC or VirtioNet.
	//
	// Possible values:
	//   "GVNIC" - GVNIC
	//   "UNSPECIFIED_NIC_TYPE" - No type specified.
	//   "VIRTIO_NET" - VIRTIO
	NicType string `json:"nicType,omitempty"`
```

tormath1 (approved):
Thanks ! This is cool. If I understand correctly, we would also need to update GCP testing script to supply `--gce-gvnic` for supported versions ?

jepio:
Yes, we'll probably add a 'cl.internet' test with this option to the GCP testing.

platform/api/gcloud/compute.go
jepio: Yeah makes sense

jepio:
Will merge this after the GCP vnic support is cherry-picked everywhere.

jepio:
The coreos-overlay PR has been cherry-picked to everything back to 3033, so it is now safe to merge this.

PR #323, opened 2022-04-13
.github: Publish mantle container image

pothos:
A container image addresses the needs of running arm64 kola tests
    without QEMU emulation but with KVM on arm64 hardware because there is
    no native Flatcar SDK for arm64 yet, and it also allows us to have a
    single place that provides a particular version of the mantle binaries
    that is more lightweight than the full SDK where rebuilding the ebuild
    package is currently needed.
    The container image is published for the main branch but uses the git
    SHA commit ID as image tag which should be updated in the "scripts"
    repository.
    There is no garbage collection yet to remove images created by PR
    events.

## How to use

To be used from the scripts repo in the ci-automation pipeline

## Testing done

For this PR an image `mantle:pr-323` was built. Checked that it gets updated on force push.
In https://github.com/pothos/mantle a `mantle:git-COMMITID` image was built for flatcar-master.

Ran `kola run -d --platform=qemu-unpriv --qemu-image /w/flatcar_production_image.bin cl.basic` inside the container for amd64 and `kola run -d --platform=qemu --board arm64-usr --qemu-bios=/flatcar_production_qemu_uefi_efi_code.fd --qemu-image /flatcar_production_image.bin cl.basic` on an arm64 host

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+# golang:1.17 is based on debian:11, this is important to ensure we have libc compatibility for the copied binary
+
+FROM docker.io/library/golang:1.17 as builder
+ENV CGO_ENABLED=1
+COPY . /usr/src/mantle
+RUN cd /usr/src/mantle && ./build
+
+FROM docker.io/library/debian:11
+RUN apt-get update && apt-get upgrade -y && apt-get install --no-install-recommends -y qemu-utils qemu-system-x86 qemu-system-aarch64 qemu-efi-aarch64 seabios ovmf lbzip2 sudo dnsmasq gnupg2 git curl iptables nftables dns-root-data ca-certificates sqlite3
+COPY --from=builder /usr/src/mantle/bin /usr/local/bin
+RUN printf '#!/bin/sh\n# Workaround for kola to find its kolet binaries, this script is stored in sbin to get precedence\nexec /usr/local/bin/kola "$@"\n' > /usr/local/sbin/kola
+RUN chmod +x /usr/local/sbin/kola
+RUN ln -s /usr/share/seabios/bios-256k.bin /usr/share/qemu/bios-256k.bin
+
+# For KVM to work, run the resulting container as: docker run --privileged --net host -v /dev:/dev --rm -it TAG
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,58 @@
+name: Build docker image
+on:
+  push:
+    branches:
+      - 'flatcar-master'
+  pull_request:
+    branches:
+      - 'flatcar-master'
+env:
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
+# golang:1.17 is based on debian:11, this is important to ensure we have libc compatibility for the copied binary
+
+FROM docker.io/library/golang:1.17 as builder
+ENV CGO_ENABLED=1
```
jepio, line 4: I don't know how, but the size of the mantle binaries built with `CGO_ENABLED=1` or `CGO_ENABLED=0` is essentially the same, so we could also get rid of this restriction.

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+COPY . /usr/src/mantle
+RUN cd /usr/src/mantle && ./build
+
+FROM docker.io/library/debian:11
+RUN apt-get update && apt-get upgrade -y && apt-get install --no-install-recommends -y qemu-utils qemu-system-x86 qemu-system-aarch64 qemu-efi-aarch64 seabios ovmf lbzip2 sudo dnsmasq gnupg2 git curl iptables nftables dns-root-data ca-certificates sqlite3
+COPY --from=builder /usr/src/mantle/bin /usr/local/bin
+RUN printf '#!/bin/sh\n# Workaround for kola to find its kola binary, stored in sbin to get precedence\nexec /usr/local/bin/kola "$@"\n' > /usr/local/sbin/kola
```
jepio, line 11: Can you explain what is happening here?

Dockerfile
pothos: I'm happy to try out disabling cgo but in a follow up PR where we make sure that all works compared to the current state

Dockerfile
pothos: Oh, it should be `to find its kolet binaries`. Didn't check how kola's equivalent to finding the path from `$0` is implemented but it somehow can't cope with being called by `kola` in another directory, only with the full path.

pothos: Updated the explaination.

From a quick look kola uses `filepath.Dir(os.Args[0])` or `/usr/lib/kola`, so it either need to be called with a full or relative path or we move the kolet binaries to `/usr/lib/kola`. I think it would be good to improve the logic in kola directly as this came up sometimes already.

jepio: Ok, now I understand.

jepio (approved):

Dockerfile
pothos: There is https://pkg.go.dev/os#Executable which we can use like this:

```
	p, err := os.Executable()
	if err != nil {
		if (strings.Contains(os.Args[0], "/")) {
			p =  os.Args[0]
		} else {
			p, err = exec.LookPath(os.Args[0])
			if err != nil {
			p =  os.Args[0]
			}
		}
	}
	fmt.Println(filepath.Dir(p))
```

pothos: Will create a follow-up PR

PR #325, opened 2022-04-20
kola/tests/misc: add testcase for oem-gce service running on GCP

jepio:
Add testcase for oem-gce.service on GCP. This test case is currently failing but will pass once https://github.com/flatcar-linux/coreos-overlay/pull/1813 is merged.

### DIFF
--- a/kola/tests/misc/gcp.go
+++ b/kola/tests/misc/gcp.go
@@ -1,0 +1,30 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package misc
+
+import (
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "cl.misc.gce.oem",
+		ClusterSize: 1,
+		Platforms:   []string{"gce"},
+		Distros:     []string{"cl"},
+		Run:         gceVerifyOEMService,
+	})
+}
+
+func gceVerifyOEMService(c cluster.TestCluster) {
+	machine := c.Machines()[0]
+	// verify that the oem-gce service is running
+	c.MustSSH(machine, "systemctl is-active oem-gce.service")
+	nrestarts := c.MustSSH(machine, "systemctl show oem-gce.service -P NRestarts")
+	if string(nrestarts) != "0" {
+		c.Fatalf("oem-gce service restarted too many times: %s", nrestarts)
+	}
+	// check that interface is configured and named correctly
+	c.MustSSH(machine, "networkctl status eth0")
+}


kola/tests/misc/gcp.go
```diff
@@ -0,0 +1,27 @@
...
+	})
+}
+
+func gceVerifyOEMService(c cluster.TestCluster) {
+	// verify that the oem-gce service is running
+	c.MustSSH(c.Machines()[0], "systemctl is-active oem-gce.service")
+	nrestarts := c.MustSSH(c.Machines()[0], "systemctl show oem-gce.service -P NRestarts")
+	if string(nrestarts) != "0" {
+		c.Fatalf("oem-gce service restarted too many times: %s", nrestarts)
+	}
```
pothos, line 26: One more thing we can check is the interface name

```suggestion
	}
	c.MustSSH(c.Machines()[0], "networkctl status eth0")
```

kola/tests/misc/gcp.go
```diff
@@ -0,0 +1,27 @@
...
+		Distros:     []string{"cl"},
+		Run:         gceVerifyOEMService,
+	})
+}
+
+func gceVerifyOEMService(c cluster.TestCluster) {
+	// verify that the oem-gce service is running
+	c.MustSSH(c.Machines()[0], "systemctl is-active oem-gce.service")
+	nrestarts := c.MustSSH(c.Machines()[0], "systemctl show oem-gce.service -P NRestarts")
+	if string(nrestarts) != "0" {
```
pothos, line 24: With the sysext workaround I had created last week it could lead to a restart, but I think we can add `After=…` to the oem-gce service to prevent a restart.

Also, maybe first do `systemctl is-system-running --wait` to prevent races?

jepio: `platform.CheckMachine ` is run in `platform.StartMachine` and loops on `systemctl is-system-running`, so this is already covered. The oem-service doesn't show up as failed because it's a "restart always" service, I think that's why we didn't catch this in the first place.

I think the oem service should not restart because `systemd-sysext` runs `Before=sysinit.target`.

kola/tests/misc/gcp.go
jepio: adding this

jepio:
Failing on main (correct): http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/gce/43/
Passes with oem-gce fix branch: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/gce/45/

kola/tests/misc/gcp.go
pothos: Good, I'll try to test again with this test case and make sure it won't have to restart :)

pothos (approved):

PR #329, opened 2022-04-28
esx: support initramfs network configuration

tormath1:
In this PR, we add the `ip` kernel argument following this pattern[^1]:
```
ip=<client-IP>:[<peer>]:<gateway-IP>:<netmask>:<client_hostname>:<interface>:{none|off|dhcp|on|any|dhcp6|auto6|ibft}[:[<dns1>][:<dns2>]] 
```
and we remove `esx` from excluded platforms for network reasons.

<hr>

Related to: https://github.com/flatcar-linux/Flatcar/issues/717

I'm not sure it takes a changelog entry

EDIT: don't merge this before getting the feature in stable - otherwise we'll prevent the impacted tests to run

[^1]: https://mirrors.edge.kernel.org/pub/linux/utils/boot/dracut/dracut.html#_network

### DIFF
--- a/kola/tests/bpf/local-gadget.go
+++ b/kola/tests/bpf/local-gadget.go
@@ -98,9 +98,6 @@ func init() {
 		Run:     localGadgetTest,
 		Name:    `bpf.local-gadget`,
 		Distros: []string{"cl"},
-		// ESX is excluded because initramfs has no network access
-		// so it's not able to download local-gadget binary.
-		ExcludePlatforms: []string{"esx"},
 		// required while SELinux policy is not correcly updated to support
 		// `bpf` and `perfmon` permission.
 		Flags: []register.Flag{register.NoEnableSelinux},
@@ -109,6 +106,12 @@ func init() {
 		MinVersion: semver.Version{Major: 3033},
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS (3033) does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for < 3034 on ESX.
+			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
+		},
 	})
 }
 
--- a/kola/tests/ignition/resource.go
+++ b/kola/tests/ignition/resource.go
@@ -105,19 +105,29 @@ func init() {
 			"Serve": Serve,
 		},
 		// https://github.com/coreos/bugs/issues/2205
-		// ESX: Currently Ignition does not support static IPs during the initramfs
-		ExcludePlatforms: []string{"esx", "do", "qemu-unpriv"},
+		ExcludePlatforms: []string{"do", "qemu-unpriv"},
 		Distros:          []string{"cl", "fcos", "rhcos"},
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS (3033) does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for < 3034 on ESX.
+			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
+		},
 		// This should run on all clouds to test initramfs networking
 	})
 	register.Register(&register.Test{
 		Name:        "coreos.ignition.resource.remote",
 		Run:         resourceRemote,
 		ClusterSize: 1,
-		// ESX: Currently Ignition does not support static IPs during the initramfs
 		// https://github.com/coreos/bugs/issues/2205 for DO
-		ExcludePlatforms: []string{"esx", "do"},
 		// This should run on all clouds to test initramfs networking
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS (3033) does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for < 3034 on ESX.
+			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
+		},
+		ExcludePlatforms: []string{"do"},
 		UserData: conf.Ignition(`{
 		  "ignition": {
 		      "version": "2.1.0"
@@ -237,12 +247,16 @@ func init() {
 	//       this test should be rolled into coreos.ignition.resources.remote
 	// Test specifically for versioned s3 objects
 	register.Register(&register.Test{
-		Name:        "coreos.ignition.resource.s3.versioned",
-		Run:         resourceS3Versioned,
-		ClusterSize: 1,
-		// ESX: Currently Ignition does not support static IPs during the initramfs
-		// https://github.com/coreos/bugs/issues/2205 for DO
-		ExcludePlatforms: []string{"esx", "do"},
+		Name:             "coreos.ignition.resource.s3.versioned",
+		Run:              resourceS3Versioned,
+		ClusterSize:      1,
+		ExcludePlatforms: []string{"do"},
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS (3033) does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for < 3034 on ESX.
+			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
+		},
 		// This does not need to run on all clouds because it is a special case of coreos.ignition.resource.remote
 		Platforms:  []string{"qemu", "qemu-unpriv"},
 		MinVersion: semver.Version{Major: 1995},
--- a/kola/tests/kubeadm/kubeadm.go
+++ b/kola/tests/kubeadm/kubeadm.go
@@ -210,14 +210,20 @@ func init() {
 				register.Register(&register.Test{
 					Name:    fmt.Sprintf("kubeadm.%s.%s%s.base", version, CNI, cgroupSuffix),
 					Distros: []string{"cl"},
-					// Network config problems in esx and qemu-unpriv
-					ExcludePlatforms: []string{"esx", "qemu-unpriv"},
 					// This should run on all clouds as a good end-to-end test
+					// Network config problems in qemu-unpriv
+					ExcludePlatforms: []string{"qemu-unpriv"},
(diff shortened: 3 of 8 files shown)


platform/api/esx/api.go
```diff
@@ -429,6 +429,12 @@ func (a *API) CreateDevice(name string, conf *conf.Conf, ips *IpPair) (*ESXMachi
 	}
 
 	if ips != nil {
+		// based on: https://mirrors.edge.kernel.org/pub/linux/utils/boot/dracut/dracut.html#_network
+		kArgs := fmt.Sprintf("ip=%s::%s:%d:localhost:ens192:off:1.1.1.1:1.0.0.1", ips.Public, ips.PublicGw, ips.SubnetSize)
```
pothos, line 433: I think instead of SubnetSize the netmask is expected: https://www.flatcar.org/docs/latest/provisioning/ignition/network-configuration/#using-static-ip-addresses-with-ignition
I would also leave the hostname empty instead of using `localhost`.
Is there a difference between `off` and `none` (our docs only say `none`)?

tormath1: I gave a try with both. Also seen in this post: https://www.redhat.com/sysadmin/network-confi-initrd  - but can't find an actual documentation mentioning the expected format. I think kernel can deal with both CIDR notation and netmask ? Initially I was calculating the netmask: https://github.com/flatcar-linux/mantle/compare/1d654413912be9aba3c24a04303b445b76cfdc4e..e175cbb3d76f19f5557f119140dc8e4e9964f619

About `none/off` they seem to have the same effect (static IP) (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-configuring_ip_networking_from_the_kernel_command_line)

pothos: The only thing that counts is the code in this script which may divert from other implementations: https://github.com/flatcar-linux/bootengine/blob/flatcar-master/dracut/03flatcar-network/parse-ip-for-networkd.sh#L41
and indeed the `elif` case seems to accept an integer (SubnetSize) instead of a netmask. And `none` is also the same as `off` in our case: https://github.com/flatcar-linux/bootengine/blob/flatcar-master/dracut/03flatcar-network/parse-ip-for-networkd.sh#L53

pothos:
```suggestion
		kArgs := fmt.Sprintf("ip=%s::%s:%d::ens192:off:1.1.1.1:1.0.0.1", ips.Public, ips.PublicGw, ips.SubnetSize)
```
So I would just leave out setting the hostname explicitly to localhost.

tormath1:
> Looks good, but I don't get the request to not merge it until a feature hits stable. Which feature? Some dracut version?

Sorry for not being clear :D I'm referring to this feature: https://github.com/flatcar-linux/bootengine/pull/44. Setting a new `guestinfo` variable to an instance is harmless but if we merge right now this PR, we'll get failures on tests modified in this commit: https://github.com/flatcar-linux/mantle/pull/329/commits/be948546d8873115c4a3a96b0ca1e5f159b274e0

krnowak:
> > Looks good, but I don't get the request to not merge it until a feature hits stable. Which feature? Some dracut version?
> 
> Sorry for not being clear :D I'm referring to this feature: [flatcar-linux/bootengine#44](https://github.com/flatcar-linux/bootengine/pull/44). Setting a new `guestinfo` variable to an instance is harmless but if we merge right now this PR, we'll get failures on tests modified in this commit: [be94854](https://github.com/flatcar-linux/mantle/commit/be948546d8873115c4a3a96b0ca1e5f159b274e0)

Cool, thanks for making it clear. :)

So, uh, is this PR going to wait for a month (until current alpha becomes stable) or the plan is to cherry-pick the bootengine changes to stable?

pothos:
We could use the min version flag but that means we stop running tests on all platforms for Stable…

kola/tests/kubeadm/kubeadm.go
```diff
@@ -210,9 +210,9 @@ func init() {
 				register.Register(&register.Test{
 					Name:    fmt.Sprintf("kubeadm.%s.%s%s.base", version, CNI, cgroupSuffix),
 					Distros: []string{"cl"},
-					// Network config problems in esx and qemu-unpriv
-					ExcludePlatforms: []string{"esx", "qemu-unpriv"},
 					// This should run on all clouds as a good end-to-end test
+					// Network config problems in qemu-unpriv
+					ExcludePlatforms: []string{"qemu-unpriv"},
```
pothos, line 215: I thought we had other network problems, great if it was just the initrd

kola/tests/ignition/resource.go
```diff
@@ -240,9 +238,6 @@ func init() {
 		Name:        "coreos.ignition.resource.s3.versioned",
 		Run:         resourceS3Versioned,
 		ClusterSize: 1,
-		// ESX: Currently Ignition does not support static IPs during the initramfs
-		// https://github.com/coreos/bugs/issues/2205 for DO
-		ExcludePlatforms: []string{"esx", "do"},
```
pothos, line 245: The change above still excludes "do", this here not, maybe worth testing if it works now.

kola/tests/kubeadm/kubeadm.go
tormath1: This PR is old but now the feature has landed on stable, I'm preparing the merge. IIRC, this test was running fine. We just need to deal with LTS: I was thinking about adding a `Skip` on all tests impacted with the following condition: if "esx" and "lts" but it's not ideal.

pothos:
We need `MinVersion: semver.Version{Major: abcd},` for the tests to be skipped on LTS.

tormath1:
> We need `MinVersion: semver.Version{Major: abcd},` for the tests to be skipped on LTS.

Seem this answer: https://github.com/flatcar/mantle/pull/329#discussion_r1025062281 I'd like to avoid skipping kubeadm tests on LTS for example :D

kola/tests/kubeadm/kubeadm.go
pothos: Ah, ok, we need our own DSL to evaluate test run conditions :D

pothos: We could add an attribute that is a lambda function: `skipFunc: func(version, platform string)boolean{ … }`

tormath1: Ah yes, nice idea. We could even extend the signature to `func(version, channel, arch, platform string)`.

kola/tests/ignition/resource.go
tormath1: That's normal, this test is flagged as `qemu` only:
> This does not need to run on all clouds because it is a special case of coreos.ignition.resource.remote

So the `ExcludePlatforms` was useless.

pothos: Right, yet in case we want to run all tests on DO it would be good to know which ones are expected to work and which were explicitly excluded.

kola/tests/kubeadm/kubeadm.go
tormath1: With OVA from: https://stable.release.flatcar-linux.net/amd64-usr/3374.2.0/
```
kola run ... --platform=esx --channel=stable -esx-ova-path ./flatcar_production_vmware_ova.ova kubeadm.v1.24.1.calico.base
    --- PASS: kubeadm.v1.24.1.calico.base/node_readiness (38.29s)
    --- PASS: kubeadm.v1.24.1.calico.base/nginx_deployment (16.50s)
PASS, output in _kola_temp/esx-2022-11-17-1542-44142
```

kola/tests/kubeadm/kubeadm.go
```diff
@@ -210,14 +210,20 @@ func init() {
...
 					Run: func(c cluster.TestCluster) {
 						kubeadmBaseTest(c, testParams)
 					},
 					MinVersion: semver.Version{Major: major},
 					Flags:      flags,
+					SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+						// LTS does not have the network-kargs service pulled in:
+						// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+						// Let's skip this test for LTS channel on ESX.
+						return channel == "lts" && platform == "esx"
```
pothos, line 225:
```suggestion
						return strings.HasPrefix(version, "3033") && platform == "esx"
```
Future LTS versions should run this test

kola/tests/kubeadm/kubeadm.go
pothos: Glad to see that we can run kubeadm tests on one more platform!

kola/tests/bpf/local-gadget.go
```diff
@@ -98,9 +98,6 @@ func init() {
 		Run:     localGadgetTest,
 		Name:    `bpf.local-gadget`,
 		Distros: []string{"cl"},
-		// ESX is excluded because initramfs has no network access
-		// so it's not able to download local-gadget binary.
-		ExcludePlatforms: []string{"esx"},
```
pothos, line 103: This needs the SkipFunc, too

pothos: Thanks, that helps if we have to do a full run on a platform

kola/tests/ignition/resource.go
```diff
@@ -105,19 +105,29 @@ func init() {
...
 		ClusterSize: 1,
-		// ESX: Currently Ignition does not support static IPs during the initramfs
 		// https://github.com/coreos/bugs/issues/2205 for DO
-		ExcludePlatforms: []string{"esx", "do"},
 		// This should run on all clouds to test initramfs networking
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for LTS channel on ESX.
+			return channel == "lts" && platform == "esx"
```
pothos, line 128:
```suggestion
			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
```

kola/tests/ignition/resource.go
```diff
@@ -237,12 +247,16 @@ func init() {
...
-		ExcludePlatforms: []string{"esx", "do"},
+		Name:             "coreos.ignition.resource.s3.versioned",
+		Run:              resourceS3Versioned,
+		ClusterSize:      1,
+		ExcludePlatforms: []string{"do"},
+		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// LTS does not have the network-kargs service pulled in:
+			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
+			// Let's skip this test for LTS channel on ESX.
+			return channel == "lts" && platform == "esx"
```
pothos, line 258:
```suggestion
			return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
```

kola/tests/ignition/resource.go
tormath1: :facepalm:

pothos (approved):
Thanks for picking this up again! The solution we found now will be useful for similar cases :)

PR #334, opened 2022-05-20
test/ignition: add `networkd` translation test

tormath1:
In this PR, we add a new test to ensure `networkd` Ignition from spec 2.x is correctly translated to `files` from spec 3.x and applied correctly.

This has been tested with current stable (where `networkd` is still supported) and with https://github.com/flatcar-linux/coreos-overlay/pull/1910

Related to: https://github.com/flatcar-linux/Flatcar/issues/741

:warning: this should be merged _after_ the coreos-overlay's PR merged, backported to alpha/beta and released - otherwise test will fail (except on stable where we still on ignition 2.x so `networkd` is still supported)


dongsupark (approved):
Looks good.

Probably add changelog?

PR #339, opened 2022-06-15
kola: add test that checks that falco module can be built and loaded

jepio:
# add falco module test

This tests uses the master ref of falcosecurity/falco-driver-loader, but can soon be moved to 0.33.0 once that is released. Selinux needs to be disabled because it blocks insmod from within the container.

## How to use

## Testing done
Test succeeded: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu/553/

### DIFF
--- a/kola/tests/misc/falco.go
+++ b/kola/tests/misc/falco.go
@@ -1,0 +1,27 @@
+package misc
+
+import (
+	"github.com/flatcar-linux/mantle/kola/cluster"
+	"github.com/flatcar-linux/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         loadFalco,
+		ClusterSize: 1,
+		Name:        "cl.misc.falco",
+		Distros:     []string{"cl"},
+		Platforms:   []string{"qemu"},
+		// selinux blocks insmod from within container
+		Flags: []register.Flag{register.NoEnableSelinux},
+	})
+}
+
+func loadFalco(c cluster.TestCluster) {
+	// load the falco binary
+	// TODO: first supported version will be 0.33.0, but use master tag for now
+	c.MustSSH(c.Machines()[0], "docker run --rm --privileged -v /root/.falco:/root/.falco -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc:ro falcosecurity/falco-driver-loader:master")
+	// Build must succeed and falco must be running
+	c.MustSSH(c.Machines()[0], "dmesg | grep falco")
+	c.MustSSH(c.Machines()[0], "lsmod | grep falco")
+}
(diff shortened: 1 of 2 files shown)


kola/tests/misc/falco.go
```diff
@@ -0,0 +1,27 @@
...
+		Distros:     []string{"cl"},
+		Platforms:   []string{"qemu"},
+		// selinux blocks insmod from within container
+		Flags: []register.Flag{register.NoEnableSelinux},
+	})
+}
+
+func loadFalco(c cluster.TestCluster) {
+	// load the falco binary
+	// TODO: first supported version will be 0.33.0, but use master tag for now
```
tormath1, line 23: On Flatcar side, does it work with all kernel version ? (5.15 and 5.10 for LTS)

kola/tests/misc/falco.go
```diff
@@ -0,0 +1,27 @@
...
+	c.MustSSH(c.Machines()[0], "docker run --rm --privileged -v /root/.falco:/root/.falco -v /proc:/host/proc:ro -v /boot:/host/boot:ro -v /lib/modules:/host/lib/modules:ro -v /usr:/host/usr:ro -v /etc:/host/etc:ro falcosecurity/falco-driver-loader:master")
+	// Build must succeed and falco must be running
+	c.MustSSH(c.Machines()[0], "dmesg | grep falco")
```
tormath1, line 26: This is maybe a bit too generic ? What about `modinfo falco` or `lsmod | grep falco` ? Just to be sure that the module is _actually_ loaded

tormath1 (commented):
Thanks for this test and for the great work around this topic ! :partying_face:

kola/tests/misc/falco.go
tormath1: Ah `modinfo` does not work since falco is out-of-tree

jepio: `lsmod | grep falco` is a good idea, i'll add that

kola/tests/misc/falco.go
jepio: It's expected to work with all Flatcar versions, going as far back as when we started shipping a kernel module build env in /lib/modules/*/build.

kola/tests/misc/falco.go
```diff
@@ -0,0 +1,27 @@
...
+// create a test case
+func init() {
+	register.Register(&register.Test{
+		Run:         loadFalco,
+		ClusterSize: 1,
+		Name:        "cl.misc.falco",
+		Distros:     []string{"cl"},
+		Platforms:   []string{"qemu"},
+		// selinux blocks insmod from within container
+		Flags: []register.Flag{register.NoEnableSelinux},
```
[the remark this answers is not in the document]
jepio: Created an issue here: https://github.com/flatcar-linux/Flatcar/issues/783

tormath1 (approved):
Thanks! Let's monitor if we hit the docker pull rate limit but it should be OK. One last thing before merging: could you add a changelog entry ?

jepio:
Done

PR #343, opened 2022-07-01
plume prune enhancement

jepio:
# plume prune enhancements

* support keeping N most recently created images when deleting
* support "soft-deleting" images (removing launchpermissions) for configurable time before permanently deleting
* support checking "lastlaunchtime" when image is being removed
* display statistics on image removal
* update aws go sdk

## How to use
```
./bin/plume prune --check-last-launched --dry-run --channel edge -v  --keep-last 2 --days 365 --days-soft-deleted 30
```

## Testing done

Tested with dry-run only, would run on "edge" channel first.

### DIFF
--- a/cmd/plume/prune.go
+++ b/cmd/plume/prune.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -32,6 +33,7 @@ import (
 
 var (
 	days        int
+	keeplast    int
 	pruneDryRun bool
 	cmdPrune    = &cobra.Command{
 		Use:   "prune --channel CHANNEL [options]",
@@ -43,6 +45,7 @@ var (
 
 func init() {
 	cmdPrune.Flags().IntVar(&days, "days", 30, "Minimum age in days for files to get deleted")
+	cmdPrune.Flags().IntVar(&keeplast, "keep-last", 0, "Number of latest images to keep")
 	cmdPrune.Flags().StringVar(&awsCredentialsFile, "aws-credentials", "", "AWS credentials file")
 	cmdPrune.Flags().StringVar(&azureProfile, "azure-profile", "", "Azure Profile json file")
 	cmdPrune.Flags().StringVar(&azureAuth, "azure-auth", "", "Azure Credentials json file")
@@ -184,6 +187,21 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 			}
 
 			plog.Infof("Got %d images with channel %q", len(images), specChannel)
+
+			// sort images by creation date
+			sort.Slice(images, func(i, j int) bool {
+				datei, _ := time.Parse(time.RFC3339Nano, *images[i].CreationDate)
+				datej, _ := time.Parse(time.RFC3339Nano, *images[j].CreationDate)
+				return datei.Before(datej)
+			})
+			if len(images) <= keeplast {
+				plog.Infof("Not enough images to prune, keeping %d", len(images))
+				continue
+			}
+			for _, image := range images[len(images)-keeplast:] {
+				plog.Infof("Keeping image %q", *image.Name)
+			}
+			images = images[:len(images)-keeplast]
 
 			now := time.Now()
 			for _, image := range images {
--- a/cmd/plume/prune.go
+++ b/cmd/plume/prune.go
@@ -178,12 +178,12 @@ func pruneAzure(ctx context.Context, spec *channelSpec) {
 }
 
 type deleteStats struct {
-	Total        int
-	Kept         int
-	Skipped      int
-	RecentlyUsed int
-	SoftDeleted  int
-	Deleted      int
+	total        int
+	kept         int
+	skipped      int
+	recentlyUsed int
+	softDeleted  int
+	deleted      int
 }
 
 func pruneAWS(ctx context.Context, spec *channelSpec) {
@@ -217,7 +217,7 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 			if err != nil {
 				plog.Fatalf("Couldn't list images in channel %q: %v", specChannel, err)
 			}
-			stats.Total += len(images)
+			stats.total += len(images)
 
 			plog.Infof("Got %d images with channel %q", len(images), specChannel)
 
@@ -229,13 +229,13 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 			})
 			if len(images) <= keepLast {
 				plog.Infof("Not enough images to prune, keeping %d", len(images))
-				stats.Kept += len(images)
+				stats.kept += len(images)
 				continue
 			}
 			for _, image := range images[len(images)-keepLast:] {
 				plog.Infof("Keeping image %q", *image.Name)
 			}
-			stats.Kept += keepLast
+			stats.kept += keepLast
 			images = images[:len(images)-keepLast]
 
 			now := time.Now()
@@ -248,7 +248,7 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 				daysOld := int(duration.Hours() / 24)
 				if daysOld < days {
 					plog.Infof("Valid image %q: %d days old, skipping", *image.Name, daysOld)
-					stats.Skipped += 1
+					stats.skipped += 1
 					continue
 				}
 				if checkLastLaunched {
@@ -261,7 +261,7 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 					daysOld := int(duration.Hours() / 24)
 					if daysOld < daysLastLaunched {
 						plog.Infof("Image %q: recently used %d days ago (%v), skipping", *image.Name, daysOld, lastLaunched)
-						stats.RecentlyUsed += 1
+						stats.recentlyUsed += 1
 						continue
 					}
 				}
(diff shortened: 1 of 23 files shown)


jepio:
Here's the log output for a dry-run on the edge channel:
[out.log](https://github.com/flatcar-linux/mantle/files/9028854/out.log)

cmd/plume/prune.go
```diff
@@ -30,9 +32,12 @@ import (
 )
 
 var (
-	days        int
-	pruneDryRun bool
-	cmdPrune    = &cobra.Command{
+	days              int
+	dayssoftdeleted   int
+	keeplast          int
```
krnowak, line 37: camelCase, please.

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
 					}
 					board := fmt.Sprintf("%s-usr", arch)
 					var version string
+					var softdeletedate string
```
krnowak, line 259: camelCase, please.

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+						_, err = api.RemoveLaunchPermission(*image.ImageId)
+						if err != nil {
+							plog.Fatalf("Error removing launch permission from %v: %v", *image.Name, err)
+						}
+						// add tag
+						err = api.CreateTags([]string{*image.ImageId}, map[string]string{"SoftDeleteDate": softdeletedate})
+						if err != nil {
+							plog.Fatalf("Error adding tag to %v: %v", *image.Name, err)
+						}
+						plog.Infof("Image %v has been soft deleted", *image.Name)
```
krnowak, line 280:
```suggestion
						plog.Infof("Image %v has been soft-deleted", *image.Name)
```

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+					} else if dayssoftdeleted != 0 {
+						// check if the image is still soft deleted
+						softdeletedate, err := time.Parse(time.RFC3339, softdeletedate)
+						if err != nil {
+							plog.Fatalf("Error converting soft delete date (%v): %v", softdeletedate, err)
+						}
+						duration := now.Sub(softdeletedate)
+						daysOld := int(duration.Hours() / 24)
+						if daysOld < dayssoftdeleted {
+							plog.Infof("Image %v soft deleted %d days ago, skipping", *image.Name, daysOld)
```
krnowak, line 292:
```suggestion
							plog.Infof("Image %v soft-deleted %d days ago, skipping", *image.Name, daysOld)
```

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+						// add tag
+						err = api.CreateTags([]string{*image.ImageId}, map[string]string{"SoftDeleteDate": softdeletedate})
+						if err != nil {
+							plog.Fatalf("Error adding tag to %v: %v", *image.Name, err)
+						}
+						plog.Infof("Image %v has been soft deleted", *image.Name)
+						stats.SoftDeleted += 1
+						continue
+					} else if dayssoftdeleted != 0 {
+						// check if the image is still soft deleted
```
krnowak, line 284:
```suggestion
						// check if the image is still soft-deleted
```

cmd/plume/prune.go
krnowak, line 37: camelCase, please. (If it appears as a double, sorry, the last one I wrote disappeared after a refresh).

krnowak (commented):
Nitpicks only.

cmd/plume/prune.go
jepio: ok

cmd/plume/prune.go
jepio: ok

cmd/plume/prune.go
jepio: ok

cmd/plume/prune.go
jepio: ok

cmd/plume/prune.go
jepio: ok

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+						}
+						// add tag
+						err = api.CreateTags([]string{*image.ImageId}, map[string]string{"SoftDeleteDate": softDeleteDate})
+						if err != nil {
+							plog.Fatalf("Error adding tag to %v: %v", *image.Name, err)
+						}
+						plog.Infof("Image %v has been soft-deleted", *image.Name)
+						stats.SoftDeleted += 1
+						continue
+					} else if daysSoftDeleted != 0 {
```
krnowak, line 283:
```suggestion
					} else if daysSoftDeleted > 0 {
```

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+						// check if the image is still soft-deleted
+						softdeletedate, err := time.Parse(time.RFC3339, softDeleteDate)
```
krnowak, line 285: camelCase, please. Likely need a different name, as you have `softdeletedate` and `softDeleteDate`.

cmd/plume/prune.go
krnowak: Maybe there should be a validation of the parameter as it should never be less than zero?

cmd/plume/prune.go
```diff
@@ -193,9 +230,24 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+					stats.Skipped += 1
 					continue
 				}
-				plog.Infof("Obsolete image %q: %d days old", *image.Name, daysOld)
+				if checkLastLaunched {
+					lastLaunched, err := api.GetImageLastLaunchedTime(*image.ImageId)
+					if err != nil {
+						plog.Warningf("Error converting last launched date (%v): %v", *image.ImageId, err)
+
+					}
```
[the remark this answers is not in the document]
jepio: I'll leave this as a warning but add a "continue" to skip the image. If we're requesting to check the lastLaunchedTime but fail with an error (which we don't expect to happen), then it's better to investigate/try again later.

cmd/plume/prune.go
```diff
@@ -204,11 +256,45 @@ func pruneAWS(ctx context.Context, spec *channelSpec) {
...
+					var softDeleteDate string
 					for _, t := range image.Tags {
 						if *t.Key == "Version" {
 							version = *t.Value
 						}
+						if *t.Key == "SoftDeleteDate" {
+							softDeleteDate = *t.Value
+						}
 					}
+					if softDeleteDate == "" && daysSoftDeleted != 0 {
```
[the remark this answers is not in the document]
jepio: ok

cmd/plume/prune.go
jepio: Changed this branch and added validation at the top.

cmd/plume/prune.go
jepio: Camelcased and renamed to `softDeleteDateTs`

krnowak (approved):

platform/api/aws/images.go
```diff
@@ -333,6 +334,42 @@ func (a *API) CreateHVMImage(snapshotID string, diskSizeGiB uint, name string, d
...
+					Group: aws.String("all"),
+				},
+			},
+		},
+	}
+	resp, err := a.ec2.ModifyImageAttribute(mod)
+	if err != nil {
+		return []byte{}, err
+	}
+	return json.Marshal(resp)
```
tormath1, line 352: Personal opinion: I tend to wrap the errors whenever it's possible to bring more context and to identify if the issue appears in the `ModifyImageAttribute` or in the `Marshal` of the value.
```suggestion
	resp, err := a.ec2.ModifyImageAttribute(mod)
	if err != nil {
		return []byte{}, fmt.Error("modifying image attribute: %w, err)
	}
	
	res, err := json.Marshal(resp)
	if err != nil {
		return []byte{}, fmt.Error("marshaling the API response: %w, err)
	}
	
	return res, nil
```

cmd/plume/prune.go
```diff
@@ -56,6 +67,18 @@ func runPrune(cmd *cobra.Command, args []string) {
 	if len(args) > 0 {
 		plog.Fatal("No args accepted")
 	}
+	if daysLastLaunched < 0 {
```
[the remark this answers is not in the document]
jepio: I think I would rather keep the if-blocks separate so that we can report a clear error messages to users. But you made me realize I need at least one more check: `if !checkLastLaunched && daysLastLaunched > 0` so that we tell users that we report that the option will be ignored.

platform/api/aws/images.go
```diff
@@ -333,6 +334,42 @@ func (a *API) CreateHVMImage(snapshotID string, diskSizeGiB uint, name string, d
 	return a.createImage(params)
 }
 
+func (a *API) AddLaunchPermission(imageid string) ([]byte, error) {
```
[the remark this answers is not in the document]
jepio: good catch, this was thought about adding this to ore but in the end didn't end up plumbing it through anywhere. will delete.

platform/api/aws/images.go
jepio: ok, added to RemoveLaunchPermission()

tormath1 (approved):
Thanks! :)

PR #345, opened 2022-07-06
kola/tests: Limit generic tests to QEMU

pothos:
The number of tests causes the test time to be many hours on some of
the platforms and also costs money and energy while many tests are not
really adding much value when being tested on each of the platforms.
Some very generic tests already were limited to QEMU which is a good
solution for now, even if that means that the execution of all possible
tests on a platform now requires a source code change because we don't
have a flag to allow running additional tests when wanted.
Limit the execution of generic tests to QEMU. These generic tests are
not expected to have a different result on a cloud environment unless
something fundamental is at error. For these fundamental things we
rely on tests that still get executed on each cloud, e.g., to test
Ignition and networking, but we also rely on end-to-end tests such as
the kubeadm test. Each test now has a comment that tells whether the
execution on all clouds is wanted or not and why. The hope is that with
focused tests we don't waste our time to wait for unnecessary tests or
even have to restart them multiple times because they are flaky. Also,
with a reduced list of tests we can easily add new ones that we think
are missing but didn't realize before because there were to many, or
didn't want to add because the test time was too high already.

See https://github.com/flatcar-linux/Flatcar/issues/654

## How to use

We can compare the `kola list --filter --platform=X` output before and after the change

## Testing done

```
kola list --filter --platform=qemu
kola list --filter --platform=aws
kola list --filter --platform=do
kola list --filter --platform=azure
kola list --filter --platform=gce
kola list --filter --platform=equinixmetal
```

Here the sample output for equinixmetal:
```
kola list --filter --platform=equinixmetal | cut  -f 1
Test Name

cl.basic
cl.cloudinit.basic
cl.etcd-member.discovery
cl.flannel.udp
cl.flannel.vxlan
cl.ignition.misc.empty
cl.ignition.v1.noop
cl.ignition.v2.btrfsroot
cl.ignition.v2.noop
cl.install.cloudinit
cl.internet
cl.metadata.equinixmetal
cl.network.initramfs.second-boot
coreos.ignition.once
coreos.ignition.resource.local
coreos.ignition.resource.remote
coreos.ignition.sethostname
coreos.ignition.ssh.key
docker.network
kubeadm.v1.22.7.calico.base
kubeadm.v1.22.7.calico.cgroupv1.base
kubeadm.v1.22.7.cilium.base
kubeadm.v1.22.7.cilium.cgroupv1.base
kubeadm.v1.22.7.flannel.base
kubeadm.v1.22.7.flannel.cgroupv1.base
kubeadm.v1.23.4.calico.base
kubeadm.v1.23.4.cilium.base
kubeadm.v1.23.4.flannel.base
kubeadm.v1.24.1.calico.base
kubeadm.v1.24.1.cilium.base
kubeadm.v1.24.1.flannel.base
```

### DIFF
--- a/kola/tests/misc/network.go
+++ b/kola/tests/misc/network.go
@@ -34,6 +34,8 @@ func init() {
 		ClusterSize: 1,
 		Name:        "cl.network.listeners",
 		Distros:     []string{"cl"},
+		// This test is normally not related to the cloud environment unless the OEM tools would unexpectedly listen on ports
+		Platforms: []string{"qemu", "qemu-unpriv"},
 		// be sure to notice listeners in the docker stack
 		UserData: conf.ContainerLinuxConfig(`systemd:
   units:
@@ -47,6 +49,8 @@ func init() {
 		Name:        "cl.network.listeners.legacy",
 		Distros:     []string{"cl"},
 		EndVersion:  semver.Version{Major: 1967},
+		// This test is normally not related to the cloud environment unless the OEM tools would unexpectedly listen on ports
+		Platforms: []string{"qemu", "qemu-unpriv"},
 	})
 	register.Register(&register.Test{
 		Run:              NetworkInitramfsSecondBoot,
--- a/kola/tests/bpf/bpf.go
+++ b/kola/tests/bpf/bpf.go
@@ -42,6 +42,8 @@ func init() {
 		Flags: []register.Flag{register.NoEnableSelinux},
 		// exclude `arm64` while `quay.io/iovisor/bcc` does not have `arm64` support.
 		Architectures: []string{"amd64"},
+		// This test is normally not related to the cloud environment
+		Platforms: []string{"qemu", "qemu-unpriv"},
 	})
 }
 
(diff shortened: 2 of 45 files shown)


kola/tests/misc/network.go
```diff
@@ -47,13 +49,17 @@ func init() {
...
+		// This test is normally not related to the cloud environment unless the OEM tools would unexpectedly listen on ports
+		Platforms: []string{"qemu", "qemu-unpriv"},
 	})
 	register.Register(&register.Test{
 		Run:              NetworkInitramfsSecondBoot,
 		ClusterSize:      1,
 		Name:             "cl.network.initramfs.second-boot",
 		ExcludePlatforms: []string{"do"},
 		Distros:          []string{"cl"},
+		// This test is normally not related to the cloud environment unless the OEM tools would unexpectedly listen on ports
```
krnowak, line 61: Is this test (cl.network.initramfs.second-boot) still about listening on ports?

tormath1 (approved):
LGTM. Thanks!

kola/tests/misc/network.go
pothos: I thought so but it's not, thanks. The test checks that networks is not started in the initramfs on the second boot. I removed the limitation to qemu as it's an integration with the cloud environment in the initramfs that's good to check.

krnowak (approved):

pothos:
Ok, merging it now, please ping me if you see something strange ;)

PR #348, opened 2022-07-29
test/networkd: add wireguard setup

tormath1:
simple setup to test IP is assigned as expected.

<hr>

To be tested with: https://github.com/flatcar-linux/coreos-overlay/pull/2061 - currently it's failing but it works when adding the following section:
```yml
storage:
  files:
    - path: /etc/systemd/network/zz-default.network.d/wg.conf
      contents:
        inline: |
          [Match]
          Type=!wireguard
```

as suggested in the `::coreos-overlay` PR.


jepio (approved):

jepio:
We could also make the test slightly more realistic, and use 2 vms, check that they can communicate etc, but this should be good enough for now.

tormath1:
@jepio yeah I also wanted to have an actual provisioning to generate wireguard keys during the boot but it can also be done in a follow-up PR.

PR #349, opened 2022-08-05
openstack: run test with only 1 public IP

tormath1:
When spawning instance in OpenStack, VMs need to attach an interface with a public IP in order to SSH into it. With this change, we can now SSH proxy jump into the `devstack` host (Openstack dev environment)  into the VM deployed in the public network.

tmp hsExzPr6Z4

This reduce the dependency to floating IPs.

Related to: https://github.com/flatcar-linux/Flatcar/issues/785

Tests:
```
sudo ./bin/kola run --openstack-network public --openstack-domain default --platform openstack --openstack-flavor ds1G --openstack-user core --openstack-host 147.28... --openstack-keyfile ./id_rsa --openstack-image flatcar-stable --openstack-config-file ./openstack.json bpf.execsnoop
```

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -161,6 +161,9 @@ func init() {
 	sv(&kola.OpenStackOptions.Network, "openstack-network", "", "OpenStack network")
 	sv(&kola.OpenStackOptions.Domain, "openstack-domain", "", "OpenStack domain ID")
 	sv(&kola.OpenStackOptions.FloatingIPPool, "openstack-floating-ip-pool", "", "OpenStack floating IP pool for Compute v2 networking")
+	sv(&kola.OpenStackOptions.Host, "openstack-host", "", "Host can be used to optionally SSH into deployed VMs from the OpenStack host")
+	sv(&kola.OpenStackOptions.User, "openstack-user", "", "User is the one used for the SSH connection to the Host")
+	sv(&kola.OpenStackOptions.Keyfile, "openstack-keyfile", "", "Keyfile is the absolute path to private SSH key file for the User on the Host")
 
 	// packet-specific options (kept for compatiblity but marked as deprecated)
 	sv(&kola.EquinixMetalOptions.ConfigPath, "packet-config-file", "", "Packet config file (default \"~/"+auth.EquinixMetalConfigPath+"\")")
--- a/network/jump.go
+++ b/network/jump.go
@@ -1,0 +1,67 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package network
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+
+	"golang.org/x/crypto/ssh"
+)
+
+var _ Dialer = (*JumpDialer)(nil)
+
+// JumpDialer is a RetryDialer from a proxy.
+type JumpDialer struct {
+	RetryDialer
+	user    string
+	addr    string
+	keyfile string
+}
+
+// NewJumpDialer initializes a JumpDialer to establish a SSH Proxy Jump.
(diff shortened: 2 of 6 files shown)


cmd/kola/options.go
```diff
@@ -161,6 +161,9 @@ func init() {
 	sv(&kola.OpenStackOptions.Network, "openstack-network", "", "OpenStack network")
 	sv(&kola.OpenStackOptions.Domain, "openstack-domain", "", "OpenStack domain ID")
 	sv(&kola.OpenStackOptions.FloatingIPPool, "openstack-floating-ip-pool", "", "OpenStack floating IP pool for Compute v2 networking")
+	sv(&kola.OpenStackOptions.Host, "openstack-host", "", "Host can be used to optionally SSH into deployed VMs from the OpenStack host")
+	sv(&kola.OpenStackOptions.User, "openstack-user", "", "User is the one used for the SSH connection to the Host")
+	sv(&kola.OpenStackOptions.Keyfile, "openstack-keyfile", "", "Keyfile is the abs. path to private SSH key file for the User on the Host")
```
krnowak, line 166:
```suggestion
	sv(&kola.OpenStackOptions.Keyfile, "openstack-keyfile", "", "Keyfile is the absolute path to private SSH key file for the User on the Host")
```

platform/api/openstack/api.go
```diff
@@ -64,6 +64,15 @@ type Options struct {
 	Domain string
 	// Floating IP Pool
 	FloatingIPPool string
+
+	// Host can be used to optionally SSH into deployed VMs from the OpenStack host
+	Host string
+
```
krnowak, line 70: Unnecessary empty line.

platform/api/openstack/api.go
```diff
@@ -64,6 +64,15 @@ type Options struct {
...
+	// User is the one used for the SSH connection to the Host
+	User string
+
```
krnowak, line 73: Ditto.

platform/machine/openstack/flight.go
```diff
@@ -44,9 +47,18 @@ func NewFlight(opts *openstack.Options) (platform.Flight, error) {
 		return nil, err
 	}
 
-	bf, err := platform.NewBaseFlight(opts.Options, Platform, ctplatform.OpenStackMetadata)
-	if err != nil {
-		return nil, err
+	var bf *platform.BaseFlight
+
+	if opts.Host != "" {
```
krnowak, line 52: Should we check if `opts.User` and `opts.Keyfile` are not empty? Like:

```go
if host != "" {
	if user == "" || keyfile == "" {
		return nil, fmt.Errorf("argh!")
	}
	…
} else {
	if user != "" || keyfile != "" {
		warning("user and keyfile ignored")
	}
	…
}
```

krnowak (commented):
Is this going to be used by a vendor test?

tormath1:
@krnowak thanks for the review. Yes, the idea is to add it as a vendor test, we recently got a couple of issues related to openstack provider, that'd be nice to ensure coverage on this platform.

network/jump.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+
+	"golang.org/x/crypto/ssh"
+)
+
+// JumpDialer is a RetryDialer from a proxy.
+type JumpDialer struct {
```
krnowak, line 14: Might be a good idea to also add `var _ network.Dialer = (*JumpDialer)(nil)` (will need importing "github.com/flatcar-linux/mantle/network") to show in the code that this type is supposed to implement a certain interface.

tormath1: Fun, never seen this :)

krnowak (approved):

tormath1:
did a final test with `kubeadm.v1.23.4.flannel.base`.

PR #350, opened 2022-08-09
openstack: create image from URL

tormath1:
In this PR, we add the ability to `ore openstack` to create an image from an URL - this is quite handy for CI context, we could just provide the bincache URL to create the Openstack image of the tested version.

```
$ IMAGE_ID=$(./bin/ore openstack create-imag --file https://alpha.release.flatcar-linux.net/amd64-usr/3305.0.0/flatcar_production_openstack_image.img.gz ...)
$ ./bin/kola run ... --platform openstack --openstack-image "${IMAGE_ID}" kubeadm.v1.23.4.flannel.base
```

Vendor has been upgraded to the most recent version (`0.25.0`)

Related to: https://github.com/flatcar-linux/Flatcar/issues/785


pothos (approved):
Looks good but I didn't test it. One related task is to teach `ore gc` to also delete left-over images and not only left-over servers.

tormath1:
> Looks good but I didn't test it. One related task is to teach `ore gc` to also delete left-over images and not only left-over servers.

@pothos thanks for the review. I was thinking to add SSH keys deletion too.

PR #351, opened 2022-08-09
network: create one connection to the jump host

tormath1:
this avoid to create a new connection for each run, we just create one
shared SSH connection for one flight instance.

<hr>

Tried with `--parallel 2 kubeadm.v1.23.4.flannel.base kubeadm.v1.23.4.calico.base`

### DIFF
--- a/network/jump.go
+++ b/network/jump.go
@@ -5,40 +5,13 @@ package network
 import (
 	"fmt"
 	"io/ioutil"
-	"net"
 
 	"golang.org/x/crypto/ssh"
 )
 
-var _ Dialer = (*JumpDialer)(nil)
-
-// JumpDialer is a RetryDialer from a proxy.
-type JumpDialer struct {
-	RetryDialer
-	user    string
-	addr    string
-	keyfile string
-}
-
-// NewJumpDialer initializes a JumpDialer to establish a SSH Proxy Jump.
-func NewJumpDialer(addr, user, keyfile string) *JumpDialer {
-	return &JumpDialer{
-		RetryDialer: RetryDialer{
-			Dialer: &net.Dialer{
-				Timeout:   DefaultTimeout,
-				KeepAlive: DefaultKeepAlive,
-			},
-			Retries: DefaultRetries,
-		},
-		user:    user,
-		addr:    addr,
-		keyfile: keyfile,
-	}
-}
-
-// Dial connects to a remote address, retrying on failure.
-func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
-	key, err := ioutil.ReadFile(d.keyfile)
+// NewJumpDialer initializes a RetryDialer with SSH Proxy Jump.
+func NewJumpDialer(addr, user, keyfile string) (*RetryDialer, error) {
+	key, err := ioutil.ReadFile(keyfile)
 	if err != nil {
 		return nil, fmt.Errorf("reading private key: %w", err)
 	}
@@ -49,7 +22,7 @@ func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
 	}
 
 	cfg := &ssh.ClientConfig{
-		User: d.user,
+		User: user,
 		// this is only used for testing - it's ok to live with that.
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 		Auth: []ssh.AuthMethod{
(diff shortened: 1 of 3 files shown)


network/jump.go
```diff
@@ -49,19 +31,33 @@ func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
...
+	return &JumpDialer{
+		RetryDialer: RetryDialer{
+			Dialer: &net.Dialer{
+				Timeout:   DefaultTimeout,
+				KeepAlive: DefaultKeepAlive,
+			},
+			Retries: DefaultRetries,
+		},
+		client: client,
+	}, nil
```
pothos, line 57: Is the retry dialer actually used? I don't see it used.

I think you can return the dialer directly instead of implementing the dialer interface for `JumpDialer`.

```suggestion
	return &RetryDialer{
			Dialer: client,
			Retries: DefaultRetries,
		}, nil
```

network/jump.go
```diff
@@ -15,30 +15,12 @@ var _ Dialer = (*JumpDialer)(nil)
...
-		user:    user,
-		addr:    addr,
-		keyfile: keyfile,
-	}
-}
-
-// Dial connects to a remote address, retrying on failure.
-func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
-	key, err := ioutil.ReadFile(d.keyfile)
+func NewJumpDialer(addr, user, keyfile string) (*JumpDialer, error) {
```
pothos, line 22: Following the suggestion below, then do this
```suggestion
func NewJumpDialer(addr, user, keyfile string) (*RetryDialer, error) {
```

network/jump.go
```diff
@@ -15,30 +15,12 @@ var _ Dialer = (*JumpDialer)(nil)
 // JumpDialer is a RetryDialer from a proxy.
 type JumpDialer struct {
 	RetryDialer
-	user    string
-	addr    string
-	keyfile string
+	client *ssh.Client
 }
```
pothos, line 19: The struct would then be not needed anymore

network/jump.go
tormath1: Ooh ok in this way - I misunderstood the initial message then :)

pothos: GitHub suggestions may be confusing, so here as diff for `git apply`:

```
diff --git a/network/jump.go b/network/jump.go
index e2cae07f..5fd4d58b 100644
--- a/network/jump.go
+++ b/network/jump.go
@@ -5,40 +5,13 @@ package network
 import (
 	"fmt"
 	"io/ioutil"
-	"net"
 
 	"golang.org/x/crypto/ssh"
 )
 
-var _ Dialer = (*JumpDialer)(nil)
-
-// JumpDialer is a RetryDialer from a proxy.
-type JumpDialer struct {
-	RetryDialer
-	user    string
-	addr    string
-	keyfile string
-}
-
 // NewJumpDialer initializes a JumpDialer to establish a SSH Proxy Jump.
-func NewJumpDialer(addr, user, keyfile string) *JumpDialer {
-	return &JumpDialer{
-		RetryDialer: RetryDialer{
-			Dialer: &net.Dialer{
-				Timeout:   DefaultTimeout,
-				KeepAlive: DefaultKeepAlive,
-			},
-			Retries: DefaultRetries,
-		},
-		user:    user,
-		addr:    addr,
-		keyfile: keyfile,
-	}
-}
-
-// Dial connects to a remote address, retrying on failure.
-func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
-	key, err := ioutil.ReadFile(d.keyfile)
+func NewJumpDialer(addr, user, keyfile string) (*RetryDialer, error) {
+	key, err := ioutil.ReadFile(keyfile)
 	if err != nil {
 		return nil, fmt.Errorf("reading private key: %w", err)
 	}
@@ -49,7 +22,7 @@ func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
 	}
 
 	cfg := &ssh.ClientConfig{
-		User: d.user,
+		User: user,
 		// this is only used for testing - it's ok to live with that.
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 		Auth: []ssh.AuthMethod{
@@ -57,11 +30,14 @@ func (d *JumpDialer) Dial(network, address string) (c net.Conn, err error) {
 		},
 	}
 
-	addr := ensurePortSuffix(d.addr, defaultPort)
+	addr = ensurePortSuffix(addr, defaultPort)
 	client, err := ssh.Dial("tcp", addr, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("creating SSH client: %w", err)
 	}
 
-	return client.Dial(network, address)
+	return &RetryDialer{
+		Dialer:  client,
+		Retries: DefaultRetries,
+	}, nil
 }
```

pothos (approved):
Assuming it still works after the changes ;)

tormath1:
> Assuming it still works after the changes ;)

Ran a test (`kubeadm.v1.23.4.flannel.base`) before pushing.

tormath1:
will fix the go fmt issue in a following PR.

PR #352, opened 2022-08-09
mod: update to go1.19

tormath1:
This has been detected in the previous PR(https://github.com/flatcar-linux/mantle/pull/351): https://github.com/flatcar-linux/mantle/actions/runs/2825875438. The only major update is around the spaces / indentations in the files but it was breaking the CI.

We don't need to follow Flatcar Go version, since Mantle has been dropped from the SDK.  


pothos (approved):
Unfortunate…

PR #353, opened 2022-08-10
kola/tests/kubeadm: Fix version check

krnowak:
The version is prefixed with the v character, so not sure if this
check ever worked before - the test was still being run for LTS 2011,
which has version 2605, so lower than 3033.


pothos (approved):
Thanks

PR #355, opened 2022-08-10
os/api: add image deletion in the GC

tormath1:

<hr>

OpenStack images will be created on the fly in the CI, we need to clean-up created images with `mantle` tags.

I don't think it needs a changelog entry - tested on the OpenStack setup.

### DIFF
--- a/platform/api/openstack/api.go
+++ b/platform/api/openstack/api.go
@@ -610,6 +610,24 @@ func (a *API) listServersWithMetadata(metadata map[string]string) ([]servers.Ser
 	return retServers, nil
 }
 
+func (a *API) listImagesWithTags(tags []string) ([]images.Image, error) {
+	listOpts := images.ListOpts{
+		Tags: tags,
+	}
+
+	allPages, err := images.List(a.imageClient, listOpts).AllPages()
+	if err != nil {
+		return nil, fmt.Errorf("listing images: %w", err)
+	}
+
+	allImages, err := images.ExtractImages(allPages)
+	if err != nil {
+		return nil, fmt.Errorf("extracting image list: %w", err)
+	}
+
+	return allImages, nil
+}
+
 func (a *API) GC(gracePeriod time.Duration) error {
 	threshold := time.Now().Add(-gracePeriod)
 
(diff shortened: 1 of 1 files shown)


platform/api/openstack/api.go
```diff
@@ -628,5 +646,17 @@ func (a *API) GC(gracePeriod time.Duration) error {
...
 	}
+
+	images, err := a.listImagesWithTags([]string{"mantle"})
+	if err != nil {
+		return fmt.Errorf("listing Mantle images: %w", err)
+	}
+
+	for _, image := range images {
+		if err := a.DeleteImage(image.ID); err != nil {
+			return fmt.Errorf("deleting image with name: %s", image.Name)
```
pothos, line 657: I think the gc functions should continue to delete as much as possible but since the server deletion errors out early the same way it's ok to have it consistent.

pothos (approved):

platform/api/openstack/api.go
tormath1: Yeah, I got the same thought - and garbage collector job is running periodically so it's ok if it fails once.

krnowak: It looks like it's trying to remove all the images with the "mantle" tag. Should there be a check if image is used or, I dunno, remove only images older than some days?

pothos: Oh, right, can we do something like done for servers and check `.Created.After(threshold)`?

tormath1: Good idea - it seems there is no way to detect if an image is being used. Let's add the threshold!

pothos (approved):

platform/api/openstack/api.go
```diff
@@ -628,5 +646,21 @@ func (a *API) GC(gracePeriod time.Duration) error {
...
+
+	images, err := a.listImagesWithTags([]string{"mantle"})
+	if err != nil {
+		return fmt.Errorf("listing Mantle images: %w", err)
+	}
+
+	for _, image := range images {
+		if image.CreatedAt.After(threshold) {
+			continue
+		}
```
krnowak, line 658: I just noticed that the similar thing could also be achieved with `ListOpts.CreatedAtQuery`.

tormath1: Cool, this can be include in a follow-up PR, for example this whole section could be simplified: https://github.com/flatcar-linux/mantle/blob/9fc1820fc76db08a7670b7b59c75c0c4b87e5561/platform/api/openstack/api.go#L585-L611

PR #356, opened 2022-08-11
sdk/verify.go: Update pubkey to include new subkey

pothos:
A new subkey was generated as the old expired.


tormath1 (approved):

pothos:
We need a new cork release now (at least for the source mirror job on Origin)

pothos:
… or I'll try to use the new sdk container for it but not sure if I can finish it this week.

PR #359, opened 2022-08-22
kola: More calico fixes

krnowak:
~I'll need to update the github action to mention in the created PR that these templates should be manually compared with the original versions.~

There seemed to be a race between using one manifest file to create CRDs and afterwards applying another manifest file that was using those CRDs. Sometimes the CRDs were not established, so the resources of that type could not be used. Use `kubectl wait` to make sure that those CRDs are ready.

While at it, sync our modified custom-resources.yaml manifest with the upstream and add a note about it to the github action. The note looks like in this PR: https://github.com/krnowak/mantle/pull/2.

Test run: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/1988/console (note the `customresourcedefinition.apiextensions.k8s.io/installations.operator.tigera.io condition met` line in the output, after CRDs were created, but before the object of that type was created).

### DIFF
--- a/kola/tests/kubeadm/templates.go
+++ b/kola/tests/kubeadm/templates.go
@@ -384,6 +384,9 @@ EOF
 
 {{ if eq .CNI "calico" }}
     kubectl create -f https://raw.githubusercontent.com/projectcalico/calico/v3.24.0/manifests/tigera-operator.yaml
+    # calico.yaml uses Installation and APIServer CRDs, so make sure that they are established.
+    kubectl -n tigera-operator wait --for condition=established --timeout=60s crd/installations.operator.tigera.io
+    kubectl -n tigera-operator wait --for condition=established --timeout=60s crd/apiservers.operator.tigera.io
     kubectl apply -f calico.yaml
 {{ end }}
 {{ if eq .CNI "flannel" }}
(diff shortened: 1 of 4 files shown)


krnowak:
Another test run: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/1989/console

krnowak:
Tested with my toolchain update build: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/204/cldsv/

kola/tests/kubeadm/templates.go
```diff
@@ -353,12 +355,22 @@ spec:
...
+    - blockSize: 26
+      cidr: {{ .PodSubnet }}
+      encapsulation: VXLANCrossSubnet
+      natOutgoing: Enabled
+      nodeSelector: all()
   flexVolumePath: /opt/libexec/kubernetes/kubelet-plugins/volume/exec/
+
+---
+
+# This section configures the Calico API server.
```
tormath1, line 367: Is this section mandatory? Or it's just a placeholder for a later further customization?

krnowak: I dunno. It appeared in the custom resources manifest file in Calico between 3.19 and 3.20. I think it's best to have it to have our test calico setup as closest as possible to what upstream suggests in their quickstart guide.

tormath1 (approved):
Thanks a lot for digging into this - it should be more reliable in this way. :)

krnowak:
Thanks for the review. I just remembered that this needs an entry in the changelog. Added one and merged. If you have some suggestions about the entry, I'll fix eventually.

PR #361, opened 2022-08-24
test/network: add vmware to wireguard test

tormath1:

<hr>

It has been raised on Matrix that wireguard does not seem working on VMWare.

### DIFF
--- a/kola/tests/misc/network.go
+++ b/kola/tests/misc/network.go
@@ -81,7 +81,7 @@ storage:
 
           [Peer]
           PublicKey = EQluuT9Wk0TFUDYlU/5fBsY3KPx/YsZxaM0lCvlLwF4=
-          AllowedIPs = 0.0.0.0/0, ::/0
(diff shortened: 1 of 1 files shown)


krnowak (commented):
So I can understand why you changed `0.0.0.0` to `192.168.0.0`, but why changing `10.100.0.2` to `172.16.0.2`?

tormath1:
> So I can understand why you changed `0.0.0.0` to `192.168.0.0`, but why changing `10.100.0.2` to `172.16.0.2`?

Correct, I initially changed from `10.100.0.2` to `172.16.0.2` because I was thinking it could conflict with other routes and forgot to change it back to normal.

krnowak (approved):

tormath1:
commit reworded and test just ran on VMWare: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/vmware-esx/1039/tapResults/

PR #364, opened 2022-08-29
equinixmetal: Exclude weird devices from flatcar installation

krnowak:
`s3.xlarge.x86` has some nvme disks attached, which have the smallest size in bytes from all the available disks. Since we are passing the `-s` flag to `flatcar-install` (so the scripts find the smallest disk to install flatcar), they get picked. For some reason, they can't be booted. The nvme devices have major 259, whereas other disks - 8.

I had two ways of fixing the issue - either exclude 259, or include only 8. Went with the former.

During trying to find the fix I did some small cleanup.

The commit messages still need to be rewritten.

I also had a commit that drops fiddling with AlwaysPXE flag, and uses reinstall action instead. This might be something to consider, but it's certainly slower, since all the disks seem to be zeroed. So in case of `s3.xlarge.x86` instance, the reinstallation caused the timeout to be triggered. Maybe something for follow-up PR.

Fixes https://github.com/flatcar-linux/Flatcar/issues/834

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 - Fix version check in kubeadm tests ([#353](https://github.com/flatcar-linux/mantle/pull/353))
 - Make Calico testing in kubeadm tests more reliable ([#359](https://github.com/flatcar-linux/mantle/pull/359))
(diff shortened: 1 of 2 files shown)


krnowak:
CI is still running, but tests for s3.xlarge.x86 have already succeeded: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/2168/

tormath1 (approved):
Nice catch!

PR #365, opened 2022-08-30
kubeadm/cilium: bump tested versions

tormath1:
* kubeadm/cilium: bump CLI and tested Cilium version

    Starting from Cilium 1.12, {live,ready}ness probes are on :9879 while
    it was on :9876 on older versions. (See: https://github.com/cilium/cilium/commit/22cd47ef496be1c0b78ff8d146b2240810e78978)

    CLI made this change on versions greater or equal to 1.10.12 (See: https://github.com/cilium/cilium-cli/pull/869/files) - it
    results with a port mismatch 9879/9876 if we test the version 1.11.5.

    Basically Cilium is running fine but its status is not ready/healthy
    from a Kubernetes PoV.

* kubeadm/cilium: patch Cilium daemon set

    This is required even with Permissive mode. Can be dropped once `spc_t`
    is supported on Flatcar.

* kubeadm/cilium: disable SELinux for Cilium

    Once the daemon set is started with `unconfined_t` it does not work with
    enforced SELinux because it hits a denial from transitioning to `kernel_t` to
    `unconfined_t` (and this normal because currently everything runs with
    `kernel_t` including container runtimes)

    Can be dropped once it works fine with `spc_t` label with the SELinux
    upgrade.

<hr>

Tested multiple times on Openstack, the CI: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/packet/309/console and with kubernetes-1.25

### DIFF
--- a/kola/tests/kubeadm/kubeadm.go
+++ b/kola/tests/kubeadm/kubeadm.go
@@ -54,8 +54,12 @@ var (
 					_ = c.MustSSH(controller, "/opt/bin/cilium uninstall")
 					version := params["CiliumVersion"].(string)
 					cidr := params["PodSubnet"].(string)
-					cmd := fmt.Sprintf("/opt/bin/cilium install --config enable-endpoint-routes=true --config cluster-pool-ipv4-cidr=%s --version=%s --encryption=ipsec --wait --wait-duration 1m", cidr, version)
-					_ = c.MustSSH(controller, cmd)
+					cmd := fmt.Sprintf("/opt/bin/cilium install --config enable-endpoint-routes=true --config cluster-pool-ipv4-cidr=%s --version=%s --encryption=ipsec --wait=false --restart-unmanaged-pods=false --rollback=false", cidr, version)
+					_, _ = c.SSH(controller, cmd)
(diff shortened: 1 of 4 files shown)


kola/tests/kubeadm/kubeadm.go
```diff
@@ -54,8 +54,12 @@ var (
 					_ = c.MustSSH(controller, "/opt/bin/cilium uninstall")
 					version := params["CiliumVersion"].(string)
 					cidr := params["PodSubnet"].(string)
-					cmd := fmt.Sprintf("/opt/bin/cilium install --config enable-endpoint-routes=true --config cluster-pool-ipv4-cidr=%s --version=%s --encryption=ipsec --wait --wait-duration 1m", cidr, version)
-					_ = c.MustSSH(controller, cmd)
+					cmd := fmt.Sprintf("/opt/bin/cilium install --config enable-endpoint-routes=true --config cluster-pool-ipv4-cidr=%s --version=%s --encryption=ipsec --wait-duration=1s --rollback=false", cidr, version)
```
krnowak, line 57: Specifying wait duration is most likely pointless if we are not waiting:

```suggestion
					cmd := fmt.Sprintf("/opt/bin/cilium install --config enable-endpoint-routes=true --config cluster-pool-ipv4-cidr=%s --version=%s --encryption=ipsec --rollback=false", cidr, version)
```

krnowak (approved):
One nitpick, otherwise looks good. Also, I think it needs an entry in the changelog.

kola/tests/kubeadm/kubeadm.go
tormath1: It actually waits (`--wait` defaults to `true`) - but it made me think that we can `--wait=false --restart-unmanaged-pods=false` to not wait at all, this is enough for this test case.

krnowak: Cool. If that works then feel free to merge the PR.

PR #366, opened 2022-09-01
kola/tests/kubeadm: Skip 1.25 on old versions too

krnowak:
But try to make it general, so the information about the test are in a
single place.

Not adding the changelog - it's a fix for a recent feature.


krnowak:
Worked in CI (no kubeadm 1.24 and 1.25 in tests on old LTS): http://jenkins.infra.kinvolk.io:8080/job/container/job/test/2306/consoleText

pothos (approved):

PR #367, opened 2022-09-05
kola/tests/devcontainer: New test

krnowak:
This is to have a dev container test in the new pipeline. It runs only on qemu.

### DIFF
--- a/kola/tests/devcontainer/devcontainer.go
+++ b/kola/tests/devcontainer/devcontainer.go
@@ -218,6 +218,8 @@ func init() {
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
 		Distros:   []string{"cl"},
+		// TODO: Revisit this flag when updating SELinux policies.
+		Flags: []register.Flag{register.NoEnableSelinux},
 	})
 }
 
--- a/kola/tests/devcontainer/devcontainer.go
+++ b/kola/tests/devcontainer/devcontainer.go
@@ -1,0 +1,292 @@
+// Copyright 2022 The Flatcar Maintainers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package devcontainer
+
+import (
+	"bytes"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"strings"
+	"text/template"
+	"unicode"
+
+	"github.com/coreos/go-semver/semver"
+
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/platform"
+	"github.com/flatcar/mantle/platform/conf"
+	"github.com/flatcar/mantle/platform/machine/qemu"
+	"github.com/flatcar/mantle/platform/machine/unprivqemu"
+)
+
+// Both template parameters may contain @ARCH@ and @VERSION@
+// placeholders, which will be substituted by real values at test run
+// time.
+type scriptTemplateParameters struct {
+	BinhostURLTemplate        string
+	ImageDirectoryURLTemplate string
+}
+
+type configTemplateParameters struct {
+	DevContainerScriptBase64Contents string
+	MainScriptBase64Contents         string
+}
+
+func trimLeftSpace(contents string) string {
+	return strings.TrimLeftFunc(contents, unicode.IsSpace)
+}
+
+var (
+	defaultScriptTemplateParameters = scriptTemplateParameters{
+		BinhostURLTemplate:        "http://bincache.flatcar-linux.net/boards/@ARCH@-usr/@VERSION@/pkgs",
+		ImageDirectoryURLTemplate: "http://bincache.flatcar-linux.net/images/@ARCH@/@VERSION@",
+	}
+
+	devContainerScriptTemplate = trimLeftSpace(`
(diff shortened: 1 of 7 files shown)


krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/2347/

krnowak:
Meh, not enough space on the device, will need to come up with something, maybe mount some local disk?

```
++++ qemu.sh: downloading flatcar_production_image.bin for 3348.0.0+nightly-20220831-2100 (amd64) ++++
+ kola run --board=amd64-usr --parallel=8 --platform=qemu --qemu-bios=/usr/share/qemu/bios-256k.bin --qemu-image=flatcar_production_image.bin --tapfile=/work/__TESTS__/qemu/results-run-1.tap --torcx-manifest=/work/__TESTS__/torcx_manifest.json --qemu-skip-mangle 'devcontainer.*'
2022-09-06T15:39:32Z kola: Using "3348.0.0" as version to filter tests...
=== RUN   devcontainer.docker
=== RUN   devcontainer.systemd-nspawn
--- FAIL: devcontainer.systemd-nspawn (29.42s)
        cluster.go:117: cp: error writing 'flatcar_developer_container.bin': No space left on device
        devcontainer.go:207: main script failed: Process exited with status 1
--- FAIL: devcontainer.docker (29.69s)
        cluster.go:117: cp: error writing 'flatcar_developer_container.bin': No space left on device
        devcontainer.go:207: main script failed: Process exited with status 1
FAIL, output in _kola_temp/qemu-2022-09-06-1539-66
harness: test suite failed
```

pothos:
A test flag in kola to increase the qemu image size would be nice

pothos (commented):
We had to disable the mangling in the GitHub Action PR builds because they use lxc containers. I propose we don't increase the size of the backing file but we increase the size of the VM disk directly in `setupDisk` where `qemu-img create` is called and we can pass a new image size by adding a final argument.

pothos:
Adding `size=NEWSIZE` to the options here https://github.com/flatcar-linux/mantle/blob/4371409823df871a691c152eba766b6b981ddf1d/platform/qemu.go#L212 also works and then we would just do it for the disks with a backing file (and only if a new size is passed). For simplicity I would require the final size to be passed by the kola argument instead of trying to fetch the size of the backing file (that would need `qemu-img info --output json` to get `virtual-size`).

pothos:
Splitting it up to avoid losetup is also good ;)

platform/machine/qemu/flight.go
```diff
@@ -82,13 +92,20 @@ func NewFlight(opts *Options) (platform.Flight, error) {
 		if info.Format != "raw" {
```
pothos, line 92: nit: The resize is possible for qcow, just the console modification not

platform/qemu.go
```diff
@@ -76,10 +77,45 @@ func MakeCLDiskTemplate(inputPath string) (output *os.File, result error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("opening %v: %v", outputPath, err)
+	}
+	return output, nil
+}
+
+func resizeRawDisk(disk, extraSize string) error {
+	// Resize the base disk, booting the image should claim the
+	// free space.
+	resize := exec.Command("qemu-img", "resize", "-f", "raw", disk,
```
pothos, line 99:
```suggestion
func resizeDisk(disk, extraSize string) error {
	// Resize the base disk, booting the image should claim the
	// free space.
	resize := exec.Command("qemu-img", "resize", disk,
```
This works for both raw and qcow2

krnowak:
I have for now dropped the changes that added an option for growing the base disk and went with the RAID test way - using the type switch and checking for qemu and qemu-unpriv clusters.

I have saved my previous work and some refactoring in a separate branch, maybe will revisit it at some point.

krnowak:
> Adding `size=NEWSIZE` to the options here
> 
> https://github.com/flatcar-linux/mantle/blob/4371409823df871a691c152eba766b6b981ddf1d/platform/qemu.go#L212
> 
> also works and then we would just do it for the disks with a backing file (and only if a new size is passed). For simplicity I would require the final size to be passed by the kola argument instead of trying to fetch the size of the backing file (that would need `qemu-img info --output json` to get `virtual-size`).

I went with that route, but the size is added to the virtual size, because at the point where I'm doing it, I already have this information. And still no flag, that maybe could be hooked up later.

pothos (approved):
Thanks

krnowak:
@pothos: Hooking up the kola option was surprisingly easy - https://github.com/flatcar-linux/mantle/commit/d55d115b368132a9675d2643f8c1b20c1ea91170

Want me to add this option to this PR too?

pothos:
Sounds good, one can also use it with kola spawn.

kola/tests/devcontainer/devcontainer.go
```diff
@@ -0,0 +1,282 @@
...
+docker import dev-container-image.tar.gz dev-container:42
+sudo rm -f dev-container-image.tar.gz
+stat /home/core/dev-container-script >&2
+ls -la /home/core >&2
+
+docker run \
+        --env PORTAGE_BINHOST="${PORTAGE_BINHOST}" \
+        --env EXPECTED_VERSION="${EXPECTED_VERSION}" \
+        --mount type=bind,source=/lib/modules,target=/lib/modules,readonly=true \
+        --mount type=bind,source=/home/core/dev-container-script,target=/home/core/dev-container-script,readonly=true \
```
pothos, line 169:
```suggestion
        -v /home/core/dev-container-script:/home/core/dev-container-script:z \
```
See https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label on how to grant access by relabeling

krnowak: It basically modifies the label on the host, so I went with calling restorecon. Let's see if it does the trick.

krnowak: Also, I'm following this: https://github.com/moby/moby/issues/32579#issuecomment-788389016

krnowak:
Tests started to work, but only with the fixed emerge-gitclone (https://github.com/flatcar/flatcar-dev-util/pull/13 and https://github.com/flatcar/coreos-overlay/pull/2166).

pothos (approved):

PR #369, opened 2022-09-07
plume/release: publish on the AWS Marketplace

tormath1:
We use the us-east-1 AMIs ID to programmatically update the AWS Marketplace offer.

CI (or human) must provide the product-id and the access role ARN to update the offer. See also the documentation for the `ore aws update-offer` subcommand.

<hr>

Not tested - we can try it during the next release. It's designed to not fail if the update of the offer fails.

Related to: https://github.com/flatcar-linux/Flatcar/issues/838
Update of the offer itself: https://github.com/flatcar-linux/mantle/pull/282

### DIFF
--- a/cmd/plume/release.go
+++ b/cmd/plume/release.go
@@ -53,6 +53,12 @@ func init() {
 	cmdRelease.Flags().StringVar(&gceReleaseKey, "gce-release-key", "", "GCE key file for releases")
 	cmdRelease.Flags().BoolVarP(&releaseDryRun, "dry-run", "n", false,
 		"perform a trial run, do not make changes")
+	cmdRelease.Flags().BoolVarP(&publishMarketplace, "publish-marketplace", "", false,
+		"publish on the AWS marketplace")
+	cmdRelease.Flags().StringVar(&accessRoleARN, "access-role-arn", "", "ARN to give marketplace access to the AMI")
+	cmdRelease.Flags().StringVar(&productID, "product-id", "", "AWS Marketplace offer ID")
+	cmdRelease.Flags().StringVar(&awsMarketplaceCredentialsFile, "aws-marketplace-credentials", "", "AWS Marketplace credentials file")
+	cmdRelease.Flags().StringVar(&username, "username", "core", "default username")
 	AddSpecFlags(cmdRelease.Flags())
 	root.AddCommand(cmdRelease)
 }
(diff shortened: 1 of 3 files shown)


cmd/plume/release.go
```diff
@@ -450,6 +456,31 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+					}
+
+					// Define the launch instance type based on the arch.
+					instanceType := "t3.medium"
+					if specBoard == "arm64-usr" {
+						instanceType = "m6g.medium"
+					}
+
+					if err := marketplace.UpdateProduct(imageID, accessRoleARN, username, specVersion, productID, instanceType, releaseDryRun); err != nil {
+						plog.Warningf("updating product with ID %s: %v", productID, err)
```
pothos, line 480: I doubt this warning is seen. I think we should make it a hard error

cmd/plume/release.go
```diff
@@ -450,6 +456,31 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+				// Publish on AWS Marketplace AMIs in us-east-1.
+				if publishMarketplace && region == "us-east-1" {
+					// Create a new API client to consume the AWS Marketplace credentials.
+					marketplace, err := aws.New(&aws.Options{
+						CredentialsFile: awsMarketplaceCredentialsFile,
+						Profile:         "default",
+						Region:          "us-east-1",
+					})
+					if err != nil {
+						plog.Warningf("creating API Marketplace client: %v", err)
```
pothos, line 469: same

pothos (approved):
I think we should merge it because we need to port the AMI creation anyway in a separate PR to be able to test/use it

tormath1:
@pothos what do you mean by "we need to port the AMI creation anyway in a separate PR" ?

pothos:
For testing/use with the new pipeline we need to consume from bincache instead of GCS.

tormath1:
> For testing/use with the new pipeline we need to consume from bincache instead of GCS.

Oh got it. All this part: https://github.com/flatcar-linux/mantle/blob/0f48f1de9dec9c663a909a62c870d3748c8cf571/cmd/plume/release.go#L78 that uses GCS. I'll merge this and send a follow-up PR then. Thanks for the review!

pothos:
Yes, we could either try to let if behave like an anonymous bucket access by using the bincache HTTPS URL, or we rewrite it more to download from regular HTTP. Another option is that we don't download anything at all but just work on local files, and the release script would have to take care of the download from bincache.

PR #371, opened 2022-09-09
plume: update GCS URL to bincache URL

tormath1:
* plume: remove BasePrivateURL
it's no more used.

* plume/release: update GCS URL to bincache URL

<hr>

I was expecting more complexity to update this URL but it seems that `storage` is able to understand both type of URL, it converts GCS URL to plain HTTP url: 
https://github.com/flatcar-linux/mantle/blob/fb7bbf6e2c863c0bcfb9dcba4c090e45eb8c4742/cmd/plume/containerlinux.go#L326-L328

### DIFF
--- a/cmd/plume/containerlinux.go
+++ b/cmd/plume/containerlinux.go
@@ -102,7 +102,7 @@ var (
 
 	specs = map[string]channelSpec{
 		"alpha": channelSpec{
-			BaseURL:      "gs://flatcar-jenkins/alpha/boards",
+			BaseURL:      "http://bincache.flatcar-linux.net/images",
 			Boards:       []string{"amd64-usr", "arm64-usr"},
 			Destinations: []storageSpec{},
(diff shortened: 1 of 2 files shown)


pothos:
Thanks, one last missing piece is to split the ami txt/json file creation out of `awsUploadAmiLists` in a separate function which we then always call. There we could store them as local files. (`awsUploadAmiLists` would still be skipped when no GCS creds are given; the signing and copy to bincache is better done by the ci-automation bash code).

pothos (approved):

pothos:
> Thanks, one last missing piece is to split the ami txt/json file creation out of `awsUploadAmiLists` in a separate function which we then always call. There we could store them as local files. (`awsUploadAmiLists` would still be skipped when no GCS creds are given; the signing and copy to bincache is better done by the ci-automation bash code).

Done in https://github.com/flatcar-linux/mantle/pull/373

PR #374, opened 2022-09-14
docs: Replace flatcar-linux with flatcar

krnowak:
Also, cork is deprecated, so point to the correct location for documentation about working with SDK.

This is just fixing documentation files, so no code changes here.

### DIFF
--- a/platforms.md
+++ b/platforms.md
@@ -7,7 +7,7 @@ create images, collect logging information, etc.
 
 Authentication differs based on the platform. Some platforms like `aws` utilize the
 configuration files from their command-line tooling while others define their
-own custom configuration format and default locations (like [DigitalOcean](https://github.com/flatcar-linux/mantle/tree/master/auth/do.go)).
+own custom configuration format and default locations (like [DigitalOcean](https://github.com/flatcar/mantle/tree/master/auth/do.go)).
 Generally if any extensions / custom configurations are needed a new file is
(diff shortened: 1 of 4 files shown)


platforms.md
```diff
@@ -79,7 +79,7 @@ create images from the image file.
  - The DO platform wraps [godo](https://github.com/digitalocean/godo).
  - By default SSH keys will be passed via both the DO metadata AND the userdata.
  - UserData is passed to the instances via the DO metadata service.
- - DigitalOcean has no method for uploading custom images, as a result the `ore do create-image` command does a [~~terrifying~~ special workaround](https://github.com/flatcar-linux/mantle/blob/master/cmd/ore/do/create-image.go#L117-L173) which specifies custom userdata that does the following (after which the machine is snapshotted):
+ - DigitalOcean has no method for uploading custom images, as a result the `ore do create-image` command does a [~~terrifying~~ special workaround](https://github.com/flatcar/mantle/blob/master/cmd/ore/do/create-image.go#L117-L173) which specifies custom userdata that does the following (after which the machine is snapshotted):
    - configure networking in the initramfs
```
pothos, line 83: This part is not true anymore, as far as I remember

krnowak: Dunno, these docs would need a review anyway, but there is a lot of more important things to work on…

PR #379, opened 2022-09-21
plume/release: upload GCE image on GCS

tormath1:
The web URL of the GCE image (like: http://bincache.flatcar-linux.net/.../flatcar_production_gce.tar.gz) can't be directly used to create the image on GCP.

It's required to temporary upload the image on GCS to get a "https://storage.googleapis.com" link which is __required__ by the API to create an image on GCP.

### DIFF
--- a/cmd/plume/release.go
+++ b/cmd/plume/release.go
@@ -232,6 +232,12 @@ func doGCE(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
 	}
 
 	name := fmt.Sprintf("%s-%s", spec.GCE.Family, sanitizeVersion())
+
+	// We extend the name with '-arm64' suffix to avoid conflicting image name.
+	if specBoard == "arm64-usr" {
+		name = name + "-arm64"
+	}
+
 	date := time.Now().UTC()
 	desc := fmt.Sprintf("%s, %s, %s published on %s", spec.GCE.Description,
 		specVersion, specBoard, date.Format("2006-01-02"))
--- a/cmd/plume/release.go
+++ b/cmd/plume/release.go
@@ -17,6 +17,8 @@ package main
 import (
 	"fmt"
 	"net/http"
+	"net/url"
+	"os"
 	"sort"
 	"strings"
 	"time"
@@ -29,6 +31,7 @@ import (
 	"github.com/flatcar/mantle/platform/api/aws"
 	"github.com/flatcar/mantle/platform/api/azure"
 	"github.com/flatcar/mantle/platform/api/gcloud"
+	"github.com/flatcar/mantle/sdk"
 	"github.com/flatcar/mantle/storage"
 	"github.com/flatcar/mantle/storage/index"
 )
@@ -97,8 +100,16 @@ func runCLRelease(cmd *cobra.Command, args []string) error {
 		plog.Fatalf("File not found: %s", verurl)
 	}
 
-	// Register GCE image if needed.
-	doGCE(ctx, client, src, &spec)
+	// We do not provide yet ARM64 image for Google.
+	if specBoard == "amd64-usr" {
+		// Create a GCS bucket client to temporary upload the GCE image on GCS.
+		gcs, err := storage.NewBucket(client, "gs://flatcar-jenkins")
+		if err != nil {
+			plog.Fatalf("creating GCE bucket client: %v", err)
+		}
+
+		doGCE(ctx, client, gcs, &spec)
+	}
 
 	// Make Azure images public.
 	doAzure(ctx, client, src, &spec)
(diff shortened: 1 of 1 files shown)


cmd/plume/release.go
```diff
@@ -275,16 +282,36 @@ func doGCE(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
 			gceWaitForImage(pending)
```
pothos, line 282: If I understand this correctly, it means that we would only upload the image once, but on a rerun the old image stays. For AWS we delete the old one and upload the new image.

cmd/plume/release.go
```diff
@@ -275,16 +282,36 @@ func doGCE(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+
+		f, err := os.Open(spec.GCE.Image)
+		if err != nil {
+			plog.Fatalf("opening GCE image: %v", err)
+		}
+
+		defer f.Close()
+
+		o := gs.Object{Name: src.Prefix() + spec.GCE.Image}
+		if err := src.Upload(ctx, &o, f); err != nil {
```
pothos, line 306: This this overwrite an existing object?

tormath1: Good question - it seems that default behavior is to leave early if there is an existing object matching a checksum.
https://github.com/flatcar/mantle/blob/a40798269e6c621d2f6fe4a1af27c6bd921ad72c/storage/bucket.go#L276-L277

We could do something like that:
```suggestion
                src.WriteAlways(true)
		if err := src.Upload(ctx, &o, f); err != nil {
```

cmd/plume/release.go
tormath1: Image is deleted _after_, I guess with such a behavior + image overwriting, we should be safe to rerun.

pothos: In `len(conflicting) == 1` we can call `api.DeleteImage(image.Name)` and remove the rest of the if branch

pothos: and then we execute the below unconditionally, not in an `else`

cmd/plume/release.go
pothos: Seems yes

cmd/plume/release.go
```diff
@@ -346,6 +373,10 @@ func doGCE(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
 			plog.Fatal(err)
 		}
 	}
+
+	if err := src.Delete(ctx, src.Prefix() + spec.GCE.Image); err != nil {
+		plog.Fatalf("deleting GCE image from GCS: %v", err)
```
pothos, line 378: Thinking about it, maybe we have to keep it around if direct deletion is not allowed. If we are able to remove the image once published in the marketplace, we could do so in a GC job, which could also delete the GCS object then.

cmd/plume/release.go
pothos: This is about the image, at the moment we delete the object at the end while the image stays.

cmd/plume/release.go
pothos: Good, makes sense to set WriteAlways

cmd/plume/release.go
pothos: Something seems to be implemented directly here, worth checking if it deletes image plus object:
```
18:25:46  2022-09-21T16:25:46Z plume: Deprecating old image flatcar-alpha-3346-0-0
18:25:47  2022-09-21T16:25:46Z plume: Pruning 1 GCE images.
18:25:47  2022-09-21T16:25:46Z plume: Deleting old image flatcar-alpha-3127-0-0
```

pothos: It runs `api.DeleteImage(old.Name)` but it doesn't delete the object yet.

pothos: This is in `329:		plog.Noticef("Pruning %d GCE images.", len(oldImages)-spec.GCE.Limit)`

tormath1: Ok, so we need to get the image, access the image raw disk and get the storage link to delete the object

cmd/plume/release.go
pothos: Solved by overwriting

cmd/plume/release.go
pothos: Or we could use the calculated GCS storage object that we would have uploaded

tormath1: Not sure because the calculated object is based on the current release - for the image `flatcar-alpha-3127-0-0` it would be a different one.

pothos: Yes, just using the same scheme for the path with the right version, but your idea is cleaner of course. We could also do a manual cleanup of old GCS objects at some point.

pothos (approved):

PR #381, opened 2022-09-23
kola/tests/devcontainer: Fallback to bzcat

krnowak:
Older images may not have lbzcat, so fall back to bzcat.


krnowak:
It's here to fix the issue that popped up in jenkins run: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/2901/

CI passed: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/2905/

tormath1 (approved):

PR #382, opened 2022-09-23
openstack/api: wait for image to be active

tormath1:
the image usually takes a few second to be unpacked on the Openstack instance.

It can leads to CI error if the image used to run the test it not yet fully active.

### DIFF
--- a/platform/api/openstack/api.go
+++ b/platform/api/openstack/api.go
@@ -546,6 +546,18 @@ func (a *API) UploadImage(name, path string) (string, error) {
 			return "", fmt.Errorf("web uploading: %w", err)
 		}
 
+		// It usually takes around 10 seconds to extract the image.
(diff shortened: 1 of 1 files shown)


platform/api/openstack/api.go
```diff
@@ -546,6 +546,18 @@ func (a *API) UploadImage(name, path string) (string, error) {
 			return "", fmt.Errorf("web uploading: %w", err)
 		}
 
+		// It usually takes around 10 seconds to extract the image.
+		if err := util.WaitUntilReady(20*time.Second, 5*time.Second, func() (bool, error) {
```
krnowak, line 550: I'd be on the safe side and would wait for a minute. :)

tormath1: Ok, let's do this!

krnowak (approved):

PR #385, opened 2022-10-10
kola: Ignore dead links in /usr/share/flatcar/etc

krnowak:
The relative dead symlinks there are copied verbatim by tmpfiles to `/etc`, where they should stop being dead. This test is going to go through those files in `/etc`, which we can treat as testing symlinks in `/usr/share/flatcar/etc`.


pothos:
There is also https://github.com/flatcar/mantle/pull/316

krnowak:
Heh, I even approved it. :) Alright, will close this one later, now the docker container is building so I'll keep it for testing.

pothos:
I merged the other one, this here also has the typo `flatcat`

PR #391, opened 2022-10-28
plume: Use plain AMI image for uploads

pothos:
Recently we ran into sporadic corruption issues for AWS EC2 AMIs. We use the streamOptimized VMDK format and it seems to cause problems at the AWS side, regardless if created by qemu-img or vmdk-convert. Switch to using the plain AMI images for uploading as workaround.

## How to use

When merged and the scripts repo has new mantle ref update PRs, merge them and retag the releases.

## Testing done

Will be tested as part of the release.

### DIFF
--- a/cmd/plume/containerlinux.go
+++ b/cmd/plume/containerlinux.go
@@ -191,7 +191,7 @@ func newAWSSpec() awsSpec {
 		BaseName:        "Flatcar",
 		BaseDescription: "Flatcar Container Linux",
 		Prefix:          "flatcar_production_ami_",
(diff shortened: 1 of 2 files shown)


dongsupark (approved):
Looks good.
Then should we probably update file names in https://github.com/flatcar/mantle/tree/flatcar-master/cmd/ore/aws?

pothos:
> Then should we probably update file names in https://github.com/flatcar/mantle/tree/flatcar-master/cmd/ore/aws?

I'm not sure if it's good to change the default value of the `object-format` command line parameter because that's an incompatible change. In https://github.com/flatcar/scripts/pull/541 I've passed it explicitly.

PR #395, opened 2022-11-11
kola: Add custom Docker sysext test

pothos:
This tests that users are able to supply a custom Docker version as sysext. It uses create_docker_sysext.sh which is documented in https://www.flatcar.org/docs/latest/provisioning/sysext/

## How to use

Replace branch checkout by git commit ID when https://github.com/flatcar/sysext-bakery/pull/6 is merged

## Testing done

`sudo ./kola run -d -k --keys --key ~/.ssh/id_rsa.pub  --board=amd64-usr --parallel=1 --platform=qemu  --qemu-image=flatcar_production_image.bin  --qemu-skip-mangle systemd.sysext.custom-docker`

### DIFF
--- a/kola/tests/systemd/sysext.go
+++ b/kola/tests/systemd/sysext.go
@@ -31,6 +31,22 @@ func init() {
         inline: |
           sysext works`),
 	})
+	register.Register(&register.Test{
+		Name:        "systemd.sysext.custom-docker",
+		Run:         checkSysextCustomDocker,
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+		// This test is normally not related to the cloud environment
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 3185},
+		UserData: conf.ContainerLinuxConfig(`storage:
+  files:
+    - path: /etc/systemd/system-generators/torcx-generator
+  directories:
+    - path: /etc/extensions/docker-flatcar
+    - path: /etc/extensions/containerd-flatcar`),
+	})
+
 }
 
 func checkHelper(c cluster.TestCluster) {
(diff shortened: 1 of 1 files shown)


kola/tests/systemd/sysext.go
```diff
@@ -50,3 +66,17 @@ func checkSysextSimple(c cluster.TestCluster) {
...
+
+func checkSysextCustomDocker(c cluster.TestCluster) {
+	cmdNotWorking := `if docker run --rm ghcr.io/kinvolk/busybox true; then exit 1; fi`
+	cmdWorking := `docker run --rm ghcr.io/kinvolk/busybox echo Hello World`
+	// First assert that Docker doesn't work because Torcx is disabled
+	_ = c.MustSSH(c.Machines()[0], cmdNotWorking)
+	// We build a custom sysext image locally because we don't host them somewhere yet (TODO: use "main" branch)
+	_ = c.MustSSH(c.Machines()[0], `git clone https://github.com/flatcar/sysext-bakery.git && git -C sysext-bakery checkout kai/containerd`)
+	// Flatcar has no mksquashfs and btrfs is missing a bugfix but at least ext4 works
+	_ = c.MustSSH(c.Machines()[0], `ONLY_DOCKER=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh 20.10.21 docker && ONLY_CONTAINERD=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh 20.10.21 containerd && sudo mv docker.raw containerd.raw /etc/extensions/`)
```
krnowak, line 78: How about avoiding the hardcoding of the version in favor of something like `curl -s https://api.github.com/repos/moby/moby/releases/latest | jq --raw-output '.tag_name' | tail --bytes +2`?

pothos: The idea of using a particular kola version referenced by `scripts` was to be reproducible. I would also use a git commit ID in the git check out for the same reason.

kola/tests/systemd/sysext.go
```diff
@@ -50,3 +66,17 @@ func checkSysextSimple(c cluster.TestCluster) {
 	// Second check after reloading the extensions (e.g., to add/remove/update them)
 	checkHelper(c)
 }
+
+func checkSysextCustomDocker(c cluster.TestCluster) {
+	cmdNotWorking := `if docker run --rm ghcr.io/kinvolk/busybox true; then exit 1; fi`
+	cmdWorking := `docker run --rm ghcr.io/kinvolk/busybox echo Hello World`
+	// First assert that Docker doesn't work because Torcx is disabled
+	_ = c.MustSSH(c.Machines()[0], cmdNotWorking)
+	// We build a custom sysext image locally because we don't host them somewhere yet (TODO: use "main" branch)
```
pothos, line 75:
```suggestion
	// We build a custom sysext image locally because we don't host them somewhere yet (TODO: use git ID)
```

kola/tests/systemd/sysext.go
krnowak: I meant the docker version that is passed to the `create_docker_sysext.sh` script, not kola. Or maybe we could take the version from the docker ebuild in coreos-overlay, but not sure how to do it here.

pothos: Yes, if the docker version is dynamic we lose the benefit of pinning kola tests for reproducibility.

krnowak: Alright, let's get this in for now. But would be great to eventually use a docker version that matches the one we package for tested flatcar version. That should be still relatively reproducible. Maybe something like:

```bash
docker_version=$(curl -fsSL https://bincache.flatcar-linux.net/images/${ARCH}/${VERSION}/flatcar_production_image_packages.txt | grep 'app-emulation/docker-[0-9]' | cut -d- -f3 | cut -d: -f1)
```

Otherwise we will need another github action that bumps the docker version from time to time to use something recent.

krnowak (approved):

kola/tests/systemd/sysext.go
pothos: Good idea to bind it to the release.
While it makes sense to test recent versions it's also meaningful to test old versions because when someone wants to use a custom Docker version (and therefore, uses sysext), it's likely that they don't update it and stick to the old version.

pothos: For binding it to the Docker version inside the image we could get the version from the license json file under `/usr/share/flatcar/`.

pothos: I'll update the test to use both and then we can explain why we want to also use an old version.

kola/tests/systemd/sysext.go
```diff
@@ -50,3 +66,24 @@ func checkSysextSimple(c cluster.TestCluster) {
...
+	// The first test is for a fixed Docker version, which with the time will get old and older but is still expected to work because users may also "freeze" their Docker version this way
+	_ = c.MustSSH(c.Machines()[0], `ONLY_DOCKER=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh 20.10.21 docker && ONLY_CONTAINERD=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh 20.10.21 containerd && sudo mv docker.raw containerd.raw /etc/extensions/`)
+	_ = c.MustSSH(c.Machines()[0], `sudo systemctl restart systemd-sysext`)
+	// We should now be able to use Docker
+	_ = c.MustSSH(c.Machines()[0], cmdWorking)
+	// The next test is with a recent Docker version, here the one from the Flatcar image to couple it to something that doesn't change under our feet
+	version := string(c.MustSSH(c.Machines()[0], `bzcat /usr/share/licenses/licenses.json.bz2 | grep -m 1 -o 'app-emulation/docker[^:]*' | cut -d - -f 3`))
+	_ = c.MustSSH(c.Machines()[0], `ONLY_DOCKER=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh `+version+` docker && ONLY_CONTAINERD=1 FORMAT=ext4 sysext-bakery/create_docker_sysext.sh `+version+` containerd && sudo mv docker.raw containerd.raw /etc/extensions/`)
```
tormath1, line 85: I think we can set the [`ARCH`](https://github.com/flatcar/sysext-bakery/blob/main/create_docker_sysext.sh#L4) argument. Otherwise it fails on arm64:
```
cluster.go:117: bash: line 1: /usr/bin/docker: cannot execute binary file: Exec format error
```

PR #399, opened 2022-12-02
Dockerfile: Add azure-cli for image gallery publishing

pothos:
While the mantle container has plume/ore for publishing cloud images, not all actions are implemented in plume or ore. We already have the awscli for some publishing steps but the "az" tool from azure-cli was missing and is useful for image gallery publishing.

## How to use

From https://github.com/flatcar/scripts/blob/380f9723290932e04f3a057a1a0cf04e7b56a93b/ci-automation/release.sh#L150

## Testing done

`docker run --rm -it ghcr.io/flatcar/mantle:pr-399` and then `az`


t-lo (approved):
LGTM, thank you Kai - that was lightning-fast!

PR #403, opened 2023-01-10
platform/azure: fetch marketplace image plan info from endpoint

jepio:
# platform/azure: fetch marketplace image plan info from endpoint

Fetch Azure marketplace plan information dynamically, so that `kola spawn` works for the Arm64 CoreVM offer.

## How to use

## Testing done

```
./bin/kola spawn --platform=azure --azure-location eastus2 --azure-publisher kinvolk --azure-offer flatcar-container-linux-corevm --azure-sku stable --azure-size "Standard_D2pls_v5" --azure-hyper-v-generation V2 --azure-auth ... --azure-profile ...
```

### DIFF
--- a/platform/api/azure/instance.go
+++ b/platform/api/azure/instance.go
@@ -71,10 +71,17 @@ func (a *API) getVMParameters(name, userdata, sshkey, storageAccountURI string, 
 			Sku:       &a.Opts.Sku,
 			Version:   &a.Opts.Version,
 		}
-		plan = &compute.Plan{
-			Publisher: imgRef.Publisher,
-			Product:   imgRef.Offer,
-			Name:      imgRef.Sku,
+		// lookup plan information for image
+		imgInfo, err := a.vmImgClient.Get(context.TODO(), a.Opts.Location, *imgRef.Publisher, *imgRef.Offer, *imgRef.Sku, *imgRef.Version)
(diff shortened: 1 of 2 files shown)


platform/api/azure/instance.go
```diff
@@ -71,10 +71,17 @@ func (a *API) getVMParameters(name, userdata, sshkey, storageAccountURI string,
...
+		imgInfo, err := a.vmImgClient.Get(context.TODO(), a.Opts.Location, *imgRef.Publisher, *imgRef.Offer, *imgRef.Sku, *imgRef.Version)
+		if err == nil && imgInfo.Plan != nil {
+			plan = &compute.Plan{
+				Publisher: imgInfo.Plan.Publisher,
+				Product:   imgInfo.Plan.Product,
+				Name:      imgInfo.Plan.Name,
+			}
+			plog.Debugf("using plan: %v:%v:%v", *plan.Publisher, *plan.Product, *plan.Name)
+		} else if err != nil {
+			plog.Warningf("failed to get image info: %v; continuing", err)
```
krnowak, line 84: What about the case, where `err == nil && imgInfo.Plan == nil`?

jepio: That's exactly what happens for the corevm offering. The plan info field needs to be null in that case.

krnowak (approved):

PR #405, opened 2023-01-11
kola: Add new cl.sysext.boot test for selection of active OEM sysext

pothos:
For OEM sysext images and optional Flatcar extensions we need on-boot selection of the active sysext image and possible moving between the rootfs and the OEM partition. This is introduced in https://github.com/flatcar/bootengine/pull/50
Add a new test to check that the boot logic works as expected by preparing a QEMU machine to cover the relevant conditions. Since the systemd units in the initrd and the mounting is different depending on whether Ignition runs and itself interacts with the OEM partition, we have two machines, one without Ignition and one with because we anyway need to aquire the full version string first to be able to create the Ignition config with the right entries.

## How to use

Only merge if https://github.com/flatcar/bootengine/pull/50 is merged and also adjust the major version number before merging to be 1 larger than that of the latest Alpha branch at that point

## Testing done

Local test run with http://bincache.flatcar-linux.net/images/amd64/9999.0.0+kai-sysext-oem-bootengine/flatcar_production_image.bin.bz2

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -29,6 +29,7 @@ import (
 	"github.com/flatcar/mantle/kola/register"
 	tutil "github.com/flatcar/mantle/kola/tests/util"
 	"github.com/flatcar/mantle/platform"
+	"github.com/flatcar/mantle/platform/conf"
 	"github.com/flatcar/mantle/platform/local"
 	"github.com/flatcar/mantle/util"
 )
@@ -51,6 +52,15 @@ func init() {
 			// (see scripts/ci-automation/vendor-testing/qemu_update.sh)
 			return kola.UpdatePayloadFile == ""
 		},
+	})
+	register.Register(&register.Test{
+		Name:        "cl.sysext.boot",
+		Run:         sysextBootLogic,
+		ClusterSize: 0,
+		Distros:     []string{"cl"},
+		// This test is uses its own OEM files and shouldn't run on other platforms
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 3481},
 	})
 }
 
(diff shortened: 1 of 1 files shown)


kola/tests/update/update.go
```diff
@@ -211,3 +222,86 @@ func splitNewlineEnv(envs string) map[string]string {
 	}
 	return m
 }
+
+func sysextBootLogic(c cluster.TestCluster) {
```
krnowak, line 226: Could you split the `noIgnition` and `withIgnition` into separate subtests? I think you could do it like:
```go
c.Run("no ignition", func(c cluster.TestCluster) { … }))
c.Run("with ignition", func(c cluster.TestCluster) { … }))
```

kola/tests/update/update.go
```diff
@@ -211,3 +222,86 @@ func splitNewlineEnv(envs string) map[string]string {
...
+	// We set up symlinks to emulate that a previous sysext was active and we store the new sysext
+	// in the rootfs instead of the OEM partition. The previous sysext is either
+	// a) stored on the rootfs and will stay there and the new one is is moved to the OEM partition
+	// b) stored on the OEM partition and gets moved to the rootfs and the new one is moved to the OEM partition
+	_ = c.MustSSH(noIgnition, fmt.Sprintf(`set -euxo pipefail
+sudo systemctl mask --now systemd-sysext
+sudo mkdir -p /etc/flatcar/sysext /etc/flatcar/oem-sysext /usr/share/oem/sysext /etc/extensions
+echo ID=test | sudo tee /usr/share/oem/oem-release
+echo myext | sudo tee /etc/flatcar/enabled-sysext.conf
+sudo touch /usr/share/oem/sysext/active-oem-test /etc/flatcar/oem-sysext/oem-test-%[1]s.raw /etc/flatcar/oem-sysext/oem-test-1.2.3.raw /etc/flatcar/sysext/flatcar-myext-%[1]s.raw
```
krnowak, line 249: Shouldn't there also be a file at `/etc/flatcar/sysext/flatcar-myext-1.2.3.raw` and a symlink to it at `/etc/extensions/flatcar-myext.raw`, so we can check below if those get replaced too?

kola/tests/update/update.go
```diff
@@ -211,3 +222,86 @@ func splitNewlineEnv(envs string) map[string]string {
...
+sudo mv /etc/flatcar/oem-sysext/oem-test-1.2.3.raw /usr/share/oem/sysext/
+sudo ln -fs /usr/share/oem/sysext/oem-test-1.2.3.raw /etc/extensions/oem-test.raw
+`, version))
+	if err := noIgnition.Reboot(); err != nil {
+		c.Fatalf("couldn't reboot: %v", err)
+	}
+	// Check that the boot logic set up the right sysext symlinks for case "b)"
+	testCmds := fmt.Sprintf(`set -euxo pipefail
+[ "$(readlink -f /etc/extensions/oem-test.raw)" = "/usr/share/oem/sysext/oem-test-%[1]s.raw" ] || { echo "OEM symlink wrong"; exit 1 ; }
+[ "$(readlink -f /etc/extensions/flatcar-myext.raw)" = "/etc/flatcar/sysext/flatcar-myext-%[1]s.raw" ] || { echo "Extension symlink wrong"; exit 1; }
```
krnowak, line 269: Should there be a test if the old oem sysext was indeed moved to rootfs?

kola/tests/update/update.go
pothos: Not sure, the `version` variable resolved in one is used in the second

kola/tests/update/update.go
pothos: Thanks!

kola/tests/update/update.go
pothos: Thanks!

kola/tests/update/update.go
```diff
@@ -211,3 +222,86 @@ func splitNewlineEnv(envs string) map[string]string {
...
+systemd:
+  units:
+    - name: systemd-sysext.service
+      mask: true
+`, version))
+	withIgnition, err := c.NewMachine(conf)
+	if err != nil {
+		c.Fatalf("creating test machine: %v", err)
+	}
+	_ = c.MustSSH(withIgnition, testCmds)
```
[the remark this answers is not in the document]
pothos: Yes, the Ignition case is different. I didn't want to do the same tests again each time with new VMs because the logic hit is the same, the main difference is that with Ignition the OEM mount point is set up a bit differently and we also just want to check that the logic is triggered at all because another systemd target unit is used to pull it in.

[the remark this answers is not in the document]
pothos: I've added a longer comment on this

kola/tests/update/update.go
krnowak: Ok, let's keep it as is then.

kola/tests/update/update.go
```diff
@@ -215,3 +225,96 @@ func splitNewlineEnv(envs string) map[string]string {
...
+	// a) stored on the rootfs and will stay there and the new one is is moved to the OEM partition
+	// b) stored on the OEM partition and gets moved to the rootfs and the new one is moved to the OEM partition
+	_ = c.MustSSH(noIgnition, fmt.Sprintf(`set -euxo pipefail
+sudo systemctl mask --now systemd-sysext
+sudo mkdir -p /etc/flatcar/sysext /etc/flatcar/oem-sysext /usr/share/oem/sysext /etc/extensions
+echo ID=test | sudo tee /usr/share/oem/oem-release
+echo myext | sudo tee /etc/flatcar/enabled-sysext.conf
+sudo touch /usr/share/oem/sysext/active-oem-test /etc/flatcar/oem-sysext/oem-test-%[1]s.raw /etc/flatcar/oem-sysext/oem-test-1.2.3.raw /etc/flatcar/sysext/flatcar-myext-%[1]s.raw
+sudo ln -fs /etc/flatcar/oem-sysext/oem-test-1.2.3.raw /etc/extensions/oem-test.raw
+sudo ln -fs /etc/flatcar/sysext/flatcar-myext-1.2.3.raw /etc/extensions/flatcar-myext.raw
```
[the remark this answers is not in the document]
pothos: Thanks! The dangling symlink case is not really interesting, the symlink gets recreated or deleted

kola/tests/update/update.go
```diff
@@ -52,6 +53,15 @@ func init() {
...
 		},
 	})
+	register.Register(&register.Test{
+		Name:        "cl.sysext.boot",
+		Run:         sysextBootLogic,
+		ClusterSize: 0,
+		Distros:     []string{"cl"},
+		// This test is uses its own OEM files and shouldn't run on other platforms
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 9999},
```
krnowak, line 63: Ha ha ha. Alright, I guess it will be fine to merge this now and update the min version when changes in bootengine are merged.

krnowak (approved):

PR #406, opened 2023-01-12
tests/kubeadm: add kubernetes 1.26

tormath1:
Kubernetes 1.26.0 is out since beginning of december. Let's cover it in the tests.

<hr>

Notified while testing CAPO that 1.26 is not being tested (https://github.com/flatcar/Flatcar/issues/921)

Tested on openstack with current Flatcar stable (3374.2.2):
```
cat _kola_temp/openstack-latest/test.tap
1..3
ok - kubeadm.v1.26.0.flannel.base
ok - kubeadm.v1.26.0.calico.base
ok - kubeadm.v1.26.0.cilium.base
```

Regarding 1.23: active support has been dropped 2 weeks ago and maintenance support will be dropped in 1 month. I think it's good to have it until end of Feb. 

Thanks @krnowak for the `checksums.sh` it's quite handy, generated this PR in a minute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Kola test for devcontainer ([#367](https://github.com/flatcar-linux/mantle/pull/367))
 - `--qemu-grow-base-disk-by` flag to grow the base disk ([#367](https://github.com/flatcar-linux/mantle/pull/367))
 - `--force-flatcar-key` flag to force injecting the Flatcar production key when testing the Flatcar update with `cl.update.payload` ([#398](https://github.com/flatcar/mantle/pull/398))
+- kubernetes 1.26.0 tests ([#406](https://github.com/flatcar/mantle/pull/406))
 
 ### Changed
 - organization renaming (`flatcar-linux` -> `flatcar`) ([#372](https://github.com/flatcar/mantle/pull/372), [#374](https://github.com/flatcar/mantle/pull/374))


kola/tests/kubeadm/kubeadm.go
```diff
@@ -75,6 +75,38 @@ var (
 	// testConfig holds params for various kubernetes releases
 	// and the nested params are used to render script templates
 	testConfig = map[string]map[string]interface{}{
+		"v1.26.0": map[string]interface{}{
+			"MinMajorVersion": 3033,
+			// from https://github.com/flannel-io/flannel/releases
+			"FlannelVersion": "v0.19.1",
```
krnowak, line 81: I don't remember if we agreed on any policy with regards to bumping these version numbers. Have we? I'm asking because looking into the URLs in comments, I can see that flannel has version v0.20.2, cilium has 1.12.5, cilium cli has 0.12.12, cni is still v1.1.1, crictl has v1.26.0 and k8s release is still v0.14.0.

So not sure, maybe a follow-up PR if these versions work for k8s 1.26?

(Also, happy that the checksums script worked for you!)

krnowak: My take on these tests is that we want to test a range of k8s versions on flatcar. Since k8s needs some networking solution and these tend to be quite dependent on running kernel, we also want to test a range of networking solutions versions. So it may be a good idea to bump the versions of all this stuff at the time of a new major k8s release. It won't give 100% coverage, but should be good enough. This is in light of a realization that having data on the most used version combinations of software is a wishful thinking, not gonna happen. :)

krnowak (approved):

PR #410, opened 2023-02-06
Update some journal strings for systemd 252

krnowak:
Systemd is at it again, changing some strings. Generalize the string search a bit and add new strings.

### DIFF
--- a/kola/tests/misc/network.go
+++ b/kola/tests/misc/network.go
@@ -17,6 +17,7 @@ package misc
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
@@ -194,6 +195,11 @@ func NetworkListeners(c cluster.TestCluster) {
 	}
 }
 
+func findString(strs []string, s string) bool {
+	idx := sort.SearchStrings(strs, s)
+	return idx < len(strs) && strs[idx] == s
+}
+
 // Verify that networking is not started in the initramfs on the second boot.
 // https://github.com/coreos/bugs/issues/1768
 func NetworkInitramfsSecondBoot(c cluster.TestCluster) {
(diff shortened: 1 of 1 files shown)


kola/tests/misc/network.go
```diff
@@ -207,11 +213,17 @@ func NetworkInitramfsSecondBoot(c cluster.TestCluster) {
 
 	// verify that the network service was started
 	found := false
+	// In v249, some services description have been updated.
+	// More details: https://github.com/systemd/systemd/commit/4fd3fc66396026f81fd5b27746f2faf8a9a7b9ee
+	// In v252, the printed line stoped using a description of the unit, going to the unit name.
```
pothos, line 218: If we don't do so already we should also make sure to use `status-unit-format=combined` (can also be set as `StatusUnitFormat=combined` in `system.conf`, best through a drop in under `/usr`, I guess)

krnowak: Sure, can be done, but do you mean here to do it via a userdata in the test case or just in general, as a part of our production image? If the former, then I don't know how is this supposed to work, since putting a config snippet during ignition run probably won't do much, since systemd is already running. If the latter, then I'd do it as a follow up. Which will allow us to maybe simplify some tests in the long run, but not any time soon, as other channels still have a different configuration.

krnowak:
CI passed here: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/597/cldsv/

kola/tests/misc/network.go
pothos: In general, I think this makes sense but we would have to update the strings here again ;)

pothos (approved):

kola/tests/misc/network.go
krnowak: Added it to my TODO list.

pothos: The meson option `status-unit-format-default` can be set to `combined`, this is a bit better than doing it via a drop-in.

pothos: But if it's worth to minimize the ebuild changes, this here in the `init` repo should be good:
```
diff --git a/systemd/system.conf.d/10-status-combined.conf b/systemd/system.conf.d/10-status-combined.conf
new file mode 100644
index 0000000..64200c4
--- /dev/null
+++ b/systemd/system.conf.d/10-status-combined.conf
@@ -0,0 +1,2 @@
+[Manager]
+StatusUnitFormat=combined
```

krnowak: Yeah, I was thinking about using a drop-in too.

PR #414, opened 2023-02-24
kola/tests: Add test for /etc overlay upcopy prevention and cleanup

pothos:
The /etc overlayfs mount is there to be able to update the lowerdir file and this only works if we have no unwanted duplicates in the upperdir (i.e., the /etc folder).
Add tests to check against upcopies caused by, e.g., systemd-tmpfiles and to check that duplicates get removed on reboot (introduced in https://github.com/flatcar/coreos-overlay/pull/2482).

## How to use

Expected to pass for https://github.com/flatcar/coreos-overlay/pull/2482

## Testing done

[CI](http://jenkins.infra.kinvolk.io:8080/job/container/job/test/7932/consoleFull)
and local run of `cl.overlay.cleanup`

### DIFF
--- a/kola/tests/misc/state.go
+++ b/kola/tests/misc/state.go
@@ -1,0 +1,62 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package misc
+
+import (
+	"fmt"
+
+	"github.com/coreos/go-semver/semver"
+
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         OverlayCleanup,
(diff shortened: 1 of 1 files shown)


kola/tests/misc/state.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/flatcar/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         OverlayCleanup,
+		ClusterSize: 1,
+		Name:        "cl.overlay.cleanup",
+		Distros:     []string{"cl"},
+		MinVersion:  semver.Version{Major: 3511},
```
krnowak, line 21: Should the minversion be higher?

```suggestion
		MinVersion:  semver.Version{Major: 3530},
```

krnowak (approved):

PR #415, opened 2023-02-28
kola/tests: Add test for OS reset

pothos:
The new flatcar-reset tool and boot logic for selective OS reset should allow to discard all local state except wanted paths that are to be preserved.
Add a test that checks if the OS reset works as expected.

## How to use

With https://github.com/flatcar/coreos-overlay/pull/2483

## Testing done

Local run of `cl.osreset.ignition-rerun cl.overlay.cleanup`

### DIFF
--- a/kola/tests/misc/state.go
+++ b/kola/tests/misc/state.go
@@ -10,6 +10,30 @@ import (
 
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/platform/conf"
+)
+
+var (
+	ignitionRerun = conf.Butane(`---
+variant: flatcar
+version: 1.0.0
+storage:
+  files:
+  - path: /file-works
+    contents:
+      inline: "something"
+systemd:
+  units:
+    - name: test.service
+      enabled: true
+      contents: |
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=touch /service-works
+        [Install]
+        WantedBy=multi-user.target
+`)
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


kola/tests/misc/state.go
```diff
@@ -10,6 +10,30 @@ import (
 
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/platform/conf"
+)
+
+var (
+	ignitionRerun = conf.ContainerLinuxConfig(`
```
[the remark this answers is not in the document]
pothos: Thanks, good idea, updated it

kola/tests/misc/state.go
```diff
@@ -22,6 +46,16 @@ func init() {
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
 	})
+	register.Register(&register.Test{
+		Run:         OsReset,
+		ClusterSize: 1,
+		Name:        "cl.osreset.ignition-rerun",
+		UserData:    ignitionRerun,
+		Distros:     []string{"cl"},
+		MinVersion:  semver.Version{Major: 3511},
```
krnowak, line 55: Should the minversion be higher?

```suggestion
		MinVersion:  semver.Version{Major: 3530},
```

kola/tests/misc/state.go
```diff
@@ -60,3 +94,35 @@ func OverlayCleanup(c cluster.TestCluster) {
...
+func OsReset(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	ignitionCheck := `sudo systemctl start test && if ! systemctl is-enabled -q test.service || test ! -e /service-works || test ! -e /file-works; then echo "Missing service/file %s: $_"; exit 1; fi`
+	_ = c.MustSSH(m, fmt.Sprintf(ignitionCheck, "on initial boot"))
+
+	prevMachineId := string(c.MustSSH(m, `cat /etc/machine-id`))
+
+	// Create some local state to discard and to preserve, convering cases
+	// where a file in a folder should be preserved an another not, or where
```
krnowak, line 109:
```suggestion
	// Create some local state to discard and to preserve, covering cases
	// where a file in a folder should be preserved and another not, or where
```

krnowak (approved):

PR #418, opened 2023-03-07
sdk/verify: Update subkey

pothos:
The new subkey will be used from July.


dongsupark (approved):

pothos:
@tormath1 not sure if we need a new release for plume?

PR #419, opened 2023-03-15
kola: add http server

tormath1:
In this PR, we add a simple HTTP server that runs on the started instance to serve `/var/www` content on port `:8080`. It can be used for example to serve a dev container archive provided by `--devcontainer-file ./local-flatcar_developer_container.bin.bz2`

We could also think about using it to test sysext images? 

NOTE: Regarding the devcontainer test, we also add a new flag `--devcontainer-url` to provide a different base URL.

Closes: https://github.com/flatcar/Flatcar/issues/976

### DIFF
--- a/kola/tests/devcontainer/devcontainer.go
+++ b/kola/tests/devcontainer/devcontainer.go
@@ -250,7 +250,7 @@ func runDevContainerTest(c cluster.TestCluster, scriptBody string) {
 	}
 
 	scriptParameters := scriptTemplateParameters{
-		BinhostURLTemplate:        "http://bincache.flatcar-linux.net/boards/@ARCH@-usr/@VERSION@/pkgs",
+		BinhostURLTemplate:        kola.DevcontainerBinhostURL,
 		ImageDirectoryURLTemplate: devcontainerURL,
 	}
 
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -75,6 +75,7 @@ func init() {
 	sv(&kola.TorcxManifestFile, "torcx-manifest", "", "Path to a torcx manifest that should be made available to tests")
 	sv(&kola.DevcontainerURL, "devcontainer-url", "http://bincache.flatcar-linux.net/images/@ARCH@/@VERSION@", "URL to a dev container archive that should be made available to tests")
 	sv(&kola.DevcontainerFile, "devcontainer-file", "", "Path to a dev container archive that should be made available to tests as alternative to devcontainer-url, note that a working devcontainer-binhost-url is still needed")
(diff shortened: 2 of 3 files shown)


t-lo:
Tested locally with `qemu_uefi` and dev container image from the [latest nightly](https://bincache.flatcar-linux.net/images/amd64/3542.0.0+nightly-20230313-2100/). Both devcontainer tests pass:
```sh
$ sudo bin/kola run --board=amd64-usr --parallel=10 --platform=qemu --qemu-bios=flatcar_production_qemu_uefi_efi_code.fd --qemu-image=flatcar_production_image.bin --tapfile=result.tap --qemu-skip-mangle --devcontainer-file=flatcar_developer_container.bin.bz2 devcontainer.systemd-nspawn devcontainer.docker
=== RUN   devcontainer.docker
=== RUN   devcontainer.systemd-nspawn
--- PASS: devcontainer.systemd-nspawn (344.18s)
--- PASS: devcontainer.docker (415.75s)
PASS, output in _kola_temp/qemu-2023-03-15-1233-645068
$ cat result.tap
1..2
ok - devcontainer.systemd-nspawn
ok - devcontainer.docker
```

kola/tests/devcontainer/devcontainer.go
```diff
@@ -225,18 +228,32 @@ func init() {
...
-func runDevContainerTest(c cluster.TestCluster, scriptParameters scriptTemplateParameters, scriptBody string) {
+func runDevContainerTest(c cluster.TestCluster, scriptBody string) {
+	devcontainerURL := kola.DevcontainerURL
+	if kola.DevcontainerFile != "" {
+		// This URL is deterministic as it runs on the started machine.
+		devcontainerURL = "http://localhost:8080"
+	}
+
+	scriptParameters := scriptTemplateParameters{
+		BinhostURLTemplate:        "http://bincache.flatcar-linux.net/boards/@ARCH@-usr/@VERSION@/pkgs",
```
pothos, line 253: Do we also need to pass the `pkgs` folder to the instance? This bincache path only works for developer builds made on Jenkins but since this here is meant to be used in GitHub Actions, nothing will be there.

t-lo: Oh right, I did not even consider this because I used images from a nightly build for my testing above.
We would also need to export the packages in the Build CI job and fetch that artifact in the kola CI job. Making a metal note for my work on https://github.com/flatcar/Flatcar/issues/794.

pothos: So, the idea would be to run a local `python -m http.server` before running kola to give it access to the pkgs folder and then use the new image flag for the image?

pothos: Or rather not use the new image flag because it would occupy disk space on the VM, and use the temporary python http server for both.

pothos: Since this runs in VMs with private addresses I guess it will be enough to let it bind to 0.0.0.0 as done by default and then use the VM's internal IP address for the HTTP URL

pothos (approved):

PR #421, opened 2023-03-17
kola/tests/ignition/oem: Cover more code paths in the initrd

pothos:
The mount unit for the OEM in the initrd gets triggered at different points when doing a reboot with Ignition, and in the current Flatcar versions this leads to a failure because Ignition can't mount when something else already triggered the mount unit.

## How to use

Should verify that https://github.com/flatcar/Flatcar/issues/979 is fixed and doesn't come back

## Testing done

Manually ran both tests for the image from https://github.com/flatcar/bootengine/pull/58

### DIFF
--- a/kola/tests/ignition/oem.go
+++ b/kola/tests/ignition/oem.go
@@ -29,20 +29,50 @@ func init() {
 		Distros:     []string{"cl"},
 		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
 		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 2983},
-		UserData: conf.ContainerLinuxConfig(`storage:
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/oem.go
```diff
@@ -29,20 +29,50 @@ func init() {
 		Distros:     []string{"cl"},
 		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
 		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 2983},
-		UserData: conf.ContainerLinuxConfig(`storage:
+		MinVersion: semver.Version{Major: 3550},
```
krnowak, line 32: I don't know what to think about this version bump - this means that `cl.ignition.oem.regular` won't be tested on currently supported versions in other channels.

krnowak (approved):

PR #423, opened 2023-04-04
another sysext test + /usr/share/oem -> /oem

krnowak:
This adds a new sysext test that builds and activates an OEM sysext image. It also prepares the tests for the move of the OEM partition mountpoint move from `/usr/share/oem` to `/oem`. It required duplicating some of the tests, so the variants with `/usr/share/oem` will be run for Flatcar versions lower than 3603.0.0 (so new alpha release being 3602.0.0 will run the `/usr/share/oem` variants), and the variants with `/oem` will be run for Flatcar 3603.0.0 or newer.

### DIFF
--- a/kola/tests/ignition/oem.go
+++ b/kola/tests/ignition/oem.go
@@ -89,7 +89,7 @@ func init() {
 		Distros:     []string{"cl"},
 		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
 		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 3603},
+		MinVersion: semver.Version{Major: 3620},
 		UserData:   conf.Butane(withNewOEMMountpoint(regularButaneConfigTemplate)),
 	})
 	register.Register(&register.Test{
@@ -113,7 +113,7 @@ func init() {
 		Distros:     []string{"cl"},
 		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
 		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 3603},
+		MinVersion: semver.Version{Major: 3620},
 		UserData:   conf.Butane(withNewOEMMountpoint(indirectButaneConfigTemplate)),
 	})
 	register.Register(&register.Test{
--- a/kola/tests/ignition/oem.go
+++ b/kola/tests/ignition/oem.go
@@ -15,31 +15,30 @@
 package ignition
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/coreos/go-semver/semver"
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
 	"github.com/flatcar/mantle/platform/conf"
 )
 
-func init() {
-	register.Register(&register.Test{
-		Name:        "cl.ignition.oem.regular",
-		Run:         reusePartition,
-		ClusterSize: 1,
-		Distros:     []string{"cl"},
-		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
-		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 3549},
-		UserData: conf.Butane(`---
+const (
+	oemPlaceholder   string = "@OEM@"
+	oldOEMMountpoint string = "/usr/share/oem"
+	newOEMMountpoint string = "/oem"
+
+	regularButaneConfigTemplate string = `---
 variant: flatcar
 version: 1.0.0
 storage:
   filesystems:
      - device: "/dev/disk/by-label/OEM"
        format: "btrfs"
-       path: /usr/share/oem
+       path: @OEM@
   files:
-    - path: /usr/share/oem/grub.cfg
+    - path: @OEM@/grub.cfg
       mode: 0644
       overwrite: true
       contents:
@@ -47,24 +46,13 @@ storage:
           set linux_append="flatcar.autologin"
           # Needed if --qemu-skip-mangle is not set
           set linux_console="console=ttyS0,115200"
-`),
-	})
-	register.Register(&register.Test{
-		// Check new behavior from https://github.com/flatcar/bootengine/pull/58
-		// to not have to specify the OEM filesystem in Ignition
-		Name:        "cl.ignition.oem.indirect",
-		Run:         reusePartition,
-		ClusterSize: 1,
-		Distros:     []string{"cl"},
-		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
-		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 3550},
-		UserData: conf.Butane(`---
+`
+	indirectButaneConfigTemplate string = `---
 variant: flatcar
 version: 1.0.0
 storage:
   files:
-    - path: /usr/share/oem/grub.cfg
+    - path: @OEM@/grub.cfg
       mode: 0644
       overwrite: true
       contents:
@@ -72,18 +60,8 @@ storage:
           set linux_append="flatcar.autologin"
           # Needed if --qemu-skip-mangle is not set
           set linux_console="console=ttyS0,115200"
-`),
-	})
-	register.Register(&register.Test{
-		Name:        "cl.ignition.oem.reuse",
-		Run:         reusePartition,
-		ClusterSize: 1,
-		Distros:     []string{"cl"},
-		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
-		Platforms:  []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 2983},
-		// Using CLC format here also covers the ign-converter case
-		UserData: conf.ContainerLinuxConfig(`storage:
+`
+	reuseContainerLinuxConfig string = `storage:
   filesystems:
      - name: oem
        mount:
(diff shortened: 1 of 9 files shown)


kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,178 @@ import (
...
+
+sudo systemd-nspawn \
+        --console=pipe \
+        --bind-ro=/home/core/dev-container-script.sh \
+        --bind="${USR_SRC_DIR}:/work" \
+        --image=flatcar_developer_container.bin \
+        --machine=flatcar-developer-container \
+        /bin/bash /home/core/dev-container-script.sh "${1}"
+
+sudo sed -i'' -e 's/^ID=.*/ID=test/' /oem/oem-release
```
pothos, line 80: I don't think `oem-release` will exist, you have to create it.

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,178 @@ import (
...
+version_id=$(source /etc/release; echo "${VERSION_ID}")
+board=$(source /usr/share/flatcar/release; echo "${FLATCAR_RELEASE_BOARD}")
+
+line='SYSEXT_LEVEL=1.0'
+if [[ "${1:-}" = 'version-id' ]]; then line="VERSION_ID=${version_id}"; fi
+
+mkdir -p /work/sysext_rootfs/usr/share/flatcar-sysext-kola-test
+echo "${version_id}" >/work/sysext_rootfs/usr/share/flatcar-sysext-kola-test/file
+mkdir -p /work/sysext_rootfs/usr/lib/extension-release.d
+sysext_arch=x86_64
```
pothos, line 45:
```suggestion
sysext_arch=x86-64
```

kola/tests/util/devcontainer.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+type (
+	downloadLibraryParameters struct {
+		ImageDirectoryURLTemplate string
+	}
+)
+
+var (
+	downloadLibraryTemplate = TrimLeftSpace(`
```
pothos, line 32: While having this is better than having to remember to add all the required `curl` flags, I think the most straightforward way in mantle is to let Ignition download things for you.
So, this function could emit the URL for use in the Ignition configuration. The unpacking with lbzip2 would have to be done on the host still but won't really need to be a template/function.

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,178 @@ import (
...
+	mainScript = util.TrimLeftSpace(`
+#!/bin/bash
+
+set -x
+
+set -euo pipefail
+
+source /home/core/download-library.sh
+
+download_dev_container_image flatcar_developer_container.bin
```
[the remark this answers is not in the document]
pothos: Using mkosi/systemd-repart would be interesting for a dm-verity signed sysext. I'm fine with keeping to use the devcontainer but maybe it won't work well because in nspawn one can't easily use loopback devices, or?

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,178 @@ import (
...
+        --bind="${USR_SRC_DIR}:/work" \
+        --image=flatcar_developer_container.bin \
+        --machine=flatcar-developer-container \
+        /bin/bash /home/core/dev-container-script.sh "${1}"
+
+sudo sed -i'' -e 's/^ID=.*/ID=test/' /oem/oem-release
+sudo mkdir -p /oem/sysext
+sudo mv "${workdir}/oem-test-*.raw /oem/sysext
+sudo touch /oem/sysext/active-oem-test
+sudo flatcar-reset --keep-machine-id
```
pothos, line 84:
```suggestion
sudo flatcar-reset --keep-machine-id --keep-paths /var/log
```

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,178 @@ import (
...
+sysext_arch=x86_64
+if [[ "${board}" = 'arm64-usr' ]]; then sysext_arch=arm64; fi
+metadata=(
+    'ID=flatcar'
+    "${line}"
+    "ARCHITECTURE=${sysext_arch}"
+)
+metadata_file=/work/sysext_rootfs/usr/lib/extension-release.d/extension-release.oem-test
+printf '%s\n' "${metadata[@]}" >"${metadata_file}"
+mksquashfs /work/sysext_rootfs "/work/oem-test-${version}.raw" -all-root
```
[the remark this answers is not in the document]
krnowak: I thought that this would be a nice integration test for checking if devcontainer could be a suitable place to create sysexts. Which was the reason I added squashfs-tools to it. :)

kola/tests/systemd/sysext.go
pothos: The kola qemu tests use the generic image, there won't be an `oem-release` file.

krnowak: Fixed.

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,186 @@ import (
...
+        echo "Expected ${sysext_file} to exist, contents of workdir:"
+        ls -la "${workdir}"
+        exit 1
+fi
+
+# Rebrand our image from "qemu" to "test".
+if [[ ! -e /oem/oem-release ]]; then
+        printf '%s\n' 'ID=test' 'VERSION_ID=1.0.0' 'NAME=testing stuff' | sudo tee /oem/oem-release >/dev/null
+else
+        sudo sed -i'' -e 's/^ID=.*/ID=test/' /oem/oem-release
```
pothos, line 90:
```suggestion
        # This only works when the OEM setup is optional, e.g., with the QEMU OEM image
        sudo sed -i'' -e 's/^ID=.*/ID=test/' /oem/oem-release
```

kola/tests/systemd/sysext.go
```diff
@@ -11,9 +11,186 @@ import (
...
+
+if [[ ! -e "${sysext_file}" ]]; then
+        echo "Expected ${sysext_file} to exist, contents of workdir:"
+        ls -la "${workdir}"
+        exit 1
+fi
+
+# Rebrand our image from "qemu" to "test".
+if [[ ! -e /oem/oem-release ]]; then
+        printf '%s\n' 'ID=test' 'VERSION_ID=1.0.0' 'NAME=testing stuff' | sudo tee /oem/oem-release >/dev/null
```
pothos, line 88:
```suggestion
        # This is the regular case, on the generic image used for kola QEMU tests there is no OEM setup
        printf '%s\n' 'ID=test' 'VERSION_ID=1.0.0' 'NAME=testing stuff' | sudo tee /oem/oem-release >/dev/null
```

krnowak:
TODO:
- needs a bump in min version for the custom oem test
- the `/usr/share/oem` to `/oem` change here need to be done in a way that does not affect running tests on alpha, beta and stable. I tried running the tests using mantle with this PR on alpha and the following tests failed:
   - cl.sysext.boot
   - systemd.sysext.custom-oem
   - systemd.sysext.simple
   - cl.ignition.oem.regular
   - cl.ignition.oem.reuse

krnowak:
TODO:
- needs a bump in min version for the custom oem test
- the `/usr/share/oem` to `/oem` change here need to be done in a way that does not affect running tests on alpha, beta and stable. I tried running the tests using mantle with this PR on alpha aand the following tests failed:
   - cl.sysext.boot
   - systemd.sysext.custom-oem
   - systemd.sysext.simple
   - cl.ignition.oem.regular
   - cl.ignition.oem.reuse

krnowak:
> TODO:
> 
> * needs a bump in min version for the custom oem test
> * the `/usr/share/oem` to `/oem` change here need to be done in a way that does not affect running tests on alpha, beta and stable. I tried running the tests using mantle with this PR on alpha and the following tests failed:
>   
>   * cl.sysext.boot
>   * systemd.sysext.custom-oem (shouldn't be run, needs a min version bump)
>   * systemd.sysext.simple
>   * cl.ignition.oem.regular
>   * cl.ignition.oem.reuse

All done.

kola/tests/ignition/oem.go
```diff
@@ -47,18 +50,46 @@ storage:
...
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 3603},
+		UserData: conf.Butane(`---
+variant: flatcar
+version: 1.0.0
+storage:
+  filesystems:
+     - device: "/dev/disk/by-label/OEM"
+       format: "btrfs"
+       path: /oem
```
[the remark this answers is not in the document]
pothos: Yes :( We could try to support it by patching the special /usr/share/oem path to /oem in Ignition's mount logic…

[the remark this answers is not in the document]
krnowak: I'll have a look.

kola/tests/ignition/oem.go
```diff
@@ -47,18 +50,46 @@ storage:
...
+          # Needed if --qemu-skip-mangle is not set
+          set linux_console="console=ttyS0,115200"
 `),
 	})
 	register.Register(&register.Test{
 		// Check new behavior from https://github.com/flatcar/bootengine/pull/58
 		// to not have to specify the OEM filesystem in Ignition
-		Name:        "cl.ignition.oem.indirect",
-		Run:         reusePartition,
+		Name:        "cl.ignition.oem.indirect.old",
```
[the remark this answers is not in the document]
pothos: I mean, we can have two tests but we don't need to limit the version of the old test, it's even better to keep it to check that we are compatible with older configs.

kola/tests/ignition/oem.go
```diff
@@ -75,13 +106,62 @@ storage:
...
+         format: "ext4"
+  files:
+    - path: /grub.cfg
+      filesystem: oem
+      mode: 0644
+      contents:
+        inline: |
+          set linux_append="flatcar.autologin"`),
+	})
+	register.Register(&register.Test{
```
pothos, line 157: Why is this duplicated? I don't see a difference (on the first look, maybe I miss something)?

kola/tests/ignition/oem.go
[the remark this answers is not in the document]
krnowak: CI with the change is running, let's see if it will work.

kola/tests/ignition/oem.go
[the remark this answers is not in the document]
krnowak: Kept the duplicated tests, but dropped the `EndVersion`, so both variants will run on a new enough Flatcar version. Changed the names too, (so instead of `foo.old` and `foo`, there's now `foo` and `foo.new`).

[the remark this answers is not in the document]
krnowak: Updated the changelog too.

krnowak:
The old and new `ignition.oem` tests seem to be passing on sufficiently new images and old tests too for older versions ([as seen in the test run](http://jenkins.infra.kinvolk.io:8080/job/container/job/test/10591/consoleText)). Before merging this, I'll bump the minversion of the new tests to 3613 or so.

kola/tests/ignition/oem.go
```diff
@@ -95,7 +73,83 @@ storage:
...
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 2983},
+		// Using CLC format here also covers the ign-converter case
+		UserData: conf.ContainerLinuxConfig(reuseContainerLinuxConfig),
+	})
+	register.Register(&register.Test{
+		Name:        "cl.ignition.oem.reuse.new",
```
pothos, line 144: I wouldn't duplicate this test case because the used config is the same and only the check differs. The check doesn't have to differ because `grep /usr/share/oem/grub.cfg` should still work since `/usr/share/oem` is a symlink to `/oem`.

kola/tests/ignition/oem.go
```diff
@@ -95,7 +73,83 @@ storage:
...
+		Run:         reusePartitionNew,
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+		// This test overwrites the grub.cfg which does not work on cloud environments after reboot
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 3603},
+		// Using CLC format here also covers the ign-converter case
+		UserData: conf.ContainerLinuxConfig(reuseContainerLinuxConfig),
```
pothos, line 152:
```suggestion
```

pothos (approved):
Otherwise good, thanks :)

PR #424, opened 2023-04-13
kola: add user journald test

tormath1:
In this PR, Kola gains the ability to use a user different from `core` for the tests and the underlying SSH operations.

It's now possible to override the `core` user when a test is register with:
```go
DefaultUser: "my-user"
```

This user will be automatically added to the `sudo` group and all operations (including the test itself) will run with `my-user`.

It's possible to customize the user with regular Ignition configuration, example if the test has:
```yaml
passwd:
  users:
    - name: flatcar
      groups:
        - systemd-journal
```

The final Ignition will be:
```json
{
  "passwd": {
    "users": [
      {
        "groups": [
          "systemd-journal",
          "sudo"
        ],
        "name": "flatcar",
        "sshAuthorizedKeys": [
          "ssh-rsa ..."
        ]
      }
    ]
  }
```

The `DefaultUser` inherits from the `core` user SSH keys, so it's possible to SSH into an instance using the created user to inspect the instance.

<hr>

The actual goal of this PR is to test `journalctl --user` which was broken on Flatcar (https://github.com/flatcar/Flatcar/issues/989)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- `DefaultUser` parameter when registering a test to use a user different from `core` ([#424](https://github.com/flatcar/mantle/pull/424))
+
 ### Change
 
 ### Removed


platform/conf/conf.go
```diff
@@ -1297,3 +1297,104 @@ func AddSSHKeys(userdata *UserData, keys *[]agent.Key) *UserData {
 	}
 	return userdata
 }
+
+// AddUserToGroups add the user to the given groups
+func (c *Conf) AddUserToGroups(user string, groups []string) {
```
krnowak, line 1302: I think this function should return an error if an incompatible kind of configuration is used (e.g. cloudconfig). Just to avoid a situation where this function silently does nothing.

krnowak: I just gave you an easy way out. :P

krnowak (approved):

PR #425, opened 2023-04-18
Replace kinvolk links and docker images with flatcar ones, sync docker images

krnowak:
- replaced some links
- replaced kinvolk/busybox, kinvolk/nginx and kinvolk/calico with flatcar variants
  - https://github.com/flatcar/scripts/pull/716 needs to go in and the actions there need to be run first
- fixed a final newline issue with kubeadm templates


krnowak:
To make the syncing work, I needed to add mantle in the actions access in every calico package settings. So "packages write" permissions of the github token was not enough.

PR #429, opened 2023-04-26
sysext: bump bakery commit ID

tormath1:
systemd-sysext will soon automatically reload the daemon - units that must be enabled should be "Upheld" by the target.

<hr>

I don't think it's harmful to merge it right now.

### DIFF
--- a/kola/tests/systemd/sysext.go
+++ b/kola/tests/systemd/sysext.go
@@ -84,7 +84,7 @@ func checkSysextCustomDocker(c cluster.TestCluster) {
 	// First assert that Docker doesn't work because Torcx is disabled
 	_ = c.MustSSH(c.Machines()[0], cmdNotWorking)
 	// We build a custom sysext image locally because we don't host them somewhere yet
-	_ = c.MustSSH(c.Machines()[0], `git clone https://github.com/flatcar/sysext-bakery.git && git -C sysext-bakery checkout 2d6ab6bdba96476138c7db841e9a4c4ed38bd3fc`)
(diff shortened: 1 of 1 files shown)


kola/tests/systemd/sysext.go
```diff
@@ -84,7 +84,7 @@ func checkSysextCustomDocker(c cluster.TestCluster) {
 	// First assert that Docker doesn't work because Torcx is disabled
 	_ = c.MustSSH(c.Machines()[0], cmdNotWorking)
 	// We build a custom sysext image locally because we don't host them somewhere yet
-	_ = c.MustSSH(c.Machines()[0], `git clone https://github.com/flatcar/sysext-bakery.git && git -C sysext-bakery checkout 2d6ab6bdba96476138c7db841e9a4c4ed38bd3fc`)
+	_ = c.MustSSH(c.Machines()[0], `git clone https://github.com/flatcar/sysext-bakery.git && git -C sysext-bakery checkout e68d2fe25c8412f4774477d1d75c40f615145c46`)
```
krnowak, line 87: Why do we even bother with checking out a specific commit?

pothos: The commit ID was for consistent kola test behavior, I'm fine with git tags/releases as well :)

pothos: For predictable behavior one could also take the last commit since the Flatcar build date but that relation is maybe a bit confusing.

pothos (approved):

PR #433, opened 2023-05-03
Add tests for the NVIDIA GPU tests

sayanchowdhury:
# Add tests for the NVIDIA GPU tests

## Testing done

### DIFF
--- a/kola/tests/misc/nvidia.go
+++ b/kola/tests/misc/nvidia.go
@@ -1,0 +1,52 @@
+package misc
+
+import (
+	"bytes"
+	"fmt"
+	"time"
+
+	"github.com/coreos/pkg/capnslog"
+	"github.com/flatcar/mantle/kola"
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/util"
+)
+
+const (
+	CmdTimeout = time.Second * 300
+)
+
+var plog = capnslog.NewPackageLogger("github.com/flatcar/mantle", "kola/tests/misc")
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "cl.misc.nvidia",
+		Run:         verifyNvidiaInstallation,
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+		// This test is to test the NVIDIA installation, limited to AZURE for now
+		Platforms:     []string{"azure"},
+		Architectures: []string{"amd64"},
+		Flags:         []register.Flag{register.NoEnableSelinux},
+	})
+}
+
+func verifyNvidiaInstallation(c cluster.TestCluster) {
+	if kola.AzureOptions.Size != "Standard_NC6s_v3" {
+		c.Skip("skipping due to wrong instance size")
+	}
+	m := c.Machines()[0]
+
+	nvidiaStatusRetry := func() error {
+		out, err := c.SSH(m, "systemctl is-active nvidia.service")
+		if !bytes.Contains(out, []byte("inactive")) {
+			return fmt.Errorf("nvidia.service: %q: %v", out, err)
+		}
+		return nil
+	}
+
+	if err := util.Retry(40, 15*time.Second, nvidiaStatusRetry); err != nil {
+		c.Fatal(err)
+	}
+	c.AssertCmdOutputContains(m, "/opt/bin/nvidia-smi", "Tesla")
+}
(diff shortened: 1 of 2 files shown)


kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+
+		return nil
+	}
+
+	if err := util.Retry(20, 15*time.Second, nvidiaStatusRetry); err != nil {
+		c.Fatal(err)
+	}
+
+	out, err := c.SSH(m, "/opt/bin/nvidia-smi")
+	if !bytes.Contains(out, []byte("525.105.17")) {
```
jepio, line 49: This is going to require maintenance at every release - can the version be queried from flatcar metadata in the image? Or maybe just check that nvidia-smi succeeds, and shows a GPU.

kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "cl.misc.nvidia",
+		Run:         verifyNvidiaInstallation,
+		ClusterSize: 1,
+		Distros:     []string{"cl"},
+		// This test is to test the NVIDIA installation, limited to AZURE for now
+		Platforms:     []string{"azure"},
+		Architectures: []string{"amd64"},
```
pothos, line 25: What instance type does this need? I think we would have to skip this by default and invoke it similar to the cl.internet tests that run on various instance types.

kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+}
+
+func verifyNvidiaInstallation(c cluster.TestCluster) {
+	m := c.Machines()[0]
```
pothos, line 30: With
```suggestion
	if kola.AzureOptions.Size != "Standard_NC6s_v3" {
		c.Skip("skipping due to wrong instance size")
	}
	m := c.Machines()[0]
```
I think we could do this in scripts:
```
diff --git a/ci-automation/vendor-testing/azure.sh b/ci-automation/vendor-testing/azure.sh
index 17081b3598..c57210d651 100755
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -74,7 +74,7 @@ query_kola_tests() {
     kola list --platform=azure --filter "${@}"
 }
 
-other_instance_types=()
+other_instance_types=('Standard_NC6s_v3')
 if [[ "${CIA_ARCH}" = 'amd64' ]]; then
     other_instance_types+=('V1')
 fi
@@ -85,6 +85,6 @@ run_kola_tests_on_instances \
     "${CIA_FIRST_RUN}" \
     "${other_instance_types[@]}" \
     '--' \
-    'cl.internet' \
+    'cl.internet' 'cl.misc.nvidia' \
     '--' \
     "${@}"
```

kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+		if !bytes.Contains(out, []byte("inactive")) {
+			return fmt.Errorf("nvidia.service: %q: %v", out, err)
+		}
+		return nil
+	}
+
+	if err := util.Retry(40, 15*time.Second, nvidiaStatusRetry); err != nil {
+		c.Fatal(err)
+	}
+	c.AssertCmdOutputContains(m, "/opt/bin/nvidia-smi", "Tesla")
```
pothos, line 51: This path would change with the move to `/usr`, don't we have `/opt/bin/` in PATH? Maybe `sh -c nvidia-smi`?

pothos: Oh no, that's installed by the installer, right?

sayanchowdhury: `setup-nvidia` installs in the `/opt/bin/`. Moving to `/usr` will be a breaking change IMO. We can move to `/usr`, but I think it would need to have a transition phase. I won't recommend it now.

kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	CmdTimeout = time.Second * 300
+)
+
+var plog = capnslog.NewPackageLogger("github.com/flatcar/mantle", "kola/tests/misc")
+
+func init() {
+	register.Register(&register.Test{
```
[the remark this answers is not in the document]
sayanchowdhury: I mean it should work ever since the NVIDIA with Flatcar has been working. it's been 3 years now, do you think we need to add a minimal version?

kola/tests/misc/nvidia.go
```diff
@@ -0,0 +1,52 @@
...
+	if kola.AzureOptions.Size != "Standard_NC6s_v3" {
+		c.Skip("skipping due to wrong instance size")
+	}
+	m := c.Machines()[0]
+
+	nvidiaStatusRetry := func() error {
+		out, err := c.SSH(m, "systemctl is-active nvidia.service")
+		if !bytes.Contains(out, []byte("inactive")) {
+			return fmt.Errorf("nvidia.service: %q: %v", out, err)
+		}
```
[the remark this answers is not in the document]
pothos: There is also `_ = c.MustSSH(…)` which fails the test.

pothos (approved):

PR #436, opened 2023-05-22
dockerfile: update Google apt key

tormath1:
See: https://github.com/kubernetes/release/issues/2862#issuecomment-1554211504


krnowak (approved):
Hopefully the new GPG download key site will be more reliable than the old one.

PR #437, opened 2023-05-22
Add tests for multipart mime in coreos-cloudinit

gabriel-samfira:
# Add tests for the multipart mime support in coreos-cloudinit

The new multipart mime support that is currently being proposed against coreos-cloudinit needs to be tested. This PR adds the needed tests.

These tests will validate that we properly parse the multipart mime user-data and all parts are run correctly. 

### DIFF
--- a/kola/tests/misc/cloudinit.go
+++ b/kola/tests/misc/cloudinit.go
@@ -232,7 +232,7 @@ func CloudInitMultipartMime(c cluster.TestCluster) {
 
 	// All files should have the same content (42). These files should have been created by the cloud-config part
 	// that declares the write_files option.
-	c.MustSSH(m, `for f in $(ls /tmp/kola_*); do OUT=$(cat $f); if [ "$OUT" != 42 ]; then exit 1; fi; done`)
+	c.MustSSH(m, `for f in /tmp/kola_*; do OUT=$(cat $f); if [ "$OUT" != 42 ]; then exit 1; fi; done`)
 	// Check that the x-shellscript part was executed.
 	c.MustSSH(m, "test -f /coreos-cloudinit_multipart.txt")
 	c.MustSSH(m, "test -f /kola_undercover")
--- a/kola/tests/misc/cloudinit.go
+++ b/kola/tests/misc/cloudinit.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
 	"github.com/flatcar/mantle/platform/conf"
@@ -183,7 +184,7 @@ chmod 600 ~core/.ssh/authorized_keys`),
 		UserData:    conf.MultipartMimeConfig(multipartMimeUserdata),
 		Distros:     []string{"cl"},
 		Platforms:   []string{"qemu", "qemu-unpriv"},
-		MinVersion: semver.Version{Major: 3620},
+		MinVersion:  semver.Version{Major: 3620},
 	})
 }
 
--- a/kola/tests/misc/cloudinit.go
+++ b/kola/tests/misc/cloudinit.go
@@ -15,12 +15,134 @@
 package misc
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
 	"github.com/flatcar/mantle/platform/conf"
 )
+
+var multipartMimeUserdata = `Content-Type: multipart/mixed; boundary="MIMEMULTIPART"
+MIME-Version: 1.0
+
+--MIMEMULTIPART
+Content-Type: text/cloud-config; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+Content-Disposition: attachment; filename="cloud-config"
+
+ssh_authorized_keys:
+  - ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBEftQIHTRvUmyDCN7VGve4srz03Jmq6rPnqq+XMHMQUIL9c/b0l7B5tWfQvQecKyLte94HOPzAyMJlktWTVGQnY=
+
+--MIMEMULTIPART
+Content-Type: text/cloud-config; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+Content-Disposition: attachment; filename="cloud-config"
+
+hostname: "example"
+
+--MIMEMULTIPART
+Content-Type: text/cloud-config; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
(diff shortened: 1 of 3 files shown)


gabriel-samfira:
It seems I don't have access to push to ```ghcr.io/flatcar/mantle```. I'll try and push to my own fork.

pothos:
> It seems I don't have access to push to `ghcr.io/flatcar/mantle`. I'll try and push to my own fork.

Strange, I thought that would work for all PRs…

gabriel-samfira:
> > It seems I don't have access to push to `ghcr.io/flatcar/mantle`. I'll try and push to my own fork.
> 
> Strange, I thought that would work for all PRs…

Nope. This section:

https://github.com/flatcar/mantle/blob/9cdaa3bb2673beddd51198c9f59c8a5fedc7629f/.github/workflows/docker.yml#L44-L49

Defines the login credentials. The actor in this case is the PR author and I do not have write access to the mantle packages.

pothos:
Is the token the same or also bound to the author?

gabriel-samfira:
Yup. The actor and the token belong to the person triggering the workflow. If triggered via PR, they belong to the author. If triggered via ```workflow_dispatch``` they belong to the person running the action.

kola/tests/misc/cloudinit.go
```diff
@@ -78,3 +225,41 @@ func CloudInitScript(c cluster.TestCluster) {
...
+	// that declares the write_files option.
+	c.MustSSH(m, "for $file in /tmp/kola_*; do OUT=$(cat $file); if [ $OUT != 42 ]; then exit 1; fi; done")
+	// Check that the x-shellscript part was executed.
+	c.MustSSH(m, "ls -l /coreos-cloudinit_multipart.txt")
+}
+
+func CloudInitMultipartMimeParsesCloudConfig(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	// The Content-Type was set to text/plain, but the content is a valid cloud-config.
```
pothos, line 259: Is this difference worth a new test or rather something that unit tests in coreos-cloudinit would cover? Ignition will check for the `multipart/mixed` type and as long as that works, I think we are ok and could rely on the larger test case above.

gabriel-samfira: I can move it and have it write a file. Then check for that file. We can skip adding a separate test.

kola/tests/misc/cloudinit.go
```diff
@@ -15,13 +15,134 @@
...
+This is just some random text.
+
+--===============1598784645116016685==
+Content-Type: application/json; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+Content-Disposition: attachment; filename="ignition.txt"
+
+{
+  "ignitionVersion": 1,
```
pothos, line 116: The idea is that this should be ignored, or? Maybe annotate this somewhere, e.g.:
```suggestion
  "ignitionVersion": 1,
  "This ignition configuration will be ignored because it's just embedded": "only cloud-init will run",
```

gabriel-samfira: yes. This will be ignored.

gabriel-samfira:
Converted to draft for now. This opened a can of worms. Apparently I need to parse the userdata here as well to determine its type and add the testing ssh keys to it so kola can access the vms. Will update it next week.

vendor/github.com/flatcar/ignition/config/shared/errors/errors.go
```diff
@@ -29,6 +29,7 @@ var (
 	ErrScript             = errors.New("not a config (found coreos-cloudinit script)")
 	ErrDeprecated         = errors.New("config format deprecated")
 	ErrCompressionInvalid = errors.New("invalid compression method")
+	ErrMultipartMime      = errors.New("not a config (found multipart mime)")
```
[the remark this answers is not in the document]
pothos: Oh, we still use `flatcar/ignition` here… That's not maintained and frozen at only having v1 and v2 spec but we could update this old branch where the tag is. In the Ignition ebuild we nowadays use the upstream repo and apply patches on top. Could we do it differently by defining the error elsewhere?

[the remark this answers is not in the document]
pothos: Anyway it's a mess, let's push your patch to the https://github.com/flatcar/ignition/ `flatcar-master` branch and create a new tag.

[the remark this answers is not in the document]
gabriel-samfira: Wait. I think it's a great idea to define these in the mantle package. We should probably pull out all the stuff we use from flatcar/ignition and just make it local.

The stuff I accidentally added to the vendor folder are small. I can move them now. Saves us from having to update 2 repos, creating tags, syncing go.mod, etc.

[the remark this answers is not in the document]
gabriel-samfira: Ahh, we'd still need to change the ```Parse()``` function. You're right. It is a mess :smile:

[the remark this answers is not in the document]
gabriel-samfira: okay. Moved some things around. PTAL

kola/tests/misc/cloudinit.go
```diff
@@ -78,3 +208,31 @@ func CloudInitScript(c cluster.TestCluster) {
...
+	c.MustSSH(m, fmt.Sprintf("grep %s ~core/.ssh/authorized_keys", expectKey))
+
+	out = c.MustSSH(m, "ls -l /tmp/kola_*| wc -l")
+	if string(strings.Replace(string(out), "\n", "", -1)) != "7" {
+		c.Fatalf("expected 7 files in /tmp, found %q", out)
+	}
+
+	// All files should have the same content (42). These files should have been created by the cloud-config part
+	// that declares the write_files option.
+	c.MustSSH(m, `for f in $(ls /tmp/kola_*); do OUT=$(cat $f); if [ "$OUT" != 42 ]; then exit 1; fi; done`)
```
pothos, line 235:
```suggestion
	c.MustSSH(m, `for f in /tmp/kola_*; do OUT=$(cat $f); if [ "$OUT" != 42 ]; then exit 1; fi; done`)
```
If we expect the files to exist we can use this and if they don't, we get the `kola_*` filename and the test will fail as well.

platform/conf/conf.go
```diff
@@ -135,6 +149,32 @@ func Script(data string) *UserData {
 	}
 }
 
+func decompressIfGzipped(data []byte) []byte {
```
pothos, line 152: We could add a hint that these functions are also implemented in coreos-cloudinit if I'm not mistaken and should be kept in sync?

pothos (approved):
Looks good, thanks a lot! I added a few comments.

platform/conf/conf.go
pothos: Also, if we have any duplication in `platform/conf/multipart.go`, that could also benefit from pointers. Even if it's not a duplication I think it's nice to link the similar code.

gabriel-samfira: Also, they don't really need to be in sync. As long as they serve their function, they can diverge. I think we should at some point create that utility package.

pothos (approved):

PR #438, opened 2023-06-05
kola/tests/systemd: Try also different category for docker package

krnowak:
We will move docker packages from app-emulation to app-containers to match what Gentoo did with their docker packages. Since the tests will be also running on older versions of Flatcar, keep searching for docker package in old category too.


krnowak:
Related to https://github.com/flatcar/scripts/pull/864.

dongsupark:
changelog entry would be great

krnowak:
Changelog added.

PR #447, opened 2023-08-10
docker: extend restart test case

tormath1:
See also: https://github.com/flatcar/Flatcar/issues/1157

---

It passes with current Flatcar releases but not if you use SIGHUP signal.

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -504,11 +504,6 @@ func dockerUserNoCaps(c cluster.TestCluster) {
 func dockerContainerdRestart(c cluster.TestCluster) {
 	m := c.Machines()[0]
 
-	pid := c.MustSSH(m, "systemctl show containerd -p MainPID --value")
-	if string(pid) == "0" {
-		c.Fatalf("Could not find containerd pid")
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -504,38 +504,62 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	// * SIGHUP has been seen to be used by third party tools to restart containerd
+	for _, signal := range []string{"SIGTERM", "SIGHUP"} {
+		pid := c.MustSSH(m, "systemctl show containerd -p MainPID --value")
+		if string(pid) == "0" {
+			c.Fatalf("Could not find containerd pid")
+		}
 
-	// kill it
-	c.MustSSH(m, "sudo kill "+string(pid))
+		c.MustSSH(m, fmt.Sprintf("sudo kill -%s %s ", signal, string(pid)))
```
krnowak, line 528:
```suggestion
		c.MustSSH(m, fmt.Sprintf("sudo kill -%s %s", signal, string(pid)))
```

krnowak (approved):

PR #451, opened 2023-09-15
kola/nvidia: Change the nvidia retry clause

sayanchowdhury:
The nvidia.service has been change to `oneshot` and sets `RemainAfterExit=true` which keeps the service active after it exits.


sayanchowdhury:
CI Passed: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/15150/console

PR #457, opened 2023-09-29
kola: Add test case for extra partition creation

jepio:
This is a testcase for a regression that passes until Flatcar 3374 and fails on Flatcar 3510 (only checked stable releases).

### DIFF
--- a/kola/tests/ignition/filesystem.go
+++ b/kola/tests/ignition/filesystem.go
@@ -325,6 +325,9 @@ systemd:
 	})
 
 	register.Register(&register.Test{
+		// Note: As long as the initrd does not mount /var for early
+		// preparations, the resulting system has a few issues, so this setup is
+		// not fully supported yet
 		Name:        "cl.ignition.partition_on_boot_disk",
 		Run:         testPartitionOnBootDisk,
 		ClusterSize: 0,
@@ -456,18 +459,23 @@ storage:
       number: 10
       start: '9GiB'
   filesystems:
-  - name: var
+  - name: VAR
     mount:
       device: /dev/disk/by-partlabel/VAR
       format: xfs
-      label: var
+      label: VAR
   files:
+  - filesystem: VAR
+    path: /hello
+    mode: 0644
+    contents:
+      inline: world
   - filesystem: root
     path: /etc/fstab
     mode: 0644
     contents:
       inline: |
-        /dev/disk/by-label/var /var xfs defaults 0 0
+        /dev/disk/by-label/VAR /var xfs defaults 0 0
 `)
 
 func testPartitionOnBootDisk(c cluster.TestCluster) {
@@ -482,6 +490,7 @@ func testPartitionOnBootDisk(c cluster.TestCluster) {
 
 	c.MustSSH(m, "mountpoint /var")
 	c.MustSSH(m, "ls -la /dev/disk/by-partlabel/VAR")
-	c.MustSSH(m, "ls -la /dev/disk/by-label/var")
+	c.MustSSH(m, "ls -la /dev/disk/by-label/VAR")
 	c.AssertCmdOutputContains(m, "findmnt /var", "xfs")
+	c.AssertCmdOutputContains(m, "cat /var/hello", "world")
 }
--- a/kola/tests/ignition/filesystem.go
+++ b/kola/tests/ignition/filesystem.go
@@ -21,6 +21,7 @@ import (
 	"github.com/coreos/go-semver/semver"
 	"github.com/flatcar/mantle/kola/cluster"
 	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/kola/tests/util"
 	"github.com/flatcar/mantle/platform/conf"
 )
 
@@ -322,6 +323,14 @@ systemd:
 		Platforms:  []string{"qemu", "qemu-unpriv"},
 		MinVersion: semver.Version{Major: 3033},
 	})
+
+	register.Register(&register.Test{
+		Name:        "cl.ignition.partition_on_boot_disk",
+		Run:         testPartitionOnBootDisk,
+		ClusterSize: 0,
+		Distros:     []string{"cl"},
+		Platforms:   []string{"qemu", "qemu-unpriv"},
+	})
 }
 
 var ext4NoClobberV2_1 = conf.Ignition(`{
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/partitions.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/kola/tests/util"
+	"github.com/flatcar/mantle/platform/conf"
+)
+
+const (
+	userdata = `{
```
pothos, line 13: I think it would be fitting to add this test under `ignition/filesystem.go` and use Butane or CLC for transpilation.

jepio: moved it and switched to CLC

kola/tests/ignition/filesystem.go
```diff
@@ -437,3 +446,42 @@ func testSwapActivation(c cluster.TestCluster) {
...
+  disks:
+  - device: /dev/disk/by-id/virtio-primary-disk
+    partitions:
+    - label: VAR
+      number: 10
+      start: '9GiB'
+  filesystems:
+  - name: var
+    mount:
+      device: /dev/disk/by-partlabel/VAR
```
pothos, line 461: For `/var` we should also make sure that the mount point is set up in the initrd with Ignition because systemd-tmpfiles should populate `/var` from the initrd. I think with CLC and the translation to Ignition v3 this is complicated and if you want to test `/var` we should rather use Butane. If it's not really about `/var` (or `/home` and similar) but an arbitrary mount point then it's easier to use CLC.

kola/tests/ignition/filesystem.go
```diff
@@ -322,6 +323,14 @@ systemd:
 		Platforms:  []string{"qemu", "qemu-unpriv"},
 		MinVersion: semver.Version{Major: 3033},
 	})
+
+	register.Register(&register.Test{
+		Name:        "cl.ignition.extra_root_part",
```
pothos, line 328: The naming sounds like a root partition gets created, how about something like `partition_on_boot_disk`?

kola/tests/ignition/filesystem.go
jepio: > For /var we should also make sure that the mount point is set up in the initrd with Ignition because systemd-tmpfiles should populate /var from the initrd.

we don't populate `/var` with `systemd-tmpfiles` from the initrd, and ignition unmounts mounted partitions in one of the last stages. It does this to ensure a system does not depend on ignition performing the mount.

> I think with CLC and the translation to Ignition v3 this is complicated and if you want to test /var we should rather use Butane. If it's not really about /var (or /home and similar) but an arbitrary mount point then it's easier to use CLC.

It's not about the mountpoint at all, it's about adding partitions to a partitioned disk (that's why this was in `partitions.go`).

For sure we need a testcase that ensures we don't break backwards compat with ignition v2/CLC. `/var` is also a location users are likely to try to put on a separate filesystem.

kola/tests/ignition/filesystem.go
jepio: ok

kola/tests/ignition/filesystem.go
pothos: We do, that's in `base_image_var.conf` but yeah, seems that we have a problem in general with separate `/var` partitions, at least on the first boot before the final system also had a chance to set up the files in `/var`. What one might see as result is either a failed unit or a misbehaving system, e.g., the first boot journal might not be persisted (this was a bug we fixed by letting the initrd populate `/var` correctly).

pothos: FYI: The topic of separate `/var` is also relevant for applying sysext images from the initrd. The suggestion from Lennart was to establish a kernel command line setting similar to what systemd has for `/usr`. It would then be properly mounted in `/sysroot/var`.

pothos: What we can also do in Flatcar is to define a partition or filesystem label to expect and then add a mount unit for that (it can be compatible with a future systemd cmdline way by having this mount unit only be set up if the cmdline option is not there, I think we have similar logic for `/usr`).

kola/tests/ignition/filesystem.go
```diff
@@ -437,3 +446,42 @@ func testSwapActivation(c cluster.TestCluster) {
...
+    partitions:
+    - label: VAR
+      number: 10
+      start: '9GiB'
+  filesystems:
+  - name: var
+    mount:
+      device: /dev/disk/by-partlabel/VAR
+      format: xfs
+      label: var
```
pothos, line 463:
```suggestion
      label: VAR
```
If we want to support a separate `/var` partition it must be found from the initrd. For that I suggest to use a label with a similar style as we already have with the uppercase `ROOT` and `OEM` filesystem labels.

kola/tests/ignition/filesystem.go
pothos: Started something for supporting separate `/var`: https://github.com/flatcar/bootengine/pull/75

kola/tests/ignition/filesystem.go
```diff
@@ -437,3 +446,42 @@ func testSwapActivation(c cluster.TestCluster) {
...
+  files:
+  - filesystem: root
+    path: /etc/fstab
+    mode: 0644
+    contents:
+      inline: |
+        /dev/disk/by-label/var /var xfs defaults 0 0
```
pothos, line 470:
```suggestion
        /dev/disk/by-label/VAR /var xfs defaults 0 0
```

kola/tests/ignition/filesystem.go
```diff
@@ -437,3 +446,42 @@ func testSwapActivation(c cluster.TestCluster) {
...
+		c.Fatal(err)
+	}
+	out := c.MustSSH(m, "lsblk -f")
+	c.Logf("lsblk -f:\n%s", out)
+	out = c.MustSSH(m, "findmnt")
+	c.Logf("findmnt:\n%s", out)
+
+	c.MustSSH(m, "mountpoint /var")
+	c.MustSSH(m, "ls -la /dev/disk/by-partlabel/VAR")
+	c.MustSSH(m, "ls -la /dev/disk/by-label/var")
```
pothos, line 485:
```suggestion
	c.MustSSH(m, "ls -la /dev/disk/by-label/VAR")
```

kola/tests/ignition/filesystem.go
pothos: Would be good to write some files here with Ignition. We also have to add one more test case with Butane an defining the `/var` path in the initrd and writing some files because the uplifted v3 path is generated and can't be controlled whereas with Butane we can use the `/sysroot/var` mount point and have to check that this works.

kola/tests/ignition/filesystem.go
```diff
@@ -437,3 +446,42 @@ func testSwapActivation(c cluster.TestCluster) {
...
+var rootDiskExtraPartition = conf.ContainerLinuxConfig(`
+storage:
+  disks:
+  - device: /dev/disk/by-id/virtio-primary-disk
+    partitions:
+    - label: VAR
+      number: 10
+      start: '9GiB'
+  filesystems:
+  - name: var
```
pothos, line 459:
```suggestion
  - name: VAR
```

kola/tests/ignition/filesystem.go
```diff
@@ -322,6 +323,14 @@ systemd:
 		Platforms:  []string{"qemu", "qemu-unpriv"},
 		MinVersion: semver.Version{Major: 3033},
 	})
+
+	register.Register(&register.Test{
+		Name:        "cl.ignition.partition_on_boot_disk",
```
pothos, line 328:
```suggestion
		// Note: As long as the initrd does not mount /var for early preparations, the resulting system has a few issues, so this setup is not fully supported yet
		Name:        "cl.ignition.partition_on_boot_disk",
```

jepio: added that comment

kola/tests/ignition/filesystem.go
jepio: so i added the file creation part, but not the butane yet

pothos (approved):

PR #459, opened 2023-10-04
misc: add fips test

tormath1:
This is documented for now but not tested: https://www.flatcar.org/docs/latest/setup/security/fips/

Out of the box config is available since OpenSSL 3.0.8 for Flatcar, so the incoming Stable will work - we only exclude LTS. It does not mean that FIPS does not work on LTS, it's just that it takes extra steps to activate it.

### DIFF
--- a/kola/tests/misc/fips.go
+++ b/kola/tests/misc/fips.go
@@ -1,0 +1,65 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package misc
+
+import (
+	"github.com/coreos/go-semver/semver"
+	"github.com/flatcar/mantle/kola/cluster"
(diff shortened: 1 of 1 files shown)


kola/tests/misc/fips.go
```diff
@@ -0,0 +1,65 @@
...
+
+}
+
+func fipsTest(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	// It works because SHA is FIPS compliant.
+	c.MustSSH(m, "echo Flatcar | openssl sha512 -")
+
+	// Should exits with 0.
```
krnowak, line 56:
```suggestion
	// Should exit with 0.
```

krnowak (approved):

krnowak (approved):

PR #464, opened 2023-10-12
Use shlex to parse version.txt

krnowak:


krnowak:
Will get back to it maybe some other time.

PR #465, opened 2023-10-13
Kola: update to handle torcx deprecation / docker and containerd sysext

t-lo:
This PR prepares kola for torcx deprecation. It is a prerequisite for https://github.com/flatcar/scripts/pull/1216 . 

- A new "sysext" test module has been introduced; the existing `sysext.go` has been moved there from the `systemd` module.
  - two additional tests for disabling the docker and containerd sysexts have been added there, too.
- The "torcx" test module has been removed entirely - it contained a single test (explicitly enabling docker) which needed to be extended to cover sysext docker. That test has been moved to `docker/enable_docker.go`.
- Torcx tests are made to run conditionally and will run only for major version below a certain threshold (currently up to version 3745 but this will need to be updated depending on when https://github.com/flatcar/scripts/pull/1216 gets merged).
- New sysext tests run for major versions greater than 3746 (same conditions apply)
- A number of tests have been amended to better handle sysext environments (some implicit assumptions don't hold true anymore): `kola/tests/bpf/bpf.go` and `kola/tests/systemd/journald.go`.
- `kola/cluster/cluster.go` has been extended to print SSH commands before execution (log level INFO) and command output after execution (log level DEBUG).

## How to use

- build the container: `docker build -t my-mantle-container .`
- Check out the torcx deprecation WIP branch https://github.com/flatcar/scripts/tree/contrib/torcx-deprecation-docker-sysext
- build packages, image, and qemu_uefi VM
- edit `sdk_container/.repo/manifests/mantle-container` and put in `my-mantle-container`
- Run `./run_local_test.sh`

Note that the `sysext.custom-oem` test may still fail - it tries to download a devcontainer which might not exist for your particular image version.

## Testing done

- Built a mantle container and ran the full suite of tests locally (see "How to use" above) on an AMD64 OS image built from the latest `torcx-deprecation-docker-sysext` branch.

### DIFF
--- a/kola/tests/docker/torcx_manifest_pkgs.go
+++ b/kola/tests/docker/torcx_manifest_pkgs.go
@@ -37,12 +37,9 @@ func init() {
 		ExcludePlatforms: []string{"do"},
 		Distros:          []string{"cl"},
 		// This test is normally not related to the cloud environment
-		Platforms: []string{"qemu", "qemu-unpriv"},
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		EndVersion: semver.Version{Major: 3745},
 		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
-			// Releases after 3745 don't ship torcx at all (these use sysext)
-			if (0 < version.Compare(semver.Version{Major: 3745})) {
-				return true
-			}
 			// LTS (3033) does not have the network-kargs service pulled in:
 			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
 			// Let's skip this test for < 3034 on ESX.
--- a/kola/tests/docker/torcx_manifest_pkgs.go
+++ b/kola/tests/docker/torcx_manifest_pkgs.go
@@ -39,10 +39,10 @@ func init() {
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
 		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
-            // Releases after 3745 don't ship torcx at all (these use sysext)
-            if ( 0 < version.Compare(semver.Version{Major: 3745}) ) {
-                return true
-            }
+			// Releases after 3745 don't ship torcx at all (these use sysext)
+			if (0 < version.Compare(semver.Version{Major: 3745})) {
+				return true
+			}
 			// LTS (3033) does not have the network-kargs service pulled in:
 			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
 			// Let's skip this test for < 3034 on ESX.
--- a/kola/tests/docker/torcx_manifest_pkgs.go
+++ b/kola/tests/docker/torcx_manifest_pkgs.go
@@ -39,6 +39,10 @@ func init() {
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
 		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+            // Releases after 3745 don't ship torcx at all (these use sysext)
+            if ( 0 < version.Compare(semver.Version{Major: 3745}) ) {
+                return true
+            }
 			// LTS (3033) does not have the network-kargs service pulled in:
 			// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
 			// Let's skip this test for < 3034 on ESX.
--- a/kola/tests/sysext/disable_containerd.go
+++ b/kola/tests/sysext/disable_containerd.go
@@ -1,4 +1,4 @@
-// Copyright 2017 CoreOS, Inc.
+// Copyright 2023 The Flatcar Maintainers.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


kola/tests/docker/torcx_manifest_pkgs.go
```diff
@@ -39,6 +39,10 @@ func init() {
 		// This test is normally not related to the cloud environment
 		Platforms: []string{"qemu", "qemu-unpriv"},
 		SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
+			// Releases after 3745 don't ship torcx at all (these use sysext)
+			if (0 < version.Compare(semver.Version{Major: 3745})) {
```
pothos, line 43: I think instead of a skipfunc we can use `EndVersion: semver.Version{Major: 3745},` here. If you run the test directly by name this gets ignored but normally it's evaluated.

kola/tests/sysext/disable_containerd.go
```diff
@@ -0,0 +1,53 @@
...
+		Name:        "sysext.disable-containerd",
+		// Only releases after 3745 ship sysext
+		MinVersion: semver.Version{Major: 3746},
+		UserData: conf.Butane(`
+variant: flatcar
+version: 1.0.0
+storage:
+  links:
+  - path: /etc/extensions/containerd-flatcar.raw
+    target: /dev/null
```
pothos, line 38: When we disable containerd but don't ship our own we should also disable Docker or it fill show up as failed in the logs.

kola/tests/sysext/disable_containerd.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+}
+
+func containerdDisable(c cluster.TestCluster) {
+	m := c.Machines()[0]
+	output := c.MustSSH(m,
+		`test ! -f /usr/bin/containerd || echo "ERROR"`)
+	if string(output) == "ERROR" {
+		c.Errorf("/usr/bin/containerd exists even when sysext is disabled!")
+	}
```
pothos, line 52: The test would be more robust by asserting what we expect instead of what we don't expect.
I think the easiest way would be:
```suggestion
	# Assert that the containerd binary does not exist
	_ = c.MustSSH(m, `test ! -f /usr/bin/containerd`)
```

kola/tests/sysext/disable_docker.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 CoreOS, Inc.
+// Copyright 2023 The Flatcar Maintainers.
```
pothos, line 1: Let's keep both lines

t-lo: Git wrongly assumes I've copied and modified that file from `torcx/torcx.go`. That's incorrect, I copied `torcx.go` to `enable_docker.go` and created this file - `disable_docker.go` from scratch.

It would not make any sense for CoreOS to have written anything of this test as it is purely sysext specific.

kola/tests/docker/torcx_manifest_pkgs.go
t-lo: Agreed, will change. Coincidentally this was the first file I touched when starting to work on sysext testing; didn't know at the time that we had this cool `MinVersion` / `EndVersion` mechanism.

kola/tests/sysext/disable_containerd.go
t-lo: It wouldn't as docker is socket activated and nothing uses docker at start-up. It's a good idea anyway because that code could serve as an accidental example. Will update.

kola/tests/sysext/disable_containerd.go
t-lo: I disagree - the suggestion makes the command fail instead of creating a meaningful test outcome. This would imply the test itself is broken.

In general, tests should end in one of 3 ways:
1. The test succeeded, signalling the build is healthy (containerd is not present).
2. The test failed, signalling an issue with the build (containerd is present).
3. The test is broken, signalling an issue with the test suite (either "test" or "echo" are broken, or SSH failed.)

I want to communicate 2., not 3.

t-lo:
Addressed all feedback, thank you for your review @pothos !

kola/tests/sysext/disable_containerd.go
pothos: Yes, it makes sense to distinguish that. The test is still not robust because it checks for errors explicitly. What you can change is to use `&& echo SUCCESS` and `if string(output) != "SUCCESS" {` to handle more things that can get wrong.

kola/tests/systemd/journald.go
```diff
@@ -110,7 +113,25 @@ storage:
 }
 
 func journalUser(c cluster.TestCluster) {
-	c.MustSSH(c.Machines()[0], "journalctl --user --unit hello.service | grep Foo")
+	if err := util.Retry(10, 2*time.Second, func() error {
+		cmd := "journalctl --user"
+		log, e := c.SSH(c.Machines()[0], cmd)
+		if e != nil {
+			return fmt.Errorf("Did not get expexted log output from '%s': %v", cmd, e)
+		}
```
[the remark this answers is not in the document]
t-lo: Right, but wouldn't we overwrite the `err` variable created in [line 116](https://github.com/flatcar/mantle/blob/e8d9164b934a8575362a9a8f35d1ba09ed55587c/kola/tests/systemd/journald.go#L116)? So if the retry loop aborts, [line 133](https://github.com/flatcar/mantle/blob/e8d9164b934a8575362a9a8f35d1ba09ed55587c/kola/tests/systemd/journald.go#L133) would print the SSH error instead of whatever we returned from the retry loop?

kola/tests/sysext/disable_containerd.go
t-lo: I'll reverse the check and catch `test` returning 1 (if the binary is not present).

t-lo: Done, please have another look.

kola/tests/systemd/journald.go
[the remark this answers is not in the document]
t-lo: Done - resolving.

kola/tests/docker/torcx_manifest_pkgs.go
t-lo: Done - resolving.

kola/tests/sysext/disable_containerd.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2023 The Flatcar Maintainers.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
t-lo: Gotcha! Updated the two files.

kola/tests/sysext/disable_containerd.go
pothos: What I meant was that `== "ERROR"` is too narrow and the test will silently break if it returns `ERROR\n` or the empty string. Therefore it's better to use something like `!= "SUCCESS"`.

PR #470, opened 2023-10-25
sdk: Use new manifests location

krnowak:
This is a part of my effort to move manifest files to a saner location within the scripts repository.


krnowak:
CI with this PR passed, see https://github.com/flatcar/scripts/pull/1304.

PR #474, opened 2023-10-27
tests/docker: add devicemapper storage test

t-lo:
Add a smoke test for the docker devicemapper storage backend.

Flatcar ships docker with support for the devicemapper storage back-end but we don't test it regularly. This has almost lead to the docker 24 update silently shipping without that storage driver.

Fixes https://github.com/flatcar/Flatcar/issues/1222


kola/tests/docker/docker.go
```diff
@@ -187,6 +187,30 @@ systemd:
...
 	})
 
+	register.Register(&register.Test{
+		Run:         func(c cluster.TestCluster) { testDockerInfo("devicemapper", c) },
+		ClusterSize: 1,
+		Name:        "docker.devicemapper-storage",
+		// This test is normally not related to the cloud environment
+		Platforms: []string{"qemu", "qemu-unpriv"},
+		// Note: copied verbatim from https://github.com/coreos/docs/blob/master/os/mounting-storage.md#creating-and-mounting-a-btrfs-volume-file
+		// Added explicit btrfs driver selection because overlay2 is the default for btrfs FS in docker 23 and above
```
krnowak, line 197: The comment should rather mention devicemapper, not btrfs2, no?

krnowak (approved):

kola/tests/docker/docker.go
t-lo: That's correct - fixed.

PR #475, opened 2023-10-30
tests/update: Add test for docker btrfs storage driver

t-lo:
This change adds an update test for instances using the docker btrfs storage driver. This storage driver was deprecated in docker release 23 and will be removed in a future release.

Docker 23 and above support backwards compatibility if BTRFS storage exists and is populated.

This test verifies that instances with btrfs storage don't break on update.

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -55,6 +55,25 @@ func init() {
 		},
 	})
 	register.Register(&register.Test{
+		Name:        "cl.update.docker-btrfs-compat",
+		Run:         btrfs_compat,
+		ClusterSize: 1,
+		NativeFuncs: map[string]func() error{
+			"Omaha": Serve,
+		},
+		// This test is normally not related to the cloud environment
+		Platforms: []string{"qemu", "qemu-unpriv"},
+		// This test verifies preservation of storage driver "btrfs" for docker.
+		// Docker releases before v23 defaulted to btrfs if the docker can only run if the update payload to test is given.
+		// The image passed must also be a release lower than or equal to 3760.0.0
+		// because newer versions ship docker 24.
+		EndVersion: semver.Version{Major: 3760},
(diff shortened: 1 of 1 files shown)


kola/tests/update/update.go
```diff
@@ -54,6 +54,25 @@ func init() {
...
+		NativeFuncs: map[string]func() error{
+			"Omaha": Serve,
+		},
+		// This test is normally not related to the cloud environment
+		Platforms: []string{"qemu", "qemu-unpriv"},
+		// This test verifies preservation of storage driver "btrfs" for docker.
+		// Docker releases before v23 defaulted to btrfs if the docker can only run if the update payload to test is given.
+		// The image passed must also be a release lower than or equal to 3760.0.0
+		// because newer versions ship docker 24.
+		EndVersion: semver.Version{Major: 3760},
```
krnowak, line 70: I'm wondering if we should have a separate test that boots the old release (but new enough to be with docker 24) with a config that forces btrfs storage driver, pulls an image and creates a container, so the storage is used. Then drops the config and updates the OS to see if btrfs driver is still used. Similar to this test, but would make sure nothing breaks after a sequence of `[old OS with old docker using btrfs storage]` -> `[newer OS with new docker using inherited btrfs storage]` -> `[shiny new OS with new docker still using the inherited btrfs storage]` (your test checks the first transition, another would test the second transition). Such test would probably need to be minimum 3760.0.0. Or maybe we could drop the EndVersion and make this test work regardless of the docker version.

t-lo: I don't think an intermediate update to Docker 24 would make any difference since I don't think docker stores any intermediate data. The "newest" Docker would need to auto-detect the need to use the btrfs storage driver either way, no matter if it was upgraded from docker 20 or docker 24.

krnowak (approved):

PR #484, opened 2023-11-15
kola: add brightbox support

tormath1:
This PR adds Brightbox support to Mantle, in order to test Flatcar on this provider (nightly or before a release, let's discuss this later).

## How to use

We first need to create the image on Brightbox.

## Testing done

```
$ BRIGHTBOX_IMAGE_ID=$(ore brightbox --brightbox-client-id ... --brightbox-client-secret ... create-image --name flatcar-nigthly --url http://bincache.flatcar-linux.net/images/amd64/3791.0.0+nightly-20231117-2100/flatcar_production_openstack_image.img)
$ time kola --parallel 1 --platform brightbox --brightbox-client-id ... --brightbox-client-secret ... --brightbox-image ${BRIGHTBOX_IMAGE_ID} run '*'
PASS, output in _kola_temp/brightbox-2023-11-20-1508-31363

real	42m42.941s
user	0m0.039s
sys	0m0.024s
1..21
ok - docker.network
ok - cl.etcd-member.discovery
ok - coreos.ignition.resource.remote
ok - cl.internet
ok - cl.install.cloudinit
ok - cl.flannel.udp
ok - coreos.ignition.sethostname
ok - cl.network.initramfs.second-boot
ok - kubeadm.v1.26.5.flannel.cgroupv1.base
ok - cl.ignition.kargs
ok - coreos.ignition.security.tls
ok - cl.ignition.v2.btrfsroot
ok - coreos.ignition.once
ok - kubeadm.v1.28.1.flannel.base
ok - cl.flannel.vxlan
ok - cl.basic
ok - coreos.ignition.ssh.key
ok - coreos.ignition.resource.local
ok - kubeadm.v1.26.5.flannel.base
ok - cl.cloudinit.basic
ok - kubeadm.v1.27.2.flannel.base
```

NOTE:
* Fetching console output is not yet supported / implemented at the moment on the provider side

### DIFF
--- a/platform/api/brightbox/api.go
+++ b/platform/api/brightbox/api.go
@@ -250,3 +250,23 @@ func (a *API) RemoveCloudIPs(ctx context.Context) error {
 
 	return nil
 }
+
+// RemoveServers remove any left overs servers before running a test.
+func (a *API) RemoveServers(ctx context.Context) error {
+	servers, err := a.client.Servers(ctx)
+	if err != nil {
+		return fmt.Errorf("getting servers: %w", err)
+	}
+
+	for _, server := range servers {
+		if server.Status == serverstatus.Deleted {
+			continue
+		}
+
+		if err := a.DeleteServer(ctx, server.ID); err != nil {
+			return fmt.Errorf("deleting server: %w", err)
+		}
+	}
+
+	return nil
+}
--- a/cmd/ore/brightbox/remove-servers.go
+++ b/cmd/ore/brightbox/remove-servers.go
@@ -1,0 +1,31 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+package brightbox
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdRemoveServers = &cobra.Command{
+		Use:   "remove-servers",
+		Short: "Remove any remaining servers",
+		Long:  `Remove left overs server from previous garbage collection`,
+		RunE:  removeServers,
+	}
+)
+
+func init() {
+	Brightbox.AddCommand(cmdRemoveServers)
+}
+
+func removeServers(cmd *cobra.Command, args []string) error {
+	if err := API.RemoveServers(context.Background()); err != nil {
+		return fmt.Errorf("removing servers: %w", err)
+	}
+
+	return nil
+}


platform/machine/brightbox/machine.go
```diff
@@ -0,0 +1,117 @@
...
+		return bm.mach.Server.CloudIPs[0].PublicIPv4
+	}
+
+	return ""
+}
+
+func (bm *machine) PrivateIP() string {
+	// TODO: We can see a Private IP from the dasbhboard but I don't know how to get it
+	// from the API.
+	return ""
```
pothos, line 42: You can use the public IP addrs, too

platform/machine/brightbox/cluster.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var keyname string
+	if !bc.RuntimeConf().NoSSHKeyInMetadata {
+		keyname = bc.flight.Name()
+	}
+	instance, err := bc.flight.api.CreateServer(bc.vmname(), keyname, conf.String())
```
pothos, line 34:
```suggestion
	instance, err := bc.flight.api.CreateServer(bc.vmname(), conf.String())
```
It's ok to not support ssh keys in metadata (same as in VMware ESXi or qemu)

pothos:
For completeness we also need `ore` support to delete all VMs/images(/IPs?). That's good for a cleanup before kola starts when we don't have unlimited VM slots. It's also good to have a GC job via ore that removes unused resources.

cmd/ore/brightbox/gc.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdGC = &cobra.Command{
+		Use:   "gc",
+		Short: "GC resources in Brightbox",
+		Long:  `Delete instances created over the given duration ago`,
```
pothos, line 16:
```suggestion
		Long:  `Delete instances and images created over the given duration ago`,
```

kola/tests/kubeadm/kubeadm.go
```diff
@@ -236,10 +238,14 @@ func init() {
...
 					SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
-						// LTS (3033) does not have the network-kargs service pulled in:
+						// * LTS (3033) does not have the network-kargs service pulled in:
 						// https://github.com/flatcar/coreos-overlay/pull/1848/commits/9e04bc12c3c7eb38da05173dc0ff7beaefa13446
-						// Let's skip this test for < 3034 on ESX.
-						return version.LessThan(semver.Version{Major: 3034}) && platform == "esx"
+						// Let's skip this test for < 3034 on ESX
+						// * For Cilium Calico/CNI on Brightbox:
+						// unprocessable_entity: User data is too long (maximum is 16384 characters)
+						// Should be reenable once we switch to Butane provisioning.
```
pothos, line 246:
```suggestion
						// Should be reenabled once we switch to Butane provisioning because of internal compression.
```

platform/api/brightbox/api.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf("destroying server from API: %w", err)
+	}
+
+	return nil
+}
+
+func (a *API) GC(gracePeriod time.Duration) error {
+	threshold := time.Now().Add(-gracePeriod)
+	// TODO: CloudIP has no creation date for now.
+	// We can't safely delete "old" cloud IPs.
```
pothos, line 163: It would work to add an explicit removal of all cloud IPs as ore subcommand which we run before `kola run`.

platform/api/brightbox/api.go
```diff
@@ -0,0 +1,234 @@
...
+		URL:      URL,
+		Username: &defaultUsername,
+		Arch:     arch.X86_64,
+	})
+	if err != nil {
+		return "", fmt.Errorf("creating image from API: %w", err)
+	}
+
+	// It usually takes around 20 seconds to extract the image.
+	if err := util.WaitUntilReady(1*time.Minute, 5*time.Second, func() (bool, error) {
```
pothos, line 221:
```suggestion
	if err := util.WaitUntilReady(2*time.Minute, 5*time.Second, func() (bool, error) {
```
In case it might take longer it could help to relax the waiting time more

platform/machine/brightbox/cluster.go
```diff
@@ -0,0 +1,84 @@
...
+func (bc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
+	conf, err := bc.RenderUserData(userdata, map[string]string{
+		"$public_ipv4":  "${COREOS_OPENSTACK_IPV4_PUBLIC}",
+		"$private_ipv4": "${COREOS_OPENSTACK_IPV4_LOCAL}",
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var cloudIP string
```
pothos, line 30:
```suggestion
	var cloudIP string
	// Allocate a free cloudIP, this only works for low enough --parallel= values because "select" does not wait
```

platform/machine/brightbox/flight.go
```diff
@@ -0,0 +1,79 @@
...
+	base, err := platform.NewBaseFlight(opts.Options, Platform, ctplatform.OpenStackMetadata)
+	if err != nil {
+		return nil, fmt.Errorf("creating base flight: %w", err)
+	}
+
+	bf := &flight{
+		BaseFlight: base,
+		api:        api,
+		// Current CloudIPs limit is 5.
+		cloudIPs: make(chan string, 5),
```
pothos, line 45: This should be a cli flag

platform/machine/brightbox/machine.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if err := bm.cluster.flight.api.DeleteServer(bm.ID()); err != nil {
+		plog.Errorf("deleting server %v: %v", bm.ID(), err)
+	}
+
+	if cloudIP != "" {
+		plog.Infof("Adding Cloud IP to the pool: %s", cloudIP)
+		bm.cluster.flight.cloudIPs <- cloudIP
+	}
```
pothos, line 88: I would move this down to be the last action of the function.

platform/machine/brightbox/flight.go
pothos: Or: The use of "select default" means it won't wait anyway, I guess it would work to do 
```suggestion
		cloudIPs: make(chan string, 999),
```
because the value has no meaning.

platform/api/brightbox/api.go
```diff
@@ -0,0 +1,259 @@
...
+func (a *API) RemoveCloudIPs() error {
+	cloudIPs, err := a.client.CloudIPs(context.TODO())
+	if err != nil {
+		return fmt.Errorf("getting cloud IPs: %w", err)
+	}
+
+	for _, cloudIP := range cloudIPs {
+		// Do not remove a mapped cloud IP - otherwise, we can end up
+		// with a server without public IP.
+		if cloudIP.Status == cloudipstatus.Mapped {
```
pothos, line 249: I thought that makes sense to ensure that the kola run can use all available IP addrs. I guess we would need a "delete-vms" ore command to also run before this here in the Jenkins job.

pothos: Because the garbage collection comes too late when a job restarts after kola crashed.

pothos (approved):
Thanks, looks good - I think for robustness we need a way of removing all leftover VMs with ore. We could also use the ore gc subcommand with a duration of 0 as workaround (if we run it before creating our image).

pothos (approved):

PR #486, opened 2023-11-27
kubeadm/calico: use public IP

tormath1:
On Digital Ocean, there are two network interfaces (one public and one private) - by default, Calico / Kubernetes was autodetecting the private one but doing this, the `calico-node` daemon set was not ready because of bgp issue.

this was not an issue until we started to have pod-to-elsewhere connections (like with NFS) 

---

Locally tested with current stable:
```
$ cat _kola_temp/do-latest/test.tap
1..1
ok - kubeadm.v1.27.2.calico.base
```

### DIFF
--- a/kola/tests/kubeadm/templates.go
+++ b/kola/tests/kubeadm/templates.go
@@ -365,6 +365,11 @@ curl --retry-delay 1 \
 # kubelet config for both controller and worker
 cgroup=$(docker info | awk '/Cgroup Driver/ { print $3}')
 
+{{ if eq .Platform "do" }}
+systemctl start --quiet coreos-metadata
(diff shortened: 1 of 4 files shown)


pothos (approved):

kola/tests/kubeadm/templates.go
```diff
@@ -379,6 +384,10 @@ kind: InitConfiguration
 nodeRegistration:
   kubeletExtraArgs:
     volume-plugin-dir: "/opt/libexec/kubernetes/kubelet-plugins/volume/exec/"
+{{ if eq .Platform "do" }}
+    # On digital ocean, the node IP is often not reachable - let's use the public one.
```
pothos, line 388: What means "often"?

pothos (approved):

PR #488, opened 2023-11-30
kola: Extend test to cover recreation of files in /etc

pothos:
For files like /etc/resolv.conf we want it to be recreated because the tmpfiles rule would do this. For a systemd bug workaround we filtered out some rules but even if we add them back, we still want to ensure that not only the file will always be recreated but that it also is not frozen as upcopy because we prefer the overlay lowerdir as source for updates.
Extend the /etc overlayfs test to ensure recreation of the file and to check against upcopies.

## How to use

Will pass for fixed releases.

## Testing done

```
sudo ./kola run -d -k --keys --key ~/.ssh/id_rsa.pub --platform=qemu  --qemu-image /var/tmp/flatcar_production_image.bin.whiteouts cl.overlay.cleanup
```
passed (and fails for current releases)

### DIFF
--- a/kola/tests/misc/state.go
+++ b/kola/tests/misc/state.go
@@ -60,7 +60,8 @@ func init() {
 
 // Check that the overlay doesn't have unexpected upcopies, e.g., due to
 // systemd-tmpfiles recreating the files/dirs or similar. Also check
-// that duplicates get removed on reboot.
+// that duplicates get removed on reboot. We also want certain files
+// to be recreated, e.g., those with a tmpfile rule.
 func OverlayCleanup(c cluster.TestCluster) {
 	m := c.Machines()[0]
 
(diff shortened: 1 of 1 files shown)


kola/tests/misc/state.go
```diff
@@ -93,6 +97,7 @@ func OverlayCleanup(c cluster.TestCluster) {
 
 	_ = c.MustSSH(m, fmt.Sprintf(overlayCheck, "after reboot"))
 	_ = c.MustSSH(m, `if sudo test -e /etc/sssd/sssd.conf || test -e /etc/kexec.conf || test -e /etc/samba || test ! -e /etc/bash/hello || test ! -e /etc/bash/bashrc ; then echo "Deletion or modification lost: $_" ; exit 1; fi`)
+	_ = c.MustSSH(m, `if test ! -e /etc/resolv.conf ; then echo "Files with tmpfile rule not recreated: $_" ; exit 1; fi && if ! sudo unshare -m bash -c 'umount /etc && test ! -e /etc/resolv.conf'; then echo "File with tmpfile rule exists as upcopy: $_"; fi`)
```
krnowak, line 100: Should the second if also have `exit 1`?

pothos: Thanks

krnowak (approved):

PR #493, opened 2024-01-10
kola: add test for TPM- and Tang-based disk encryption

simoncampion:
# Add test for TPM- and Tang-based disk encryption

This PR adds four new tests, for root and non-root encryption using TPM and Tang. It accompanies PRs in the [bootengine](https://github.com/flatcar/bootengine/pull/83) and [scripts](https://github.com/flatcar/scripts/pull/1560) repositories.

## How to use

Run the new tests, e.g. with `kola run  -k -b cl -p qemu-unpriv --board amd64-usr --qemu-image $SOME_IMAGE cl.tpm.* cl.tang.*` in a Docker container. Note that the Docker image needs to be rebuilt because of the new dependency on `swtpm` for the TPM emulation with QEMU. If you run the tests natively, you'll also need to ensure that `swtpm` is in `$PATH`.

The four tests should fail for Flatcar images without TPM- and Tang-based disk encryption support and succeed for images with such support.

## Testing done

I ran the four new tests in Docker. On images built from the PR in the scripts repository, the nonroot tests pass. The root tests fail because the `systemd-cryptsetup@rootencrypted.service` fails on reboot, although the disks are decrypted properly. This is because the root disk has to be decrypted in the initramfs, causing the subsequent `systemd-cryptsetup` unit to fail. I will wait for the discussion on how to best fix this in the PR in the scripts repository.

```
root@saturn:/# kola run  -k -b cl -p qemu-unpriv --board amd64-usr --qemu-image /mnt/images/developer-3844.0.0+nightly-20240109-2100-22-g4dbe3e0505-a1_flatcar_production_qemu_image.img cl.tpm.* cl.tang.*                                                                                           
=== RUN   cl.tpm.root                                                                                                                              
=== RUN   cl.tang.nonroot                                                                                                                          
=== RUN   cl.tpm.nonroot                                                                                                                           
=== RUN   cl.tang.root                                                                                                                             
--- FAIL: cl.tpm.root (48.45s)                                                                                                                     
        tpm.go:176: could not reboot machine: machine "5cb8b0b9-059b-4ed9-8255-9e30151e0b4f" failed basic checks: some systemd units failed:       
● systemd-cryptsetup@rootencrypted.service loaded failed failed Cryptography Setup for rootencrypted                                               
status:                                                                                                                                                                                                                                                                                               
journal:-- No entries --                                                                                                                           
--- FAIL: cl.tang.root (48.50s)                                                                                                                    
        tang.go:217: could not reboot machine: machine "9f1b1acd-7f49-4296-ab38-5244784765f4" failed basic checks: some systemd units failed:                                                                                                                                                         
● systemd-cryptsetup@rootencrypted.service loaded failed failed Cryptography Setup for rootencrypted                                                                                                                                                                                                  
status:                                                                                                                                                                                                                                                                                               
journal:-- No entries --                                                                                                                                                                                                                                                                              
--- PASS: cl.tpm.nonroot (48.63s)                                                                                                                  
--- PASS: cl.tang.nonroot (48.73s)                                                                                                                 
FAIL, output in _kola_temp/qemu-unpriv-2024-01-10-2004-294                                                                                         
harness: test suite failed  
```

<!-- For coreos-overlay ebuild modifications that include a CROS_WORKON_COMMIT bump, did you bump too the ebuild revision ? -->


kola/tests/misc/tang.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/flatcar/mantle/platform/machine/qemu"
+	"github.com/flatcar/mantle/platform/machine/unprivqemu"
+)
+
+const (
+	// In the Ignition config below, we create a partition labeled ROOT on a second disk.
+	// As advised in the Flatcar documentation, we then remove the ROOT label from the existing
+	// root partition, which is vda9 in the QEMU disk image.
+	IgnitionConfigRootTang = `{
+		"ignition": {
```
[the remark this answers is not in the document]
simoncampion: Good point!

My thinking was that it would be best to first merge clevis support and after that open a PR to update the Butane spec, to avoid confusion. But once the Butane spec is updated, I'd be happy to bump the `butane` Go dependency in `mantle` and rewrite the config here. We can either hold off the merge of this PR until then, or I can open a second PR later. I don't have a preference either way; let me know what option you think is best.

kola/tests/misc/tang.go
```diff
@@ -0,0 +1,300 @@
...
+					"format": "ext4",
+					"label": "ROOT"
+				}
+			]
+		},
+		"systemd": {
+			"units": [{
+			"name": "remove-root-label.service",
+			"enabled": true,
+			"contents": "[Service]\nType=oneshot\nExecStart=wipefs -a /dev/vda9\n[Install]\nWantedBy=multi-user.target"
```
pothos, line 77: Isn't it possible to reformat this partition with Ignition? I would prefer that because otherwise it's not clear which rootfs is used in the initrd in the first run.

kola/tests/misc/tang.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+func init() {
+	tangPort := 8007
+	// TODO: I let the Tang server bind to an IP of a networking interface in the root networking namespace. This works because
+	// traffic from within the VM to that IP will be routed to the host. However, this approach introduces an unappealing dependency
+	// between this test and how the network is setup for the VMs, and also it might bind Tang to an interface with a public IP.
+	// An alternative approach would be to add another TAP interface to the bridge and let the Tang server bind there, but that would
+	// require the Tang setup to happen outside of these tests and introducing more complexity in different parts of the code base.
+	// I'll decide whether to rewrite this or leave it as it is based on feedback by reviewers.
```
pothos, line 142: With `Platforms:   []string{"qemu"},` that seems ok

kola/tests/misc/tpm.go
```diff
@@ -0,0 +1,223 @@
...
+					"format": "ext4",
+					"label": "ROOT"
+				}
+			]
+		},
+		"systemd": {
+			"units": [{
+			"name": "remove-root-label.service",
+			"enabled": true,
+			"contents": "[Service]\nType=oneshot\nExecStart=wipefs -a /dev/vda9\n[Install]\nWantedBy=multi-user.target"
```
pothos, line 64: (same comment as above)

kola/tests/misc/tang.go
pothos: I think some filesystems entry like that could work, here in butane yaml:

```
variant: flatcar
version: 1.0.0
storage:
  filesystems:
  - device: /dev/disk/by-partlabel/ROOT
    wipe_filesystem: true
    format: none
```
Ignition json:
```
{
  "ignition": {
    "version": "3.3.0"
  },
  "storage": {
    "filesystems": [
      {
        "device": "/dev/disk/by-partlabel/ROOT",
        "format": "none",
        "wipeFilesystem": true
      }
    ]
  }
}

```

simoncampion: Yes, that works. I changed it as you suggested.

kola/tests/misc/tang.go
simoncampion: Okay, I removed the `TODO` comment.

simoncampion:
I added the kernel arguments to the tests that are now required to enable clevis unlock in the initramfs (see the thread [here](https://github.com/flatcar/bootengine/pull/83#discussion_r1504383584)).

I also solved the issue with tests failing because `systemd-cryptsetup@rootencrypted.service` in late userspace fails as the disk is already unlocked from early userspace by masking `systemd-cryptsetup@rootencrypted.service` .

kola/tests/misc/tpm.go
```diff
@@ -0,0 +1,222 @@
...
+					"uuid": "12345678-9abc-def0-1234-56789abcdef0",
+					"clevis": {
+						"tpm2": true
+					}
+				}
+			],
+			"filesystems": [
+				{
+					"device": "/dev/disk/by-partlabel/ROOT",
+					"format": "none",
```
pothos, line 56: What also works is to use this partition as luks partition (the `luks:` section need the entry to wipe the existing content), then we could drop the secondary disk. I think this is even more interesting to test.

kola/tests/misc/tpm.go
```diff
@@ -0,0 +1,221 @@
...
+func init() {
+	runRootTPM := func(c cluster.TestCluster) {
+		tpmTest(c, conf.Ignition(IgnitionConfigRootTPM), "/")
+	}
+	register.Register(&register.Test{
+		Run:         runRootTPM,
+		ClusterSize: 0,
+		Platforms:   []string{"qemu"},
+		Name:        "cl.tpm.root",
+		Distros:     []string{"cl"},
```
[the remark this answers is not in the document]
pothos: Assuming we merge the scripts PR before we branch the next Alpha you can write "Major: 3880" which does not exist but is higher than the current Alpha's major number
(also needed for the other tests)

kola/tests/misc/tpm.go
pothos: By the way, did you see the `needsNetwork` entry for clevis? I wonder if that triggers the network target in the initrd, didn't try it yet…

simoncampion: > What also works is to use this partition as luks partition (the luks: section needs a wipeVolume entry to wipe the existing content), then we could drop the secondary disk. I think this is even more interesting to test.

Sounds good. I changed the cl.tang.root and cl.tpm.root to test encryption of the ROOT partition on the primary disk.

> By the way, did you see the custom: \n needsNetwork: entry for clevis? I wonder if that triggers the network target in the initrd, didn't try it yet…

Maybe you've already seen this, but once you set `custom.needsNetwork`, you also need to set `custom.pin` and `custom.config`. So, I'm not sure whether that would be a good alternative to how we currently do it in the tests, setting the `rd.networkd=1` cmdline parameter.

kola/tests/misc/tpm.go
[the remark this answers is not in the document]
simoncampion: Makes sense. Given that the scripts PR is merged now, I added "Major: 3880".

pothos (approved):

kola/tests/misc/tpm.go
pothos: Ok, didn't know that - and I also think that it might not help much because on second boot the initrd wouldn't know about this setting if it ends up in crypttab.

pothos:
Oh, the Go compiler is not yet happy about `net.FlagRunning` not being defined

simoncampion:
> Oh, the Go compiler is not yet happy about `net.FlagRunning` not being defined

That symbol was introduced in [Go 1.20](https://tip.golang.org/doc/go1.20), but the pipeline [used Go 1.19](https://github.com/flatcar/mantle/actions/runs/8280577973/job/22657809455). I suppose the pipeline sees that the `go.mod` file only requires Go 1.19 and then runs that Go version. This issue does not arise when testing locally with the dockerized build because the `Dockerfile` uses `docker.io/amd64/golang:1.21-bookworm` as its base.

I bumped the Go version in `go.mod` to Go 1.20. If that's undesirable, I could alternatively rewrite the code to be Go 1.19-compatible.

pothos:
Thanks, that should work then

PR #494, opened 2024-01-17
brightbox: increase timeout in image creation

tormath1:
it's more frequent to have timeouts when the image is pulled from bincache (and not from origin servers) - a similar behavior can be seen on digital ocean.

### DIFF
--- a/platform/api/brightbox/api.go
+++ b/platform/api/brightbox/api.go
@@ -220,7 +220,7 @@ func (a *API) UploadImage(ctx context.Context, name, URL string) (string, error)
 	}
 
 	// It usually takes around 20 seconds to extract the image.
-	if err := util.WaitUntilReady(2*time.Minute, 5*time.Second, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


platform/api/brightbox/api.go
```diff
@@ -220,7 +220,7 @@ func (a *API) UploadImage(ctx context.Context, name, URL string) (string, error)
 	}
 
 	// It usually takes around 20 seconds to extract the image.
-	if err := util.WaitUntilReady(2*time.Minute, 5*time.Second, func() (bool, error) {
+	if err := util.WaitUntilReady(4*time.Minute, 5*time.Second, func() (bool, error) {
```
pothos, line 223:
```suggestion
	if err := util.WaitUntilReady(5*time.Minute, 5*time.Second, func() (bool, error) {
```
How about 5 or even 10 to make sure we don't hit this occasionally?

pothos (approved):

sayanchowdhury (approved):

PR #497, opened 2024-02-07
azure: Add support for using an external availability set and resource group

jepio:
An alternative approach to reusing a resource group: only store VM object in the passed resource group and all other objects go in the kola managed RG. Instances already get deleted separately from rgs at the end of the run.

This is needed to support availability set that may be pinned to a specific compute resource. VMs that are part of an availability set must share a resource group with the availability set and must share a vnet.


jepio:
This is not meant to be merged - i just needed the CI to build a container image

jepio:
Actually i think we do - I use this regularly. Do you mind if I rebase, drop the reverts and resubmit?

jepio:
@tormath1 would you review?

PR #504, opened 2024-03-13
Add kola tests for zfs

jepio:
These tests cover multiple use cases:
- fetching the sysext from the release server
- systemd units work and datasets get mounted
- docker zfs driver works
- setting up network shares works

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -513,7 +513,7 @@ func getClusterSemver(flight platform.Flight, outputDir string) (*semver.Version
 		return nil, fmt.Errorf("creating new machine for semver check: %v", err)
 	}
 
-	out, stderr, err := m.SSH("grep ^VERSION_ID= /etc/os-release")
+	out, stderr, err := m.SSH("grep ^VERSION= /etc/os-release")
 	if err != nil {
 		return nil, fmt.Errorf("parsing /etc/os-release for VERSION_ID: %v: %s", err, stderr)
 	}
--- a/kola/tests/sysext/zfs.go
+++ b/kola/tests/sysext/zfs.go
@@ -1,0 +1,214 @@
+// Copyright The Mantle Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package sysext
+
+import (
+	"fmt"
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+	"github.com/flatcar/mantle/platform"
+	"github.com/flatcar/mantle/platform/conf"
+	"github.com/flatcar/mantle/platform/machine/qemu"
+	"github.com/flatcar/mantle/platform/machine/unprivqemu"
+)
+
+var zfsUserData = conf.Butane(`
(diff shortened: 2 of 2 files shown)


kola/tests/sysext/zfs.go
```diff
@@ -0,0 +1,214 @@
...
+		Run:         checkZfsNfs,
+		ClusterSize: 0,
+		Distros:     []string{"cl"},
+		// This test is normally not related to the cloud environment
+		Platforms:  []string{"qemu", "qemu-unpriv"},
+		MinVersion: semver.Version{Major: 3902},
+	})
+}
+
+func createZfsMachine(c cluster.TestCluster, userdata *conf.UserData) platform.Machine {
```
pothos, line 157: Either here or in each top function I would add a `c.Skip` call to skip these tests on GitHub Action runs because they won't be able to download the extension. We can identify this condition by `reply.StatusCode != 200` for `reply, err := client.Head(fmt.Sprintf("https://bincache.flatcar-linux.net/images/%s/%s/version.txt", arch, version))`

pothos: Ah, and we have to skips this for releases tests because the release can't download from the release server.
I guess it would be better to move the skipping to a temporary extra machine to first run `keySum := string(c.MustSSH(m, `md5sum /usr/share/update_engine/update-payload-key.pub.pem | cut -d " " -f 1`))` and skip `if keySum != "7192addf4a7f890c0057d21653eff2ea"`

Code extracts are from the end of `/kola/testsupdate/update.go`

pothos: Thinking about a solution: Maybe we could inject a flag file (or kernel parameter) to the image which would tell the initrd to bind-mount the dev key instead of the prod update key, so that it would then be able to download from bincache and verify that payload.

pothos: > I guess it would be better to move the skipping to a temporary extra machine to first run

To reduce the impact of booting a helper machine we could organize the tests under one main function

jepio: Oh this is getting messy.

I had added a nice clean skip func implementation for GHA: https://github.com/flatcar/mantle/pull/504/commits/7d90c84c4d97a9064532950c00c3ec1cfcf5b9ee. 
But i didn't think that release tests wouldn't work.

I'll probably need to put this on hold.

jepio:
@pothos I've added a check that makes this only run for dev builds/nightlies. Better that we have some testing before we figure out how to get zfs tested for releases (before they're released).

kola/harness.go
```diff
@@ -513,7 +513,7 @@ func getClusterSemver(flight platform.Flight, outputDir string) (*semver.Version
 		return nil, fmt.Errorf("creating new machine for semver check: %v", err)
 	}
 
-	out, stderr, err := m.SSH("grep ^VERSION_ID= /etc/os-release")
+	out, stderr, err := m.SSH("grep ^VERSION= /etc/os-release")
```
pothos, line 516: I hope this won't break how the version of maintenance nightlies gets processed.

pothos (approved):

PR #507, opened 2024-03-14
Fix test script for Go >=1.20

jepio:
go test -i is deprecated and has been removed. It no longer serves a purpose. See: https://github.com/golang/go/issues/41696

### DIFF
--- a/test
+++ b/test
@@ -24,8 +24,7 @@ trap "rm -rf _testbin/" EXIT
 pkgs=$(go list $PKG | grep -v /vendor/)
 src=$(find . -name '*.go' -not -path "./vendor/*")
 
-echo "Building tests..."
-go test -mod=vendor -i "$@" $pkgs
+echo "Building"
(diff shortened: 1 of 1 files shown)


pothos (approved):

test
```diff
@@ -24,10 +24,6 @@ trap "rm -rf _testbin/" EXIT
 pkgs=$(go list $PKG | grep -v /vendor/)
 src=$(find . -name '*.go' -not -path "./vendor/*")
 
-echo "Building tests..."
-go test -mod=vendor -i "$@" $pkgs
-go install -mod=vendor $pkgs
```
pothos, line 29: Should we keep this? Otherwise we need to also remove the `Running commands...` section

jepio: yes bringing it back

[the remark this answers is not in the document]
jepio: ack

pothos (approved):

PR #509, opened 2024-03-15
kola/tests/tang: Ignore interfaces with no IPs

jepio:
We use Equinix Metal machines in our CI that have bonded network interfaces, which means we end up with nics with no ips. The tang setup code needs to handle this. We will also want to rework the code to provide tang in the same way as we do for etcd.

### DIFF
--- a/kola/tests/devcontainer/devcontainer.go
+++ b/kola/tests/devcontainer/devcontainer.go
@@ -54,6 +54,7 @@ if [[ "${EXPECTED_VERSION}" != "${FLATCAR_RELEASE_VERSION}" ]]; then
 fi
 
 export PORTAGE_BINHOST="${PORTAGE_BINHOST}"
+export FEATURES="-ipc-sandbox -network-sandbox"
 emerge-gitclone
 emerge --getbinpkg --verbose coreos-sources
 zcat /proc/config.gz >/usr/src/linux/.config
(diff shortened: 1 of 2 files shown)


pothos:
For arm64 I prepared a PR: https://github.com/flatcar/scripts/pull/1755

simoncampion:
Thanks for fixing this, @jepio !

Regarding the failing service: I should've run the tests on arm64 locally, then I would have caught the issue before the merge. Sorry about that, I completely forgot about arm64.

I think the issue is that we need the `rd.luks.name=...` parameter on the second boot but not on the first boot. We need it on the second (and all subsequent) boots because it triggers disk unlocking in the initrd for the root disk. But, as you say, it looks like it creates a race condition on the first boot, with the systemd-cryptsetup unlocking (sometimes?) failing. There is no reason why systemd-cryptsetup should be running during first boot, since Ignition takes care of opening the newly created LUKS device.

We could fix this issue by changing the Ignition config in the tests so that the `rd.luks.name=...` parameter is passed only on the second boot. I'm not sure what's the best way to achieve this would be---maybe adding a systemd unit that modifies the grub.cfg. Alternatively, we can remove the `rd.luks.name=...` parameters and implement a generator in `bootengine` that doesn't do anything on first boot, but on subsequent boot runs systemd-cryptsetup-generator to generate a systemd-cryptsetup unit for unlocking the root disk. We had discussed this as a follow-up improvement, but it might be more urgent now than initially thought. The generator solution would probably be a lot cleaner and also has other advantages. I'll try to implement and test it soon (see [this draft PR](https://github.com/flatcar/bootengine/pull/90)).

Regarding the masking: Yep, the masking might be unnecessary now. I am relatively sure it doesn't make the tests fail at the moment though, so I'd suggest we get the arm64 tests to pass first and then I'll check whether tests still pass without the masking and open a PR if they do.

PR #510, opened 2024-03-20
Fix tang test (part 2)

jepio:
Fix the tang test by moving the tang listener into the qemu network namespace.

Also an attempt to speed up container build so the second commit will be pushed after the first build completes.

### DIFF
--- a/platform/local/cluster.go
+++ b/platform/local/cluster.go
@@ -93,6 +93,20 @@ func (lc *LocalCluster) GetOmahaHostPort() (string, error) {
 	return net.JoinHostPort(lc.hostIP(), port), nil
 }
 
+func (lc *LocalCluster) NewListenerInsideClusterNS() (*net.Listener, error) {
+	nsExit, err := ns.Enter(lc.flight.nshandle)
+	if err != nil {
+		return nil, fmt.Errorf("failed to enter cluster ns: %w", err)
+	}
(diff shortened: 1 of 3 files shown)


platform/local/cluster.go
```diff
@@ -93,6 +93,20 @@ func (lc *LocalCluster) GetOmahaHostPort() (string, error) {
 	return net.JoinHostPort(lc.hostIP(), port), nil
 }
 
+func (lc *LocalCluster) NewTangListener() (*net.Listener, error) {
```
krnowak, line 96: This does not seem to be specific at all to Tang, so maybe name it `NewListenerInsideClusterNS`?

jepio: makes sense - pushed the change

jepio:
I ran a test with this PRs container image here: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/20754/console

krnowak:
> I ran a test with this PRs container image here: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/20754/console

Are they still failing?

```
[2024-03-20T10:54:32.376Z] Failed tests:
[2024-03-20T10:54:32.376Z] cl.tang.root
[2024-03-20T10:54:32.376Z] cl.tpm.root
```

PR #511, opened 2024-03-24
kola: remove cmdline parameters for Tang and TPM tests

simoncampion:
# Remove kernel cmdline parameters for Tang and TPM tests

Removes `rd.luks...` cmdline parameters that are unnecessary once [this PR](https://github.com/flatcar/bootengine/pull/90) in flatcar/bootengine is merged. Removing the cmdline parameters should also fix an issue with tests failing because of failing superfluous `systemd-cryptsetup` units during the first boot.

## How to use

Run the `cl.tpm.*` and `cl.tang.*` tests.

## Testing done

See the description of testing done in [this PR](https://github.com/flatcar/bootengine/pull/90).


kola/tests/misc/tang.go
```diff
@@ -29,14 +29,13 @@ const (
 			"version": "3.3.0"
 		},
 		"kernelArguments": {
-			"shouldExist": ["rd.luks.name=12345678-9abc-def0-1234-56789abcdef0=rootencrypted", "rd.networkd=1", "systemd.mask=systemd-cryptsetup@rootencrypted.service"]
+			"shouldExist": ["rd.networkd=1", "systemd.mask=systemd-cryptsetup@rootencrypted.service"]
```
pothos, line 32:
```suggestion
			"shouldExist": ["rd.networkd=1"]
```
The masking can also go, or? I think it didn't work and if it would then unlocking wouldn't work.

kola/tests/misc/tpm.go
```diff
@@ -25,14 +25,13 @@ const (
 			"version": "3.3.0"
 		},
 		"kernelArguments": {
-			"shouldExist": ["rd.luks.name=12345678-9abc-def0-1234-56789abcdef0=rootencrypted", "systemd.mask=systemd-cryptsetup@rootencrypted.service"]
+			"shouldExist": ["systemd.mask=systemd-cryptsetup@rootencrypted.service"]
```
pothos, line 28: With this removed the whole kernelArguments section wouldn't be needed and we can skip the reboot that Ignition does.

pothos:
Thanks! When we backport your bootengine changes this will even work for the old Alpha, but maybe we should still bump the minimum image version for the test to be ` 3913.0.1` (should work for all cases, a bugfix Alpha, a new Alpha, or Beta promotion).

kola/tests/misc/tang.go
simoncampion: You are right, the masking can go. I removed it.

Aside:

> if it would then unlocking wouldn't work

`systemd.mask` only masks units during main boot, as opposed to `rd.systemd.mask`, which masks units in initramfs ([docs](https://www.freedesktop.org/software/systemd/man/latest/systemd-debug-generator.html#Description)). Thus, the `systemd-cryptsetup@rootencrypted` in initramfs wouldn't be masked by this cmdline parameter, and unlocking in initramfs would work. I thought I had seen that systemd creates a second `systemd-cryptsetup@rootencrypted` for the main boot, too, because Ignition added an entry for the root partition to `/etc/crypttab`, and that this second unit failed because the root partition was already unlocked during the initramfs. But I tested it and apparently, that's not the case. So, the masking is superfluous, as you suggested.

kola/tests/misc/tpm.go
simoncampion: Yup, it's removed now.

simoncampion:
> maybe we should still bump the minimum image version for the test to be  3913.0.1

Sounds good, I bumped the min version to 3913.0.1.

kola/tests/misc/tang.go
pothos: Thanks for checking this, glad we can get rid of it :)

pothos (approved):

PR #512, opened 2024-03-26
fix check for systemd-timesyncd.service

jepio:
systemd-timesyncd.service will not be the only ntp implementation enabled by default on all platforms, so the test needs to take that into account.

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -322,27 +322,38 @@ func TestServicesActive() error {
 	return servicesActive([]string{
 		"multi-user.target",
 		"docker.socket",
-		"systemd-timesyncd.service",
 		"update-engine.service",
+	}, []string{
+		"systemd-timesyncd.service",
+		"chronyd.service",
+		"ntpd.service",
 	})
 }
 
 func TestServicesActiveCoreOS() error {
 	return servicesActive([]string{
 		"multi-user.target",
-	})
+	}, []string{})
 }
 
-func servicesActive(units []string) error {
+func servicesActive(allOf []string, anyOf []string) error {
 	//t.Parallel()
-	for _, unit := range units {
+	for _, unit := range allOf {
 		c := exec.Command("systemctl", "is-active", unit)
 		err := c.Run()
 		if err != nil {
-			return fmt.Errorf("Services Active: %v", err)
+			return fmt.Errorf("services Active: %s: %v", unit, err)
 		}
 	}
-	return nil
+	var err error
+	for _, unit := range anyOf {
+		c := exec.Command("systemctl", "is-active", unit)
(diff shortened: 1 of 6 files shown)


kola/tests/coretest/core.go
```diff
@@ -322,26 +322,41 @@ func TestServicesActive() error {
...
-	for _, unit := range units {
+	for _, unit := range allOf {
 		c := exec.Command("systemctl", "is-active", unit)
 		err := c.Run()
 		if err != nil {
-			return fmt.Errorf("Services Active: %v", err)
+			return fmt.Errorf("services Active: %s: %v", unit, err)
+		}
+	}
+	anyActive := false
```
pothos, line 348:
```suggestion
	anyActive := false
	var err error
```
How about defining the error here to be able to report it below?

kola/tests/coretest/core.go
```diff
@@ -322,26 +322,41 @@ func TestServicesActive() error {
...
+	for _, unit := range anyOf {
+		c := exec.Command("systemctl", "is-active", unit)
+		err := c.Run()
```
pothos, line 351:
```suggestion
		// Note: Don't shadow the error here, i.e., don't use err :=
		err = c.Run()
```

jepio: we expect some of these to fail :) the ntp services can't all be running at the same time

kola/tests/coretest/core.go
```diff
@@ -322,26 +322,41 @@ func TestServicesActive() error {
...
+		if err == nil {
+			anyActive = true
+			break
 		}
 	}
+	if !anyActive {
+		return fmt.Errorf("no NTP service active")
```
pothos, line 358:
```suggestion
		return fmt.Errorf("no NTP service active (error: %v)", err)
```
That also works with a nil error shown as `<nil>` but I think we wouldn't run into this

kola/tests/coretest/core.go
jepio: alright

kola/tests/coretest/core.go
```diff
@@ -322,26 +322,41 @@ func TestServicesActive() error {
...
+			return fmt.Errorf("services Active: %s: %v", unit, err)
+		}
+	}
+	anyActive := false
+	for _, unit := range anyOf {
+		c := exec.Command("systemctl", "is-active", unit)
+		err := c.Run()
+		if err == nil {
+			anyActive = true
+			break
```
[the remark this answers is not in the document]
jepio: ok

pothos (approved):

jepio:
Test ran successfully. http://jenkins.infra.kinvolk.io:8080/job/container/job/test/20926/

PR #515, opened 2024-04-03
kola: Increase amount of disk corruption in verity test

pothos:
For a particular image build the filesystem seemed to be more resistant to the corruption test, be it due to the corruption hitting other structures or some non-flushable cache covering more of the corrupted area. One can still trigger the verity panic by overwriting 100 MB instead of 10 MB.
Increase the amount of zeros written by the corruption test.

## Testing done

Tested with the amd64 image here https://bincache.flatcar-linux.net/images/amd64/9999.9.9+kai-remove-acbuild/flatcar_production_image.bin.bz2 which was able to avoid the direct panic on the 10 MB corruption due to caching effects.

### DIFF
--- a/kola/tests/misc/verity.go
+++ b/kola/tests/misc/verity.go
@@ -96,8 +96,8 @@ func VerityCorruption(c cluster.TestCluster) {
 	// get usr device, probably vda3
 	usrdev := util.GetUsrDeviceNode(c, m)
 
-	// write zero bytes to first 10 MB
-	c.MustSSH(m, fmt.Sprintf(`sudo dd if=/dev/zero of=%s bs=1M count=10 status=none`, usrdev))
+	// write zero bytes to first 100 MB
+	c.MustSSH(m, fmt.Sprintf(`sudo dd if=/dev/zero of=%s bs=1M count=100 status=none`, usrdev))
 
 	// make sure we flush everything so the filesystem has to go through to the device backing verity before fetching a file from /usr
 	// (done in one execution because after flushing command itself runs the corruption could already be detected,


kola/tests/misc/verity.go
```diff
@@ -96,8 +96,8 @@ func VerityCorruption(c cluster.TestCluster) {
 	// get usr device, probably vda3
 	usrdev := util.GetUsrDeviceNode(c, m)
 
-	// write zero bytes to first 10 MB
-	c.MustSSH(m, fmt.Sprintf(`sudo dd if=/dev/zero of=%s bs=1M count=10 status=none`, usrdev))
+	// write zero bytes to first 100 MB
+	c.MustSSH(m, fmt.Sprintf(`sudo dd if=/dev/zero of=%s bs=1M count=100 status=none`, usrdev))
```
ader1990, line 100: the flush can be done directly by dd `oflag=dsync`

ader1990: On a decent NVME:

```
$: time dd if=/dev/zero of=t bs=1M count=100 status=none oflag=dsync

real    0m0.114s
user    0m0.000s
sys     0m0.057s

$: time dd if=/dev/zero of=t bs=1M count=100 status=none

real    0m0.057s
user    0m0.000s
sys     0m0.053s
```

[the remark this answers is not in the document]
pothos: Yes, but only when it gets read and even with the `ls` if the contents are cached, this is not the case

jepio: yes

pothos: Merging the commands would also work when combined with `&&` to keep the error checking for the first one. When settings flags for `dd` to remove the `sync` we would just have to do enough tests to make sure that the test doesn't become flaky, or we could also keep the extra `sync` just to be sure.

PR #521, opened 2024-04-10
kola: Add tests for encrypted root disks with TPM PCR binding

pothos:
As documented in https://github.com/flatcar/flatcar-website/pull/317 we can use PCR binding in Flatcar with some limitations and workarounds. I think we should be able to get rid of the first boot PCR difference by handling the first-boot flag detection in userspace and if we can guarantee that the outcome of setting the first boot or not is the same, i.e., we would have to always measure the effective Ignition config. For the rebinding on update the best bet we have is to create a Flatcar variant with sd-boot for signed PCR policies. Anyway, these are future topics and it's already good that we can make some encryption setups work.

## Testing done

Tested both new tests with the image from https://github.com/flatcar/scripts/pull/1861

### DIFF
--- a/kola/tests/misc/tpm.go
+++ b/kola/tests/misc/tpm.go
@@ -31,9 +31,9 @@ variant: flatcar
 version: 1.0.0
 storage:
   luks:
-  - name: rootencrypted
-    wipe_volume: true
-    device: "/dev/disk/by-partlabel/ROOT"
+    - name: rootencrypted
+      wipe_volume: true
+      device: "/dev/disk/by-partlabel/ROOT"
   filesystems:
     - device: /dev/mapper/rootencrypted
       format: ext4
@@ -69,9 +69,9 @@ storage:
         inline: |
           SERVER=disabled
   luks:
-  - name: rootencrypted
-    wipe_volume: true
-    device: "/dev/disk/by-partlabel/ROOT"
+    - name: rootencrypted
+      wipe_volume: true
+      device: "/dev/disk/by-partlabel/ROOT"
   filesystems:
     - device: /dev/mapper/rootencrypted
       format: ext4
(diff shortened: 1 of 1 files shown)


kola/tests/misc/tpm.go
```diff
@@ -47,6 +54,127 @@ systemd:
...
+  files:
+    - path: /etc/flatcar/update.conf
+      overwrite: true
+      contents:
+        inline: |
+          SERVER=disabled
+  luks:
+  - name: rootencrypted
+    wipe_volume: true
+    device: "/dev/disk/by-partlabel/ROOT"
```
krnowak, line 74: Do these lines ought to be indented further by two more spaces?

kola/tests/misc/tpm.go
```diff
@@ -47,6 +54,127 @@ systemd:
...
+          if [ -e /etc/luks/rootencrypted-bound ]; then
+            mv /etc/luks/rootencrypted-bound /etc/luks/rootencrypted-bind
+          fi
+          # But since a reboot inbetween could have bound it again,
+          # remove the PCR binding for every run
+          systemd-cryptenroll --tpm2-device=auto --unlock-key-file=/etc/luks/rootencrypted-bind --wipe-slot=tpm2 --tpm2-pcrs= /dev/disk/by-partlabel/ROOT
+  luks:
+  - name: rootencrypted
+    wipe_volume: true
+    device: "/dev/disk/by-partlabel/ROOT"
```
krnowak, line 138: Same here.

kola/tests/misc/tpm.go
pothos: Yes, can do it, YAML is weird

pothos: Fixed it for the existing test, too

krnowak (approved):

PR #522, opened 2024-04-10
qemu: extract swtpm

jepio:
So that kola spawn can support `--qemu-tpm` for testing.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -227,6 +227,7 @@ func init() {
 	sv(&kola.QEMUOptions.BIOSImage, "qemu-bios", "", "BIOS to use for QEMU vm")
 	bv(&kola.QEMUOptions.UseVanillaImage, "qemu-skip-mangle", false, "don't modify CL disk image to capture console log")
 	sv(&kola.QEMUOptions.ExtraBaseDiskSize, "qemu-grow-base-disk-by", "", "grow base disk by the given size in bytes, following optional 1024-based suffixes are allowed: b (ignored), k, K, M, G, T")
+	bv(&kola.QEMUOptions.EnableTPM, "qemu-tpm", false, "enable TPM device in QEMU. Requires installing swtpm. Use only with 'kola spawn', test cases are responsible for creating a VM with TPM explicitly.")
 
 	// BrightBox specific options
 	sv(&kola.BrightboxOptions.ClientID, "brightbox-client-id", "", "Brightbox client ID")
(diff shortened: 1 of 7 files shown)


cmd/kola/options.go
```diff
@@ -227,6 +227,7 @@ func init() {
 	sv(&kola.QEMUOptions.BIOSImage, "qemu-bios", "", "BIOS to use for QEMU vm")
 	bv(&kola.QEMUOptions.UseVanillaImage, "qemu-skip-mangle", false, "don't modify CL disk image to capture console log")
 	sv(&kola.QEMUOptions.ExtraBaseDiskSize, "qemu-grow-base-disk-by", "", "grow base disk by the given size in bytes, following optional 1024-based suffixes are allowed: b (ignored), k, K, M, G, T")
+	bv(&kola.QEMUOptions.EnableTPM, "qemu-tpm", false, "enable TPM device in QEMU. Requires installing swtpm.")
```
pothos, line 230: That's a bit confusing because the option also shows up for `kola run` but we don't want it for `kola run`. Can you move this into `cmd/kola/spawn.go`?

jepio: i could, but is it worth the complexity? if someone passes this to `kola run` then they just get a tpm in every test case, shouldn't be problematic (eg. all GCP instances have a tpm)

platform/machine/qemu/cluster.go
```diff
@@ -43,6 +43,7 @@ type Cluster struct {
 func (qc *Cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
 	options := platform.MachineOptions{
 		ExtraPrimaryDiskSize: qc.flight.opts.ExtraBaseDiskSize,
+		EnableTPM:            qc.flight.opts.EnableTPM,
```
pothos, line 46:
```suggestion
		// Used for 'kola spawn', for 'kola run' the test cases that need a TPM should use NewMachineWithOptions
		EnableTPM:            qc.flight.opts.EnableTPM,
```

cmd/kola/options.go
pothos: Ok to leave it here if we add more info
```suggestion
	bv(&kola.QEMUOptions.EnableTPM, "qemu-tpm", false, "enable TPM device in QEMU. Requires installing swtpm. This should only be set for 'kola spawn', and for 'kola run' the test case is responsible for creating a VM with a TPM explicitly")
```

pothos:
FYI, the qemu script with TPM support can already be downloaded here: https://bincache.flatcar-linux.net/images/amd64/3930.0.0+nightly-20240404-2100/flatcar_production_qemu_uefi.sh

cmd/kola/options.go
jepio: ok added a comment like that

platform/machine/qemu/cluster.go
jepio: ok added a comment like that

pothos (approved):
One gofmt run needed before merging

PR #524, opened 2024-04-12
kubeadm: run tests with sysext images

tormath1:
In this PR, we test Kubernetes using the sysext images from the sysext-bakery. It eases the test writing as we remove the installation of Kubernetes components and we are actually testing sysext images that are being used in other projects (Kubernetes deployments and CAPI deployments).

## Changes:
* use systemd sysext instead of binaries
* convert CLC to Butane configuration
* remove useless variables
* remove 'checksum.sh'
* enable calico and cilium CNI tests on Brightbox (they were deactivated due to CLC limitation)
* sync Kubernetes version with available sysext images for 1.27 and 1.28

## Testing done

```
1..3
ok - kubeadm.v1.27.6.calico.cgroupv1.base
ok - kubeadm.v1.28.7.flannel.base
ok - kubeadm.v1.29.2.cilium.base
```

## Notes for reviewers

* As usual `testdata` content is generated from `templates.go`, no need to check the content of this folder, only `templates.go` is important

### DIFF
--- a/kola/tests/kubeadm/kubeadm.go
+++ b/kola/tests/kubeadm/kubeadm.go
@@ -175,12 +175,6 @@ func init() {
 					},
 					MinVersion: semver.Version{Major: major},
 					Flags:      flags,
-					SkipFunc: func(version semver.Version, channel, arch, platform string) bool {
-						// * LTS (3033) does not have the network-kargs service pulled in:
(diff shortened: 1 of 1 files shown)


pothos (approved):
Looks good, thanks!
We could rerun the Kubernetes test on all platforms or let the nightly do it next time.

Do you think we should have a clilium and help sysext image? These two are downloaded as plain binaries now and it might be good to keep them up-to-date, not so important for this test but it might be helpful for users.

PR #526, opened 2024-04-16
kola: Run qemu and swtpm with the machine folder as working dir

pothos:
The long UNIX socket path caused problems and the workaround was to shorten the test names but a real solution is to ensure that the path is always short. This can be done by setting the current working directory for the swtpm process to the machine folder which also holds the socket and, in addition, setting the current working directory for the qemu process to the same folder. Then we can use 'tpm/socket' as path which is ensured to be below the limit of 108 characters.

## Testing done

Tests [passed on GitHub Actions](https://github.com/flatcar/scripts/actions/runs/8705614424?pr=1908) in this [test PR](https://github.com/flatcar/scripts/pull/1908)

### DIFF
--- a/platform/local/tpm.go
+++ b/platform/local/tpm.go
@@ -3,6 +3,7 @@ package local
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/coreos/pkg/capnslog"
 	"github.com/flatcar/mantle/system/exec"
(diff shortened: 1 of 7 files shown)


platform/local/tpm.go
```diff
@@ -3,26 +3,36 @@ package local
...
+	dir     string
+	subDir  string
+	testDir string
 }
 
-func NewSwtpm(dir string) (*SoftwareTPM, error) {
-	swtpm := &SoftwareTPM{}
+func NewSwtpm(testDir string, tpmDir string) (*SoftwareTPM, error) {
+	dir := filepath.Join(testDir, tpmDir)
+	swtpm := &SoftwareTPM{dir: dir, subDir: tpmDir, testDir: testDir}
```
jepio, line 22: this is super confusing :/
- testDir does not need to be stored in struct swtpm, it's not used after `NewSwtpm()`.
- get rid of the tpmDir parameter, and make the "tpm" dir internal to the class
- `SocketSubDirPath` -> `SocketRelativePath` (relative to the passed directory)
- dir -> something more expressive, cwd maybe?

pothos: Yes, updated it now to use more descriptive names

jepio (approved):
these names are long 😆

PR #530, opened 2024-05-03
kola: Exclude user dbus session entries from dead symbolic links check

pothos:
After enabling the user dbus session, the dead link check fails because the path always contains dead links but that's expected behavior.

## How to use

Fixes test failure in https://github.com/flatcar/scripts/pull/1964

## Testing done

Local run passes for the built image


krnowak (approved):
For the record, the symlinks in the directory look like `invocation:dbus.service -> 090442be6a314eaa9cc03c23cd1d90bc`, which is to say - the entries are not treated as symlinks, but more like key-value mappings.

PR #535, opened 2024-06-20
plume/azure: Add support for managed identity

sayanchowdhury:
Add the support to run the pre-release and the release jobs via managed idenities 

to be merged with https://github.com/flatcar/scripts/pull/2048


jepio (approved):
lgtm, can be merged once tested

PR #551, opened 2024-08-28
kola/tests: Add valid SUID binaries in /usr/bin

krnowak:
The mount.nfs and unix_chkpwd binaries used to reside in sbin, but after switching to the merged-usr filesystem layout, sbin became a symlink to bin. But keep the entries for sbin too, for older Flatcar versions.

Needed by https://github.com/flatcar/scripts/pull/2068.


krnowak:
Thanks for the review, I'll merge it once the other PRs are ready.

PR #552, opened 2024-08-29
api/equinixmetal: allows using different port for webserver storage

tormath1:
One can now provide a URL like: ssh+http://1.2.3.4:8088 to upload the temporary artifacts on server 1.2.3.4 via SSH and download the artifacts from http://1.2.3.4:8088


krnowak (commented):
Just noticed that this is for a review for 3 months already. Heh.

Um, looks ok I think. What is this for?

PR #553, opened 2024-08-29
kola: add board->arch mapping for Hetzner

apricote:
# Fix kolet binary architecture on Hetzner

Add a mapping for `hetzner` platform so the right `kolet` binary is uploaded.

Without this, kola would always upload the amd64 kolet, even for arm64 tests.

This failed with:

    bash: line 1: ./kolet: cannot execute binary file: Exec format error

## How to use

See https://github.com/flatcar/scripts/pull/2142 for usage.

## Testing done

Tested it locally with (only important flags):

```
bin/kola --board="arm64-usr" --platform hetzner --hetzner-server-type cax11 run cl.basic
```

Which failed without these changes and succeeds now.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -673,6 +673,9 @@ func architecture(pltfrm string) string {
 	if pltfrm == "azure" && AzureOptions.Board != "" {
 		nativeArch = boardToArch(AzureOptions.Board)
 	}
+	if pltfrm == "hetzner" && HetznerOptions.Board != "" {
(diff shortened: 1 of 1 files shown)


apricote:
I do not really understand the purpose of the `architecture` function. Maybe this can be removed altogether?

apricote:
> (i.e all the <platform>Options.Board should be dropped in favor of a generic one)

That is already the generic field from `Options.Board` that just get embedded in the platform options.

PR #560, opened 2024-10-03
kola: cork: Handle QCOW2 EFI firmare images and fix arm64 Secure Boot

chewi:
# Handle QCOW2 EFI firmare images and fix Secure Boot

Forthcoming Flatcar releases will use QCOW2 instead of raw .fd images.

Only pass SMM-related arguments to QEMU for amd64-usr. EDK2 doesn't have SMM for arm64 yet and these arguments break it when Secure Boot is enabled.

Skip tests that change the verity hash when Secure Boot is enabled. Changing the verity hash breaks Secure Boot verification, causing GRUB to error and then just sit at the menu forever. It's not clear why these tests worked before we applied the Red Hat patches to GRUB, but it's now behaving as it should.

## How to use

Run tests against one of the new Flatcar Secure Boot builds.

## Testing done

Jenkins without qemu_uefi_secure is [all green](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1772/cldsv/). I'm running qemu_uefi_secure now.

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -372,9 +372,22 @@ func CreateQEMUCommand(board, uuid, firmware, ovmfVars, consolePath, confPath, d
 		"-device", "virtio-rng-pci,rng=rng0",
 	)
 	if ovmfVars != "" {
+		var fwFormat, varsFormat string
+
+		if strings.HasSuffix(firmware, ".qcow2") {
+			fwFormat = "qcow2"
+		} else {
+			fwFormat = "raw"
+		}
+		if strings.HasSuffix(ovmfVars, ".qcow2") {
+			varsFormat = "qcow2"
+		} else {
+			varsFormat = "raw"
+		}
+
 		qmCmd = append(qmCmd,
-			"-drive", fmt.Sprintf("if=pflash,unit=0,file=%v,format=raw,readonly=on", firmware),
-			"-drive", fmt.Sprintf("if=pflash,unit=1,file=%v,format=raw", ovmfVars),
+			"-drive", fmt.Sprintf("if=pflash,unit=0,file=%v,format=%v,readonly=on", firmware, fwFormat),
+			"-drive", fmt.Sprintf("if=pflash,unit=1,file=%v,format=%v", ovmfVars, varsFormat),
 		)
 		if enableSecureboot {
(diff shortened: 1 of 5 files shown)


platform/qemu.go
```diff
@@ -372,11 +372,24 @@ func CreateQEMUCommand(board, uuid, firmware, ovmfVars, consolePath, confPath, d
...
+		}
+
 		qmCmd = append(qmCmd,
-			"-drive", fmt.Sprintf("if=pflash,unit=0,file=%v,format=raw,readonly=on", firmware),
-			"-drive", fmt.Sprintf("if=pflash,unit=1,file=%v,format=raw", ovmfVars),
+			"-drive", fmt.Sprintf("if=pflash,unit=0,file=%v,format=%v,readonly=on", firmware, fwFormat),
+			"-drive", fmt.Sprintf("if=pflash,unit=1,file=%v,format=%v", ovmfVars, varsFormat),
 		)
-		if enableSecureboot {
+		if enableSecureboot && board == "amd64-usr" {
```
ader1990, line 392: was this a bug or just that there was no test on arm64 with secure boot?

chewi: We're only just getting Secure Boot working on arm64 for the first time now.

ader1990:
Can you please link the PR that changes the production of the artifact flatcar_production_qemu_uefi_efi_code.qcow2?

chewi:
There is no Flatcar PR yet because the changes are still being finalised, but you see the relevant commit at flatcar/scripts@577cd06579b7f2a0a4f41d746dee49e982fbcbd4.

There is a Gentoo PR though, as the change stems from there. See gentoo/gentoo#38813.

chewi:
Turns out qemu_uefi_secure was failing due to another Kola issue, so I've added another commit. This change is not arm64-specific, rather it broke once we started applying Red Hat's patches to GRUB.

ader1990 (approved):
Small note: currently, this PR will not change the behaviour regarding the usage of `flatcar_production_qemu_uefi_efi_code.qcow2` as the CI will override the values and does not need the https://github.com/gentoo/gentoo/pull/38813.

When Flatcar/scripts is updated with the upstream https://github.com/gentoo/gentoo/pull/38813, then it will be used.

PR #561, opened 2024-10-24
cgroupv1: set Linux cmd arg SYSTEMD_CGROUP_ENABLE_LEGACY_FORCE=1 to w…

ader1990:
…ork with systemd >= v256

## Testing done


ader1990:
This fix does not work in the context of systemd 256 refusing to boot with cgroupv1 enabled, as ignition starts after systemd.

PR #573, opened 2025-01-20
test/nfs: lift qemu restriction

tormath1:
this test uses network and private IP so it makes sense to enable it on cloud providers.

Tested on OpenStack / Brightbox:
```
1..2
ok - linux.nfs.v3
ok - linux.nfs.v4
```


t-lo (approved):
LGTM, thank you for the swift turnaround!

PR #576, opened 2025-02-05
Revert ".github: Update docker/build-push-action to v6"

tormath1:
This reverts commit 6859cfff42184e7d090c2bb58ef2afe7fbcefde4.

Let's see if it fixes ARM64 build issue.


krnowak:
The reason I tried bumping the action to v6 was to fix the same issue. It didn't help, so I think it's unrelated, really.

PR #579, opened 2025-02-05
Dockerfile: Repeat apt-get several times until it succeeds

krnowak:
Last-ditch effort to work-around issues with docker image builds.


krnowak:
Filed an issue here: https://github.com/flatcar/Flatcar/issues/1634

PR #580, opened 2025-02-06
Dockerfile: Fix casing

krnowak:
The warnings are appearing inside GitHub Actions about the inconsistent casing of keywords "from" and "as". Silence them.


krnowak:
Will merge this one together with #572, when it gets approved. We can avoid a useless bump in scripts repo that way.

PR #583, opened 2025-02-27
nvidia arm64 & GPU operator test

jepio:
- Add SkipFunc implementation for skipping test on unsupported instance types
- Add GPU operator test (includes nvidia-runtime sysext test)
- Add Arm64 support to both tests
- Add AWS support

### DIFF
--- a/kola/tests/misc/nvidia.go
+++ b/kola/tests/misc/nvidia.go
@@ -3,8 +3,10 @@ package misc
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/pkg/capnslog"
 	"github.com/flatcar/mantle/kola"
 	"github.com/flatcar/mantle/kola/cluster"
@@ -28,13 +30,19 @@ func init() {
 		Platforms:     []string{"azure"},
 		Architectures: []string{"amd64"},
 		Flags:         []register.Flag{register.NoEnableSelinux},
+		SkipFunc:      skipOnNonGpu,
 	})
 }
 
-func verifyNvidiaInstallation(c cluster.TestCluster) {
-	if kola.AzureOptions.Size != "Standard_NC6s_v3" {
-		c.Skip("skipping due to wrong instance size")
+func skipOnNonGpu(version semver.Version, channel, arch, platform string) bool {
+	// N stands for GPU instance obviously :)
+	if platform == "azure" && strings.Contains(kola.AzureOptions.Size, "NC") {
+		return false
 	}
+	return true
+}
+
+func verifyNvidiaInstallation(c cluster.TestCluster) {
 	m := c.Machines()[0]
 
 	nvidiaStatusRetry := func() error {
--- a/kola/tests/misc/nvidia.go
+++ b/kola/tests/misc/nvidia.go
@@ -21,27 +21,12 @@ const (
 	CmdTimeout = time.Second * 300
 )
 
-const nvidiaDriverVersionOverride = `
-variant: flatcar
-version: 1.0.0
-storage:
-  files:
-  - path: /etc/flatcar/nvidia-metadata
-    contents:
-      inline: |
-        {{ .NVIDIA_DRIVER_VERSION_LINE }}
-`
-
 const nvidiaOperatorTemplate = `
 variant: flatcar
 version: 1.0.0
 
 storage:
   files:
-  - path: /etc/flatcar/nvidia-metadata
-    contents:
-      inline: |
-        NVIDIA_DRIVER_VERSION=570.86.15
   - path: /opt/extensions/kubernetes-v1.30.8-{{ .ARCH_SUFFIX }}.raw
     contents:
       source: https://github.com/flatcar/sysext-bakery/releases/download/latest/kubernetes-v1.30.8-{{ .ARCH_SUFFIX }}.raw
@@ -63,7 +48,7 @@ func init() {
 	register.Register(&register.Test{
 		Name:          "cl.misc.nvidia",
 		Run:           verifyNvidiaInstallation,
-		ClusterSize:   0,
+		ClusterSize:   1,
 		Distros:       []string{"cl"},
 		Platforms:     []string{"azure", "aws"},
 		Architectures: []string{"amd64", "arm64"},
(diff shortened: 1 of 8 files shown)


kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,239 @@ package misc
...
+variant: flatcar
+version: 1.0.0
+
+storage:
+  files:
+  - path: /etc/flatcar/nvidia-metadata
+    contents:
+      inline: |
+        NVIDIA_DRIVER_VERSION=570.86.15
+  - path: /opt/extensions/kubernetes-v1.30.4-{{ .ARCH_SUFFIX }}.raw
```
krnowak, line 45: Is there a reason for picking exactly this version? As opposed to, say, 1.30.8 or any newer major version. Asking, because I think that eventually we will need to be bumping those versions, so maybe in future we will want to change the code to always download the latest version of k8s or nvidia runtime.

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,239 @@ package misc
...
+const nvidiaOperatorTemplate = `
+variant: flatcar
+version: 1.0.0
+
+storage:
+  files:
+  - path: /etc/flatcar/nvidia-metadata
+    contents:
+      inline: |
+        NVIDIA_DRIVER_VERSION=570.86.15
```
krnowak, line 44: Should we upgrade the nvidia drivers in the coreos-overlay instead of hardcoding this version?

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,239 @@ package misc
...
+}
+
+func verifyNvidiaInstallation(c cluster.TestCluster) {
+	params := map[string]string{}
+	// Earlier driver versions have issue building on arm64 with kernel 6.6
+	if kola.QEMUOptions.Board == "arm64-usr" {
+		params["NVIDIA_DRIVER_VERSION_LINE"] = "NVIDIA_DRIVER_VERSION=570.86.15"
+	} else {
+		params["NVIDIA_DRIVER_VERSION_LINE"] = ""
+	}
```
krnowak, line 119: Would this still be necessary if we updated the nvidia-drivers package in overlay?

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,239 @@ package misc
...
+	_ = c.MustSSH(m, "sudo kubeadm init --pod-network-cidr=10.244.0.0/16")
+	_ = c.MustSSH(m, `mkdir -p $HOME/.kube
+	sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
+	sudo chown $(id -u):$(id -g) $HOME/.kube/config`)
+	_ = c.MustSSH(m, "kubectl apply -f https://github.com/flannel-io/flannel/releases/latest/download/kube-flannel.yml")
+	_ = c.MustSSH(m, "kubectl taint nodes --all node-role.kubernetes.io/control-plane-")
+	_ = c.MustSSH(m, "kubectl describe nodes $HOSTNAME")
+	err = util.Retry(5, 10*time.Second, func() error {
+		// notice the extra space before "Ready", it's to not catch
+		// "NotReady" nodes
```
krnowak, line 172: This is probably outdated.

jepio: Yes, removed it.

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,239 @@ package misc
...
+		SkipFunc:      skipOnNonGpu,
 	})
 }
 
-func verifyNvidiaInstallation(c cluster.TestCluster) {
-	if kola.AzureOptions.Size != "Standard_NC6s_v3" {
-		c.Skip("skipping due to wrong instance size")
+func skipOnNonGpu(version semver.Version, channel, arch, platform string) bool {
+	// N stands for GPU instance obviously :)
+	if platform == "azure" && strings.Contains(kola.AzureOptions.Size, "N") {
```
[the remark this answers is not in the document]
jepio: I've updated this to say "NC" because that's correct too and narrows it down.

kola/tests/misc/nvidia.go
jepio: any version is good enough, i've bumped it to 1.30.8.

I think we'll need to update this eventually but i don't expect it to have to always be the latest - the GPU operator itself has a matrix of supported k8s version.

I think this test is going to be most helpful to test the nvidia-runtime sysext and any changes to flatcar's nvidia.service that may break the GPU operator.

kola/tests/misc/nvidia.go
jepio: R570 seems to be the first one that compiles correctly for aarch64 with kernel 6.6 and our gcc version, older ones have a build error. But according to nvidia docs (https://docs.nvidia.com/datacenter/tesla/drivers/index.html#cuda-and-drivers-table) R535 is actually LTS, so it makes for a better default for Flatcar. Nvidia driver's on amd64 are also much more widely used.

How about I upgrade only arm64 to 570 in coreos-overlay?

krnowak: > R570 seems to be the first one that compiles correctly for aarch64 with kernel 6.6 and our gcc version, older ones have a build error. But according to nvidia docs (https://docs.nvidia.com/datacenter/tesla/drivers/index.html#cuda-and-drivers-table) R535 is actually LTS, so it makes for a better default for Flatcar. Nvidia driver's on amd64 are also much more widely used.
> 
> How about I upgrade only arm64 to 570 in coreos-overlay?

Sounds good. amd64 will eventually catch up to arm64.

Another idea that came to my mind is to have a patch for R535 to fix the build issue, but I have no clue if it makes sense and how much effort that would be.

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,209 @@ package misc
...
+      source: https://github.com/flatcar/sysext-bakery/releases/download/latest/kubernetes-v1.30.8-{{ .ARCH_SUFFIX }}.raw
+  - path: /opt/extensions/nvidia_runtime-v1.16.2-{{ .ARCH_SUFFIX }}.raw
+    contents:
+      source: https://github.com/flatcar/sysext-bakery/releases/download/latest/nvidia_runtime-v1.16.2-{{ .ARCH_SUFFIX }}.raw
+  links:
+  - path: /etc/extensions/kubernetes.raw
+    target: /opt/extensions/kubernetes-v1.30.8-{{ .ARCH_SUFFIX }}.raw
+    hard: false
+  - path: /etc/extensions/nvidia_runtime.raw
+    target: /opt/extensions/nvidia_runtime-v1.16.2-{{ .ARCH_SUFFIX }}.raw
```
krnowak, line 41: Can I ask you to make kubernetes version and nvidia runtime versions template parameters? And the actual version could be constants defined next to the `CmdTimeout` constant above. This would make it easier to update the versions in the future, either automatically or manually.

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,209 @@ package misc
...
+		}
+		return nil
+	})
+	if err != nil {
+		c.Fatalf("%v", err)
+	}
+	_ = c.MustSSH(m, "/opt/bin/helm repo add nvidia https://helm.ngc.nvidia.com/nvidia  && /opt/bin/helm repo update")
+	_ = c.MustSSH(m, `/opt/bin/helm install --wait --generate-name \
+	-n gpu-operator --create-namespace \
+	--version v24.9.2 \
```
krnowak, line 158: I think this version could also be defined as a constant, next to kubernetes and nvidia runtime version constants. Thanks.

kola/tests/misc/nvidia.go
```diff
@@ -3,50 +3,209 @@ package misc
...
+	_ = c.MustSSH(m, `kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name: cuda-vectoradd
+spec:
+  restartPolicy: OnFailure
+  containers:
+  - name: cuda-vectoradd
+    image: "nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.7.1-ubuntu20.04"
```
krnowak, line 188: Maybe this docker tag could also be a constant?

kola/tests/misc/nvidia.go
jepio: nope, done

kola/tests/misc/nvidia.go
jepio: done

kola/tests/misc/nvidia.go
jepio: done

PR #594, opened 2025-03-28
fix ore azure segfault

jepio:
Missing ResourceGroupBasename in `ore` caused `ore` to have to process all rgs, also those missing the tags mantle expects. Ensure `ore azure gc` processes rgs that start with `kola-cluster` by default.


chewi (approved):

jepio:
Had to rerun `go fmt`.

sayanchowdhury (approved):

PR #600, opened 2025-04-23
kola: Simplify NewMachineWithOptions usage

krnowak:
It's a drive-by change I made while tinkering with the raid tests.


chewi (approved):
I'm lacking the Go knowledge to understand this fully, but I think it looks good.

PR #622, opened 2025-05-27
test/nftables: adjust query for LTS-3510

tormath1:
`jq` version of LTS 3510 do not like the `.foo.[]` but the `.foo[]` syntax.

Which seems to be the correct one [documented](https://jqlang.org/manual/#array-object-value-iterator)

## Testing done

Tested with LTS 3510 and the on-going Alpha release.


chewi (approved):
Ah! My bad! :flushed:

PR #648, opened 2025-08-07
openstack fixes

jepio:
# openstack fixes

* long running tests end up hitting token expiry on openstack. enable reauth
* implement GC for keypairs. turns out they do export a "created_at" field. already executed.
* logging didn't work for ore commands because of how mantle set `PreRun()` functions. fixed this by calling `startLogging()` in the wrapper around a usersupplied function.

## How to use

## Testing done

### DIFF
--- a/platform/api/openstack/api.go
+++ b/platform/api/openstack/api.go
@@ -112,6 +112,15 @@ func New(opts *Options) (*API, error) {
 		Username:         profile.Username,
 		Password:         profile.Password,
 		DomainID:         profile.DomainID,
+		// Enable automatic re‑authentication so that long‑running
+		// kola test suites do not fail once the Keystone token
+		// expires (typically after one hour).  With AllowReauth set
+		// to true, gophercloud will transparently obtain a fresh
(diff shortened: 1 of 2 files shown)


platform/api/openstack/api.go
```diff
@@ -583,6 +592,106 @@ func (a *API) DeleteImage(imageID string) error {
...
+			}
+		}
+		if parseErr != nil {
+			plog.Warningf("unable to parse created_at for keypair %s: %v", kp.Name, parseErr)
+			continue
+		}
+
+		if now.Sub(createdTime) > olderThan {
+			if err := a.DeleteKey(kp.Name); err != nil {
+				plog.Warningf("failed deleting stale keypair %s: %v", kp.Name, err)
```
pothos, line 683: The behavior in this case differs a bit from other deletions where it's not a warning but a hard error, yet I think it makes sense to continue and maybe the other calls could also be adjusted to allow continuing - it gives less visibility though and I think that piled up keys are less harmful than piled up running VMs. So, in the end the different behavior for these cases is a good tradeoff?

PR #670, opened 2025-09-22
kola: Change cl.update.payload-boot-part-too-small to always fill /boot

pothos:
The test case used a fixed 20 MB file to fill the /boot disk so that the second kernel copy wouldn't fit. However, with the reduced minimal initrd the kernel fits four times and thus the single 20 MB file doesn't prevent the update.
Change the dd call to write as much as possible so that the disk is always filled afterwards.

## How to use/Testing done

With the image and update payload from [bincache](https://bincache.flatcar-linux.net/images/amd64/9999.9.9+kai-initrd-in-usr/) to have a case where the vmlinuz file fits multiple times.

Note that this test isn't actually executed on CI, we need to add a call in `ci-automation/vendor-testing/qemu_update.sh`

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -266,7 +266,8 @@ func payload(c cluster.TestCluster) {
 
 func payloadBootPartTooSmall(c cluster.TestCluster) {
 	addr, m := payloadPrepareMachine(nil, c)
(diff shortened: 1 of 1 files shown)


chewi (approved):

ader1990:
Yeap, I left the 20MB as a good size on purpose so that, if there was the case that the test somehow passed, there was more size available and this kind of PR was necessary.

PR #853, opened 2026-05-19
Dockerfile: Add zstd

zhijianli88:
The ./run_local_tests.sh always expects this command to be existed. Add zstd to avoid the error messages

 $ ./run_local_tests.sh arm64 2 cl.cloudinit.script
 =================================
 Using Mantle docker image 'ghcr.io/flatcar/mantle:git-ca80a2eaee4cc195ae6e17f9202c1d72e729d460'
 =================================
 Running qemu_uefi tests
 git-ca80a2eaee4cc195ae6e17f9202c1d72e729d460: Pulling from flatcar/mantle
 Digest: sha256:cd37c1ba21b8051eb2632e171d4f1f9bc7e91708883575343265d3d58940d5df
 Status: Image is up to date for ghcr.io/flatcar/mantle:git-ca80a2eaee4cc195ae6e17f9202c1d72e729d460
 zstd could not be found. unpacking container image may fail.
 ++++ Running qemu_uefi.sh inside __TESTS__/qemu_uefi ++++
 ++++ qemu_uefi.sh: Using existing /work/__build__/images/images/arm64-usr/latest/flatcar_production_image.bin for testing 4593.2.0+4-gbc1fd0010d (arm64) ++++
 ++++ qemu_uefi.sh: Using existing /work/__build__/images/images/arm64-usr/latest/flatcar_production_qemu_uefi_efi_code.qcow2 ++++
 ++++ qemu_uefi.sh: Using existing /work/__build__/images/images/arm64-usr/latest/flatcar_production_qemu_uefi_efi_vars.qcow2 ++++
 === RUN   cl.cloudinit.script
 --- PASS: cl.cloudinit.script (216.26s)
 PASS, output in _kola_temp/qemu-2026-05-19-0704-13                                                                                                                                                                                     git-ca80a2eaee4cc195ae6e17f9202c1d72e729d460: Pulling from flatcar/mantle                                                                                                                                                              Digest: sha256:cd37c1ba21b8051eb2632e171d4f1f9bc7e91708883575343265d3d58940d5df
 Status: Image is up to date for ghcr.io/flatcar/mantle:git-ca80a2eaee4cc195ae6e17f9202c1d72e729d460
 zstd could not be found. unpacking container image may fail.

# [Title: describe the change in one sentence]

[ describe the change in 1 - 3 paragraphs ]

## How to use

## Testing done

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN for i in $(seq 1 5); do \
     if apt-get --fix-broken install --no-install-recommends -y \
         apt-transport-https awscli ca-certificates curl dns-root-data dnsmasq git gnupg2 iptables jq \
         lbzip2 nftables ovmf python-is-python3 python3 python3-pip qemu-efi-aarch64 qemu-system-aarch64 qemu-system-x86 \
-        qemu-utils rclone seabios sqlite3 sudo swtpm; then \
+        qemu-utils rclone seabios sqlite3 sudo swtpm zstd; then \
             echo "succeeded at ${i}"; \
             break; \
     elif test $i -eq 5; then \


chewi:
I believe this error comes from the scripts repo where `run_local_tests.sh` is run from, so adding zstd to the Mantle container won't make any difference. Even then, I think the error is harmless because we're not unpacking the image in this context.

zhijianli88:
> I believe this error comes from the scripts repo where `run_local_tests.sh` is run from, so adding zstd to the Mantle container won't make any difference. 

Not exactly. The run_local_tests.sh script launches the Mantle container to run the kola tests.


> Even then, I think the error is harmless because we're not unpacking the image in this context.

That's true. I was also wondering whether I should update the related script in the scripts repo. However, this script is a dependency for many other tools, so I'm a bit hesitant to change it.

```
scripts$ git grep -A2 'command -v zstd'
ci-automation/ci-config.env:if ! command -v zstd > /dev/null; then
ci-automation/ci-config.env-  # we require zstd and it is included by default on flatcar
ci-automation/ci-config.env-  echo >&2 "zstd could not be found. unpacking container image may fail."

```

chewi:
Ah okay, on closer inspection, the check is done both outside and inside the container since the scripts repo gets mounted inside it. zstd is only needed when building Flatcar though, not when running the tests. The check can probably be dropped, as it's more confusing than helpful, or maybe it could be moved to a better location.

zhijianli88:
> Ah okay, on closer inspection, the check is done both outside and inside the container since the scripts repo gets mounted inside it. zstd is only needed when building Flatcar though, not when running the tests. The check can probably be dropped, as it's more confusing than helpful, or maybe it could be moved to a better location.

Understood. I folded this update to scripts: https://github.com/flatcar/scripts/pull/4037/changes/505fa4f6fc20ccd998235ab417a2dc645029aee4

PR #882, opened 2026-07-09
Update STACKIT SDK

robinschneider:
# Update STACKIT SDK & improve garbage collection and deletion

STACKIT resource deletion fails regularly and the garbage collection fails to clean up these resources.
- delete reources best effort
- use new SDK path instead of deprecated ones
- manually add label(s) if they are missing
- use a single security group only

## How to use

## Testing done


platform/api/stackit/api.go
```diff
@@ -405,7 +405,34 @@ func (a *API) CreateIPAddress(ctx context.Context) (*PublicIP, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &PublicIP{ipAddress}, nil
+	// The labels sometimes get lost during creation. Verify them and patch the
+	// public IP if they are missing. Retry as the IP might not be ready yet.
+	var verifiedIpAddress *iaas.PublicIp
```
tormath1: Do we need to create a dedicated variable here? From what I understand, `ipAddress` is a pointer so we should  be able to update `ipAddress.Labels` direcly.

platform/api/stackit/api.go
```diff
@@ -398,13 +452,40 @@ func (a *API) CreateIPAddress(ctx context.Context) (*PublicIP, error) {
...
+				Labels: labels,
+			}).Execute()
+			if err != nil {
+				return fmt.Errorf("updating public IP labels: %w", err)
+			}
+			ip.Labels = labels
+			break
+		}
+		verifiedIpAddress = ip
+		return nil
```
tormath1:
```suggestion
			ipAddress.Labels = labels
			return nil
		}
		return nil
```

platform/api/stackit/api.go
```diff
@@ -547,6 +578,31 @@ func (a *API) gcNetworks(ctx context.Context, createdCutoff time.Time) error {
 	return nil
 }
 
+func (a *API) gcFailedNetworks(ctx context.Context, createdCutoff time.Time) error {
```
tormath1, line 640: You might be interested to put this into another commit - this one is about the public IP.

platform/api/stackit/api.go
```diff
@@ -398,13 +452,40 @@ func (a *API) CreateIPAddress(ctx context.Context) (*PublicIP, error) {
...
 	}
-	return &PublicIP{ipAddress}, nil
+	// The labels sometimes get lost during creation. Verify them and patch the
+	// public IP if they are missing. Retry as the IP might not be ready yet.
+	var verifiedIpAddress *iaas.PublicIp
+	err = util.Retry(6, 5*time.Second, func() error {
+		ip, err := a.client.GetPublicIP(ctx, a.projectID, a.region, ipAddress.GetId()).Execute()
+		if err != nil {
+			return fmt.Errorf("getting public IP: %w", err)
+		}
```
tormath1, line 467: Instead of "getting" the IP, can't we just update it directly? If the labels are already here, it's a no-op otherwise it sets them?

robinschneider: Good point, but now I am thinking about if we should keep it and log it to find out if the issue get fixed in the future?

platform/api/stackit/api.go
robinschneider: It is now outdated, not sure if it is resolved as requested

platform/api/stackit/api.go
tormath1: Your call here. Even if it's fixed, I think this code section will stay for a while :D It's just to reduce the number of calls (and the number of issues they can create)

tormath1 (approved):

PR #907, opened 2026-08-04
enhancement(stackit): alignment with other providers

Harsh-2005d:
# Stackit: Alignment with other providers

Fixes- https://github.com/flatcar/Flatcar/issues/2287
- Make DeleteServer synchronous using wait.DeleteServerWaitHandler.
- consistency of error handling across all deletion methods

## Testing done

Build was successful.

### DIFF
--- a/platform/api/stackit/api.go
+++ b/platform/api/stackit/api.go
@@ -166,17 +166,20 @@ func (a *API) CreateKeyPair(ctx context.Context, name, publicKey string) (*Keypa
 		Labels:    DefaultLabels,
 	}
 	keypairResponse, err := a.client.CreateKeyPair(ctx).CreateKeyPairPayload(keypairPayload).Execute()
-	if err != nil {
-		return nil, fmt.Errorf("error creating keypair: %s", err)
-	}
 	if isOpenAPINotFound(err) {
 		return nil, ErrorNotFound
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error creating keypair: %s", err)
 	}
 	return &Keypair{keypairResponse}, err
 }
 
 func (a *API) DeleteKeyPair(ctx context.Context, name string) error {
 	keypair, err := a.client.GetKeyPair(ctx, name).Execute()
+	if isOpenAPINotFound(err) {
+		return nil
+	}
 	if err != nil {
 		return fmt.Errorf("failed to get keypair: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


chewi:
I suspect @robinschneider deliberately did it with a retry for a reason.

Harsh-2005d:
@chewi I believe the retry was originally intended to account for the time it takes for the server to be deleted. Since `DeleteServer` now uses `wait.DeleteServerWaitHandler` to block until the resource is deleted, making the retry call redundant. Maybe @robinschneider  can clarify more on it.

robinschneider:
> @chewi I believe the retry was originally intended to account for the time it takes for the server to be deleted. Since `DeleteServer` now uses `wait.DeleteServerWaitHandler` to block until the resource is deleted, making the retry call redundant. Maybe @robinschneider can clarify more on it.

We should not remove the retry mechanism, OpenStack consists of multiple APIs, even if an API such as the one for server responds the server is already deleted, the other APIs might not yet know that and are not yet in sync.
This will cause problems if removed.

Harsh-2005d:
> We should not remove the retry mechanism, OpenStack consists of multiple APIs, even if an API such as the one for server responds the server is already deleted, the other APIs might not yet know that and are not yet in sync. This will cause problems if removed.

Understood. I have restored the retry loop.

platform/api/stackit/api.go
```diff
@@ -701,7 +719,7 @@ func (a *API) gcKeyPairs(ctx context.Context, createdCutoff time.Time) error {
 			continue
 		}
 
-		err := a.client.DeleteKeyPair(ctx, *keyPair.Name).Execute()
+		err := a.DeleteKeyPair(ctx, *keyPair.Name)
```
robinschneider, line 734: You changed it here, but now `gcNetworks`, `gcSecurityGroups` and `gcServers` are inconsistent.

platform/api/stackit/api.go
robinschneider, line 171: This was not part of your change, but can you also align the order of the error cheks here?

robinschneider (changes requested):

platform/api/stackit/api.go
Harsh-2005d: Fixed.

platform/api/stackit/api.go
Harsh-2005d: updated all the gc functions. Also wrapped the `DeleteImage` call and updated the error logs.

PR #917, opened 2026-08-10
kola: Add tests for azure-vm-utils NVMe friendly naming

### LINKED ISSUE #1607: Add tests for azure-vm-utils (NVME friendly naming)  for Azure OEM
We should also consider adding a Kola test specifically for the nvme utils; most of the work is done via udev rules which might clash when integrating updates of udev / systemd with Flatcar.

azure-vm-utils maintainers would appreciate the extended test coverage.

Reference: https://github.com/Azure/azure-vm-utils/blob/main/doc/disk-identification.md#disk-identification-examples-for-various-vm-sizes--configurations

amanmaurya92:
## Description

This PR introduces a new Kola test (`azure.nvme-friendly-naming`) to verify that the `azure-vm-utils` configuration and udev rules are properly installed and correctly identifying NVMe disks on Azure instances. 

Specifically, this test:
- Validates the presence of `azure-nvme-id` and associated `udev` rules (`80-azure-disk.rules`, `10-azure-unmanaged-sriov.rules`).
- Uses `udevadm verify` to ensure the rules don't conflict with existing systemd/udev updates.
- Dynamically discovers NVMe interfaces inside the test environment (`/dev/nvme0n1`) and asserts that the expected friendly symlinks (`/dev/disk/azure/root` or `/dev/disk/azure/os`) have been created properly.

Resolves [#1607](https://github.com/flatcar/Flatcar/issues/1607)

## How to test

Run Kola targeting the Azure platform:
```bash
kola run --platform azure azure.nvme-friendly-naming

### DIFF
--- a/kola/tests/azure/nvme.go
+++ b/kola/tests/azure/nvme.go
@@ -1,0 +1,59 @@
+// Copyright 2024 The Flatcar Maintainers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"strings"
+
+	"github.com/flatcar/mantle/kola/cluster"
+	"github.com/flatcar/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         checkAzureNVMe,
+		ClusterSize: 1,
+		Platforms:   []string{"azure"},
+		Name:        "azure.nvme-friendly-naming",
+		Distros:     []string{"cl"},
+	})
+}
+
+func checkAzureNVMe(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	// Check for the presence of the azure-nvme-id binary
+	c.MustSSH(m, "test -x /usr/sbin/azure-nvme-id")
+
+	// Dynamically detect if NVMe drives are attached to this test instance.
+	// We do not fail if NVMe isn't present because Kola test runner instances
+	// might still use SCSI depending on configuration/generation.
+	out, err := c.SSH(m, "ls /dev/nvme0n1")
+	if err != nil || len(strings.TrimSpace(string(out))) == 0 {
+		// NVMe is not present on this machine, meaning the udev rules won't trigger the symlink creation.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: kola/tests/azure/nvme.go [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+	c.MustSSH(m, "sudo udevadm verify /usr/lib/udev/rules.d/10-azure-unmanaged-sriov.rules")
+
+	// Dynamically detect if NVMe drives are attached to this test instance.
+	// We do not fail if NVMe isn't present because Kola test runner instances
+	// might still use SCSI depending on configuration/generation.
+	out, err := c.SSH(m, "ls /dev/nvme0n1")
+	if err != nil || len(strings.TrimSpace(string(out))) == 0 {
+		// NVMe is not present on this machine, meaning the udev rules won't trigger the symlink creation.
+		// We already checked rules are present and valid, which satisfies basic conditions.
+		return
```
tormath1:
```suggestion
		c.Skip("no NVMe disk found on this instance")
```

tormath1 (commented):
Thanks it looks good. Can you just sign-off your commit?

amanmaurya92:
@tormath1 i have signed off now

kola/tests/azure/nvme.go
```diff
@@ -0,0 +1,67 @@
...
+	// Check for the presence of the azure-nvme-id binary
+	c.MustSSH(m, "test -x /usr/sbin/azure-nvme-id")
+
+	// Check for the presence of the udev rules
+	c.MustSSH(m, "test -f /usr/lib/udev/rules.d/80-azure-disk.rules")
+	c.MustSSH(m, "test -f /usr/lib/udev/rules.d/10-azure-unmanaged-sriov.rules")
+
+	// Verify that the udev rules are syntactically valid by checking udevadm
+	c.MustSSH(m, "sudo udevadm verify /usr/lib/udev/rules.d/80-azure-disk.rules")
+	c.MustSSH(m, "sudo udevadm verify /usr/lib/udev/rules.d/10-azure-unmanaged-sriov.rules")
```
chewi:
```suggestion
```

I would not be surprised if these files get renamed at some point. We can test for azure-nvme-id to ensure that the package was installed, but beyond that, I trust that the rules are there and that they are valid. azure-vm-utils has its own tests, and testing the symlinks below also proves that at least the disk rules are working.

chewi (changes requested):

kola/tests/azure/nvme.go
amanmaurya92: That makes total sense! I've removed the hardcoded file checks to avoid them breaking in the future, and added a quick test for azure-nvme-id just to ensure the package is installed. Thanks for the guidance!

chewi:
DCO is still complaining, probably because of the merge commits. We prefer clean history anyway, so please squash this back to a single commit.

amanmaurya92:
@chewi I have squashed this back to a single commit.

chewi:
Great, thanks. I manually run this against Azure, and it worked.

PR #927, opened 2026-08-19
kola/docker: convert docker.service symlink to upholds dependency

amanmaurya92:
Since the upgrade to systemd 254+, we can now use the stronger multi-user.target.upholds dependency instead of multi-user.target.wants to ensure docker.service automatically restarts if it goes down, which was the original intent of the test.

This PR resolves a longstanding `TODO FIXME` in `kola/tests/docker/enable_docker.go`. It updates the Butane configuration string inside the `docker.enable-service.sysext` test to utilize the `.upholds/` drop-in directory instead of `.wants/`. 

## How to use

Reviewers can validate this PR by running the modified Kola test and confirming that Docker successfully auto-starts on boot using the new `upholds` dependency mapping. 

## Testing done

Compiled the `kola/tests/docker` package locally to ensure there are no syntactic regressions, using `$env:GOOS="linux"; cd kola/tests/docker; go build`.

```text
$env:GOOS="linux"; cd kola/tests/docker; go build
(exit code 0)
```

### DIFF
--- a/kola/tests/docker/enable_docker.go
+++ b/kola/tests/docker/enable_docker.go
@@ -65,13 +65,11 @@ systemd:
     enabled: true
 storage:
   links:
-  - path: /etc/systemd/system/multi-user.target.wants/docker.service
+  - path: /etc/systemd/system/multi-user.target.upholds/docker.service
     target: /usr/lib/systemd/system/docker.service
     hard: false
     overwrite: true
 `),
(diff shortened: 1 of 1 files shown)


tormath1 (approved):
Thanks!

amanmaurya92:
Thanks @tormath1 ,  I have signed off correctly now as expected of DCO.

PR #928, opened 2026-08-20
Update gophercloud to v2

imshubham22apr-gif:
This PR resolves issue #1667 by upgrading the `github.com/gophercloud/gophercloud` dependency from v1 to v2. 

The upgrade to v2 introduces several breaking API changes that have been addressed in this PR. Most notably, the `gophercloud` SDK is now context-aware, requiring a `context.Context` object to be passed as the first argument to all functions that make HTTP calls. All relevant OpenStack API calls in `platform/api/openstack/api.go` have been updated to use `context.TODO()`.

Additionally, several modules have been reorganized or deprecated in v2. The `imageservice` module has been renamed to `image` (Glance API). Furthermore, the deprecated Compute (Nova) Floating IP extension has been completely removed in v2, so the floating IP logic has been structurally rewritten to use the modern Networking (Neutron) API.

## How to use

Reviewers should verify that the new Neutron Floating IP logic correctly provisions and associates floating IPs during server creation, as the underlying OpenStack API calls have completely changed from Nova to Neutron. Ensure that `Mantle` can still successfully create and delete instances with floating IPs on OpenStack.

## Testing done

Successfully built the `openstack` platform package locally targeting the Linux OS environment to verify there are no compilation errors with the new `gophercloud/v2` SDK.

Command run:
`$env:GOOS="linux"; go build ./platform/api/openstack/...`
Output: Extracted successfully with exit code 0.

- [x] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.


tormath1:
I'll have to close this PR in favor of: https://github.com/flatcar/mantle/pull/897 - this is clearly a duplicate. @imshubham22apr-gif  in open-source, you usually check if someone is already working on a an issue (https://github.com/flatcar/Flatcar/issues/1667#issuecomment-5077714104). If you want to contribute on this specific topic, you can help reviewing the initial PR or picking-up another issue. Thanks for your contribution :)

PR #929, opened 2026-08-20
kola/misc: modernize root-on-raid to use Butane

amanmaurya92:
This resolves an ancient `TODO` by migrating the `cl.disk.raid*.root` ignition configuration from legacy JSON to a modern Butane payload. 

Modern Ignition supports `wipe_partition_entry: true`, which allows us to selectively zero out partition 9 directly on the primary boot disk and replace it with partitions 9 and 10 for the RAID array. Because of this upgrade, we have completely removed the artificial requirement for a secondary virtual disk and unlocked the RAID root test to run natively on all cloud platforms, rather than just `qemu`.

## How to use

Reviewers can validate this PR by running the modified Kola test and confirming that the nodes successfully boot with root on RAID natively across different platforms:
```bash
kola run --platform aws cl.disk.raid1.root
kola run --platform qemu cl.disk.raid0.root
```

## Testing done

Executed `$env:GOOS="linux"; cd kola/tests/misc; go build` to syntactically verify the package logic and ensure no compilation regressions occurred. We also wrote a scratch script to run `conf.Butane()` locally, proving that the YAML parser natively parses and correctly translates the `wipe_partition_entry: true` payload into valid Ignition v3 JSON!


amanmaurya92:
can maintainers please review this change?

amanmaurya92:
cc @chewi @tormath1 , please take a look

PR #936, opened 2026-08-27
api/stackit: fix nil pointer exceptions

tormath1:
It happens that nil pointer exception occurs here. I have two hypothesis:
* STACKIT APIs are returning an empty object without error
* The context is ended, for another reason, so the wait call is ending before it succeeds

---

e.g: https://jenkins.flatcar.org/job/container/job/test/2447/console - cc @robinschneider

### DIFF
--- a/platform/api/stackit/api.go
+++ b/platform/api/stackit/api.go
@@ -228,10 +228,18 @@ func (a *API) CreateServer(ctx context.Context, name string, networkId *string, 
 	if err != nil {
 		return nil, fmt.Errorf("error creating server: %s", err)
 	}
-	server, err := wait.CreateServerWaitHandler(ctx, a.client, a.projectID, a.region, *serverResponse.Id).WaitWithContext(ctx)
 
+	var server *iaas.Server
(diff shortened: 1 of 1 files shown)


krnowak (approved):
This API is weird, I must say. I guess it's fine what you did, but I'm wondering - how about just ignoring the server/network returned by the WaitHandler function and just always use serverResponse/networkResponse?

tormath1:
> This API is weird, I must say. I guess it's fine what you did, but I'm wondering - how about just ignoring the server/network returned by the WaitHandler function and just always use serverResponse/networkResponse?

I _think_ it's because the API returns the `serverResponse/networkResponse` as soon the entity is "created" but "created" is different from "ready". Usually in Mantle we use the `util.Retry` function to check the entity to be sure its status is "running" or "ready" from the cloud provider point of view. This is what we do here through the `WaitHandler` method.

But as said, this `WaitHandler` seems to return (for an unknown reason) a nil pointer from time to time, so let's just reuse the previous response in best effort.

PR #952, opened 2026-09-09
lang/worker: add test coverage for WorkerGroup and Parallel

magic-peach:
WorkerGroup and Parallel had no tests at all, despite handling concurrency limits, error capture and cancellation that other parts of mantle rely on. This adds coverage for the concurrency limit, error propagation from a failed worker, cancellation of remaining work after a failure, and the WaitError fallback behavior.

## How to use

Review the two new test files under lang/worker. No behavior in non-test code changed, so there is nothing to run beyond the test suite itself.

## Testing done

go build ./lang/worker/...
go test -race ./lang/worker/... -v

All 9 tests pass, ran 5 times in a row to check for flakiness with no failures.

Not applicable: this only adds test files for existing library code, it does not change any build output or user-facing behavior.

### DIFF
--- a/lang/worker/group_test.go
+++ b/lang/worker/group_test.go
@@ -1,0 +1,188 @@
+// Copyright 2026 The Flatcar Maintainers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


magic-peach:
Recreating without draft status due to a permission issue transitioning out of draft.

